PR #1, opened 2021-03-30
Tokio-uring design proposal

carllerche:
[Rendered](https://github.com/tokio-rs/tokio-uring/blob/design-doc/DESIGN.md)

The RFC proposes a new asynchronous Rust runtime backed by io-uring as a new crate: tokio-uring. The API aims to be as close to idiomatic Tokio, but will deviate when necessary to provide full access to io-uring's capabilities. It also will be compatible with existing Tokio libraries. The runtime will use an isolated thread-per-core model, and many types will be `!Send`.

The repo contains a proof-of-concept implementation.

### DIFF
--- a/DESIGN.md
+++ b/DESIGN.md
@@ -258,7 +258,7 @@ let mut rt = tokio_uring::runtime::Runtime::new()?;
 
 // Provide the buffer pool to the kernel. This passes
 // ownership of the pool to the kernel.
-let pool_token = rt.provide_buffers(MY_POOL, 256, 4 * 4096)?;
+let pool_token = rt.provide_buffers(Mmy_pool)?;
     
 rt.block_on(async {
     // ...
--- a/DESIGN.md
+++ b/DESIGN.md
@@ -712,7 +712,7 @@ select! {
     _ = future::ready(()) => {}
 }
 
-let buf = IoBuf::with(capacity(4096))
+let buf = IoBuf::with_capacity(4096);
 let (res, buf) = my_tcp_stream.read(buf.slice(..)).await;
 ```
 
--- a/DESIGN.md
+++ b/DESIGN.md
@@ -654,6 +654,19 @@ earliest.
 
 ## Alternatives
 
+### Use a work-stealing scheduler
+
+An alternative approach could use a work-stealing scheduler, allowing
+underutilized worker threads to steal work from overloaded workers. While
+work-stealing is an efficient strategy for balancing load across worker threads,
+required synchronization adds overhead. An earlier article on the Tokio blog
+includes an overview of various scheduling strategies.
+
+The tokio-uring crate targets use-cases that can benefit from taking advantage
+of io-uring at the expense of discarding Tokio's portable API. These use cases
+will also benefit from reduced synchronization overhead and fine-grained control
+over thread load balancing strategies.
+
 ### Read and write methods on TcpStream
 
 Read and write operations on byte stream types, such as `TcpStream`, are
@@ -684,3 +697,11 @@ operation and return it as part of the second read. The process would not lose
 data, but the runtime would need to perform an extra copy or return the caller a
 different buffer than the one it submitted.
 
+### Expose a raw io-uring operation submission API
+
+The proposed tokio-uring API does not include a strategy for the user to submit
+custom io-uring operations. Any raw API would be unsafe or would require the
+runtime to support taking opaque data as a trait object. Given that io-uring has
+a well-defined set of supported options, tokio-uring opts to support each
+operation explicitly. The application also can create its own set of io-uring
+queues using the io-uring crate directly.
--- a/DESIGN.md
+++ b/DESIGN.md
@@ -734,7 +734,7 @@ queues using the io-uring crate directly.
 
 ### Load-balancing spawn function
 
-The tokio-uring crate omits a spawn function that balances tasks across tasks,
+The tokio-uring crate omits a spawn function that balances tasks across threads,
 leaving this to future work. Instead, the application should manage its load
 balancing. While there are a few common balancing strategies, such as
 round-robin, randomized, and power of two choices, there is no one-size-fits-all
--- a/DESIGN.md
+++ b/DESIGN.md
@@ -212,7 +212,7 @@ let BufResult(res, buf) = file.read_at(0, buf.slice(0..1024)).await;
 res?;
 
 // Write some data back to the file.
-let BufResult(res, _) = file.write_at(1024, buf.slice(0..512).await;
+let BufResult(res, _) = file.write_at(1024, buf.slice(0..512)).await;
 res?;
 ```
 
--- a/DESIGN.md
+++ b/DESIGN.md
@@ -310,7 +310,8 @@ On drop, if the `IoBuf` is a buffer pool member, it is checked back in. If the
 kernel initially checked out the buffer as part of a read operation, an io-uring
 operation is issued to return it. Submitting the io-uring operation requires the
 buffer to remain on the same thread that checked it out and is enforced by
-making the `IoBuf` type `!Send`. 
+making the `IoBuf` type `!Send`. Buffer pools will also be `!Send` as they
+contain `IoBuf` values.
 
 `IoBuf` provides an owned slice API allowing the caller to read to and write from
 a buffer's sub-ranges. 
@@ -378,15 +379,15 @@ and within an asynchronous context, the drop handler should be non-blocking.
 Closing an io-uring resource requires canceling any in-flight operations, which
 is an asynchronous process. Consider an open TcpStream associated with
 file-descriptor (FD) 10. A task submits a read operation to the kernel, but the
-`TcpStream` is dropped and closed before it sees it. A new `TcpStream` is
-accepted, and the kernel reuses FD 10. At this point, the kernel sees the
+`TcpStream` is dropped and closed before the kernel sees it. A new `TcpStream`
+is accepted, and the kernel reuses FD 10. At this point, the kernel sees the
 original read operation request with FD 10 and completes it on the new
 `TcpStream`, not the intended `TcpStream`. The runtime receives the completion
 of the read operation. It discards the result because the associated operation
-future is gone, resulting in the caller losing data from the second
-TcpStream. This problem occurs even issuing a cancellation request for the read
-operation. There is no guarantee the kernel will see the cancellation request
-before completing the operation.
+future is gone, resulting in the caller losing data from the second TcpStream.
+This problem occurs even issuing a cancellation request for the read operation.
+There is no guarantee the kernel will see the cancellation request before
+completing the operation.
 
 There are two options for respecting both requirements, neither ideal: closing
 the resource in the background or blocking the thread in the drop handler. If
--- a/DESIGN.md
+++ b/DESIGN.md
@@ -668,7 +668,9 @@ The most obvious area is Tokio's file system API, currently implemented using a
 thread-pool. The current API would remain, and the implementation would use
 io-uring when supported by the operating system. The tokio-uring APIs may form
 the basis for a Tokio 2.0 release, though this cannot happen until 2024 at the
-earliest.
+earliest. As an intermittent step, the tokio-uring crate could explore
+supporting alternate backends, such as epoll, kqueue, and iocp. The focus will
+always remain on io-uring.
 
 The current design does not cover registering file descriptors with the kernel,
 which improves file system access performance. After registering a file
--- a/DESIGN.md
+++ b/DESIGN.md
@@ -431,6 +431,8 @@ between two byte streams.
 my_tcp_stream.fill_buf().await?;
 let buf: IoBuf = my_tcp_stream.take_read_buf();
 
+// Mutate `buf` if needed here.
+
 my_other_stream.place_write_buf(buf);
 my_other_stream.flush().await?;
 ```
@@ -488,8 +490,6 @@ waiting on completion events by blocking the thread on the io-uring completion
 queue, the tokio-uring runtime registers the completion queue with the epoll
 handle. By building tokio-uring on top of Tokio's runtime, existing Tokio
 ecosystem crates can work with the tokio-uring runtime.
-
-**TODO: mind blowing architecture diagram**
 
 When the kernel pushes a completion event onto the completion queue,
 "epoll_wait" unblocks and returns a readiness event. The Tokio current-thread
--- a/DESIGN.md
+++ b/DESIGN.md
@@ -723,3 +723,34 @@ runtime to support taking opaque data as a trait object. Given that io-uring has
 a well-defined set of supported options, tokio-uring opts to support each
 operation explicitly. The application also can create its own set of io-uring
 queues using the io-uring crate directly.
+
+### Load-balancing spawn function
+
+The tokio-uring crate omits a spawn function that balances tasks across tasks,
+leaving this to future work. Instead, the application should manage its load
+balancing. While there are a few common balancing strategies, such as
+round-robin, randomized, and power of two choices, there is no one-size-fits-all
+strategy. Additionally, some load-balancing methods require application-specific
+metrics such as worker load.
+
+Additionally, consider pseudocode representing a typical accept loop pattern.
+
+```rust
+loop {
+    let (socket, _) = listener.accept().await?;
+
+    spawn(async move {
+        let request = read_request(&socket).await;
+        let response = process(request).await;
+        write_response(response, &socket).await;
+    });
+}
+```
+
+Often, the request data is already buffered when the socket is accepted. If the
+spawn call results in the task moving to a different worker thread, it will
+delay reading the request due to cross-thread synchronization. Additionally, it
+may be possible to batch an accept operation with an operation that reads from
+the accepted socket in the future, complicating moving the accepted socket to a
+new thread. The details of such a strategy are still not finalized but may
+impact a load-balancing spawn function.
--- a/DESIGN.md
+++ b/DESIGN.md
@@ -79,11 +79,45 @@ The runtime communicates with the kernel using two single-producer,
 single-consumer queues. It submits operation requests, such as accepting a TCP
 socket, to the kernel using the submission queue. The kernel then performs the
 operation. On completion, the kernel returns the operation results via the
-completion queue and notifies the process. Syscalls are not required to transfer
-operation requests and completion results between the process and the kernel,
-though the runtime will use a syscall to block the thread waiting on completion
+completion queue and notifies the process. The `io_uring_enter` syscall flushes
+the submission queue and acquires any pending completion events. Upon request,
+this syscall may block the thread waiting for a minimum number of completion
 events. Both queues use memory shared between the process and the kernel and
 synchronize with atomic operations.
+
+Operation futures provide an asynchronous cancel function, enabling the caller
+to await on a clean cancellation.
+
+```rust
+let accept = tcp_listener.accept();
+
+tokio::select! {
+    (tcp_stream, addr) = &mut accept => { ... }
+    _ = tokio::time::sleep(Duration::from_millis(100)) => {
+        // Accept timed out, cancel the in-flight accept.
+        match accept.cancel().await {
+            Ok(_) => { ... } // operation canceled gracefully
+            Err(Completed(tcp_stream, addr)) => {
+                // The operation completed between the timeout
+                // and cancellation request.
+            }
+            Err(_) => { ... }
+        }
+    }
+}
+```
+
+In practice, operation timeouts will deserve a dedicated API to handle the
+boilerplate.
+
+```rust
+let (tcp_stream, addr) = tcp_listener
+    .accept()
+    .timeout(Duration::from_millis(100))
+    .await?
+```
+
+The cancel and timeout function are inherent methods on operation future types.
 
 If the operation's future drops before the operation completes, the runtime will
 submit a cancellation request for the operation. However, cancellation is
@@ -115,12 +149,9 @@ becomes challenging. Applications must take care to ensure load remains balanced
 across threads, and strategies tend to vary. For example, a TCP server can
 distribute accepted connections across multiple threads, ideally while
 maintaining equal load across threads. One approach is to submit "accept"
-operations for the same listener concurrently across all the runtime threads.
-Overloaded workers can back-off starting new "accept" operations. Defining
-runtime load is out of scope for this design document, though pseudocode is
-included below.
-
-Pseudocode:
+operations for the same listener concurrently across all the runtime threads
+while ensuring overloaded workers back-off. Defining runtime load is out of this
+crate's scope and left to the application, though pseudocode follows.
 
 ```rust
 let listener = Arc::new(TcpListener::bind(...));
@@ -363,10 +394,12 @@ in-flight operation. Once all existing in-flight operations complete, the
 runtime will submit a close operation.
 
 If the drop handler must process closing a resource in the background, it will
-notify the developer by writing a warning message to `STDERR`. In the future, it
-may be possible for Rust to provide a `#[must_not_drop]` attribute. This
-attribute will result in compilation warnings if the developer drops a resource
-without using the explicit close method.
+notify the developer by emitting a warning message using [`tracing`]. In the
+future, it may be possible for Rust to provide a `#[must_not_drop]` attribute.
+This attribute will result in compilation warnings if the developer drops a
+resource without using the explicit close method.
+
+[`tracing`]: https://github.com/tokio-rs/tracing
 
 ## Byte streams
 
@@ -596,12 +629,15 @@ result, transitions the lifecycle to `Completed`, and notifies the waker. The
 next time the caller's task executes, it polls the operation's future which
 completes and returns the stored result.
 
-If the operation's future drops before the operation completes, it sets the
-lifecycle to `Ignored` and submits a cancellation request to the kernel. The
-cancellation request will attempt to terminate the operation, causing it to
-complete immediately with an error. Cancellation is best-effort; the operation
-may or may not terminate early. If the operation does complete, the runtime
-discards the result.
+If the operation's future drops before the operation completes and the operation
+request is still in the submission queue, the drop function removes the request.
+Otherwise, it sets the lifecycle to `Ignored` and submits a cancellation request
+to the kernel. The cancellation request will attempt to terminate the operation,
+causing it to complete immediately with an error. Cancellation is best-effort;
+the operation may or may not terminate early. If the operation does complete,
+the runtime discards the result. The runtime maintains the internal operation
+state until the completion as this state owns data the kernel may be
+referencing.
 
 ## Read operations
 
@@ -622,9 +658,14 @@ This proposal draws heavily from ideas presented there, tweaking concepts to
 line up with Tokio's idioms. [`@withoutboats`][boats] has explored the space
 with [ringbahn].
 
+The tokio-uring crate is built on the pure-rust [io-uring] crate, authored by
+[@quininer]. This crate provides a low-level interface to the io-uring syscalls.
+
 [Glommio]: https://github.com/DataDog/glommio
 [boats]: https://github.com/withoutboats
 [ringbahn]: https://github.com/ringbahn/ringbahn
+[@quininer]: https://github.com/quininer/
+[io-uring]: https://github.com/tokio-rs/io-uring/
 
 ## Future work
 
--- a/DESIGN.md
+++ b/DESIGN.md
@@ -1,0 +1,636 @@
+# Summary
+
+The RFC proposes a new asynchronous Rust runtime backed by [io-uring] as a new
+crate: tokio-iouring. The API aims to be as close to idiomatic Tokio, but will
+deviate when necessary to provide full access to io-uring's capabilities. It
+also will be compatible with existing Tokio libraries. The runtime will use an
+isolated thread-per-core model, and many types will be `!Send`.
+
+[io-uring]: https://kernel.dk/io_uring.pdf?source=techstories.org
+
+# Motivation
+
+Tokio's current Linux implementation uses non-blocking system calls and epoll
+for event notification. With epoll, a tuned TCP proxy will spend [70% to
+80%][overhead] of CPU cycles outside of userspace, including cycles spent
+performing syscalls and copying data between the kernel and userspace. In 2019,
+Linux added a new API, io-uring, which reduces overhead by eliminating most
+syscalls and mapping memory regions used for byte buffers ahead of time. Early
+benchmarks comparing io-uring against epoll are promising; a TCP echo client and
+server show up to [60% improvement][bench]. Though not yet measured, using
+io-uring instead of Tokio's thread-pool strategy will likely provide significant
+gains for file system operations.
+
+Because io-uring differs significantly from epoll, Tokio must provide a new set
+of APIs to take full advantage of the reduced overhead. However, Tokio's
+[stability guarantee][stability] means Tokio APIs cannot change until 2024 at
+the earliest. Additionally, the io-uring API is still evolving with [new
+functionality][tweet] planned for the near future. Instead of waiting for
+io-uring to mature and a Tokio 2.0 release, we will release a standalone crate
+dedicated to exposing an io-uring optimal API. This new crate will be able to
+iterate rapidly with breaking changes without violating Tokio's stability
+guarantee. Applications deployed exclusively on Linux kernels 5.10 or later may
+choose to use this crate and take advantage of io-uring's benefits.
+
+[overhead]: https://www.usenix.org/system/files/conference/nsdi14/nsdi14-paper-jeong.pdf
+[bench]: https://github.com/frevib/io_uring-echo-server/blob/master/benchmarks/benchmarks.md
+[stability]: https://tokio.rs/blog/2020-12-tokio-1-0#a-stability-guarantee
+[tweet]: https://twitter.com/axboe/status/1371978266806919168
+
+# Guide-level explanation
+
+The tokio-uring crate provides a module layout similar to the Tokio crate,
+including modules such as net (TCP, UDP, and Unix domain sockets), fs (file
+system access), io (standard in, out, and error streams). It also provides a
+runtime module containing an io-uring specific runtime. Modules such as sync and
+any other containing async Rust utilities are not included by tokio-uring, in
+favor of the ones provided by the main Tokio crate.
+
+```toml
+[dependencies]
+tokio-uring = "0.1"
+```
+
+```rust
+fn main() {
+    let rt = tokio_uring::runtime::Runtime::new().unwrap();
+    rt.block_on(async {
+        // The rest of the application comes here.
+    });
+}
+```
+
+The application's `main` function starts a tokio-uring runtime and launches its
+asynchronous logic within that. The tokio-uring runtime can drive both io-uring
+specific resources (e.g., `TcpStream` and `TcpListener`) and Tokio resources,
+enabling any library written for Tokio to run within the io-uring specific
+runtime.
+
+## Submit-based operations
+
+Operations on io-uring backed resources return futures, representing the
+operation completion. The caller awaits the future to get the operation result.
+
+```rust
+let socket = my_listener.accept().await?;
+```
+
+The runtime communicates with the kernel using two single-producer,
+single-consumer queues. It submits operation requests, such as accepting a TCP
+socket, to the kernel using the submission queue. The kernel then performs the
+operation. On completion, the kernel returns the operation results via the
+completion queue and notifies the process. Syscalls are not required to transfer
+operation requests and completion results between the process and the kernel,
+though the runtime will use a syscall to block the thread waiting on completion
+events. Both queues use memory shared between the process and the kernel and
+synchronize with atomic operations.
+
+If the operation's future drops before the operation completes, the runtime will
+submit a cancellation request for the operation. However, cancellation is
+asynchronous and best-effort, the operation may still complete. In that case,
+the runtime discards the operation result.
+
+The queue's single-producer characteristic optimizes for a single thread to own
+a given submission queue. Supporting a multi-threaded runtime requires either
+synchronizing pushing onto the submission queue or creating a queue pair per
+thread. The tokio-uring runtime uses an isolated thread-per-core model,
+differing from Tokio's other runtimes. Unlike Tokio's primary multi-threaded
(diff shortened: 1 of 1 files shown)


DESIGN.md
```diff
@@ -0,0 +1,756 @@
...
+```rust
+let socket = my_listener.accept().await?;
+```
+
+The runtime communicates with the kernel using two single-producer,
+single-consumer queues. It submits operation requests, such as accepting a TCP
+socket, to the kernel using the submission queue. The kernel then performs the
+operation. On completion, the kernel returns the operation results via the
+completion queue and notifies the process. The `io_uring_enter` syscall flushes
+the submission queue and acquires any pending completion events. Upon request,
```
pnkfelix, line 85: What does it mean for a specific syscall to "acquire a completion event" ?  (i.e. The syscall itself cannot own responsibility for handling that event, right? So is it handing off that responsibility to whomever in user space invoked the syscall ?)

DESIGN.md
```diff
@@ -0,0 +1,756 @@
...
+
+```rust
+impl TcpStream {
+    async fn close(self) { ... }
+}
+
+my_tcp_stream.close().await;
+```
+
+The resource must still tolerate the caller dropping it without being explicitly
```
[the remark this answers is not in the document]
carllerche: "move ownership" is probably a poor choice of words here. It is a logical transfer of ownership. The value is already stored in the runtime's internal storage. The implementation changes the operation state to `Ignored` which signals to the runtime that it owns that state now.

DESIGN.md
pnkfelix: (Or, I guess since the syscall is allowed to block the thread, it *can* have that responsibility? But below you say waiting for a *minimum* number of completion events; not *all* of the events... so I'm still confused I guess.)

DESIGN.md
```diff
@@ -0,0 +1,756 @@
...
+pub struct IoBuf {
+    kind: Kind,
+}
+
+enum Kind {
+    /// A vector-backed buffer
+    Vec(Vec<u8>),
+
+    /// Buffer pool backed buffer. The pool is managed by io-uring.
+    Provided(ProvidedBuf),
```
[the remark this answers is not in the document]
carllerche: Those are good points for registering buffers, but what about for read operations to a buffer pool instead of an explicit buffer, that isn't about performance but about reducing memory overhead.

DESIGN.md
carllerche: Acquire at the memory ordering level (think atomic load operation w/ Acquire ordering).

DESIGN.md
```diff
@@ -0,0 +1,756 @@
...
+    fn as_mut(&mut self) -> &mut [u8] {
+        &mut self.mem[..]
+    }
+}
+```
+
+This read function takes ownership of the buffer; however, any pointer to the
+buffer obtained from a value becomes invalid when the value moves. Storing the
+buffer value at a stable location while the operation is in-flight should be
+sufficient to satisfy safety.
```
seanmonstar, line 372: It's technically safe to return a new slice or pointer on each call to `as_mut()`, as bonkers as that may be, so it might need something like `unsafe trait TrustedAsMut: AsMut<[u8]> {}`

quininer: If we hold ownership of T and guarantee to call as_mut only once, I think this is trusted.

carllerche: This is correct, but I am *pretty* sure that as long as `as_mut()` is called only once and is kept at a stable location, it should be fine.... but that is also why I am punting 😆

DESIGN.md
```diff
@@ -0,0 +1,756 @@
...
+An alternative approach could use a work-stealing scheduler, allowing
+underutilized worker threads to steal work from overloaded workers. While
+work-stealing is an efficient strategy for balancing load across worker threads,
+required synchronization adds overhead. An earlier article on the Tokio blog
+includes an overview of various scheduling strategies.
+
+The tokio-uring crate targets use-cases that can benefit from taking advantage
+of io-uring at the expense of discarding Tokio's portable API. These use cases
+will also benefit from reduced synchronization overhead and fine-grained control
+over thread load balancing strategies.
```
seanmonstar, line 686: I think not *requiring* synchronization in a sub-library is a good call. Those wanting finer grained control can require pinning to a thread. What about the possibility of making `tokio-uring` support `!Send` or `Send` generically, so that if someone *did* want to use a work-stealing multi-threaded scheduler, they can do so safely without wrapping the types in `Mutex`s? Or is that basically the best solution anyways?

DESIGN.md
```diff
@@ -0,0 +1,756 @@
...
+```rust
+use tokio_uring::buf;
+
+/// The result of an operation that includes a buffer. The buffer must
+/// be returned to the caller when the operation completes successfully
+/// or fails.
+///
+/// This is implemented as a new type to implement std::ops::Try once
+/// the trait is stabilized.
+type BufResult<T> = (std::io::Result<T>, buf::Slice);
```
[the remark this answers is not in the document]
seanmonstar: The newtype could implement a method to convert into a `Result` and drop the buffer on errors, I suppose. `let buf = file.read_(0, buf).await.result()?;` or something.

DESIGN.md
```diff
@@ -0,0 +1,756 @@
...
+The proposed tokio-uring API does not include a strategy for the user to submit
+custom io-uring operations. Any raw API would be unsafe or would require the
+runtime to support taking opaque data as a trait object. Given that io-uring has
+a well-defined set of supported options, tokio-uring opts to support each
+operation explicitly. The application also can create its own set of io-uring
+queues using the io-uring crate directly.
+
+### Load-balancing spawn function
+
+The tokio-uring crate omits a spawn function that balances tasks across tasks,
```
pnkfelix, line 729: "balances tasks across tasks" -- typo? (did you mean "across threads"?)

DESIGN.md
```diff
@@ -0,0 +1,756 @@
...
+    .buffer_size(16_384)
+    .num_buffers(256)
+    .build();
+
+// Create the runtime
+let mut rt = tokio_uring::runtime::Runtime::new()?;
+
+// Provide the buffer pool to the kernel. This passes
+// ownership of the pool to the kernel.
+let pool_token = rt.provide_buffers(MY_POOL, 256, 4 * 4096)?;
```
carllerche, line 261: This line is incorrect.

```rust
rt.provide_buffers(my_pool);
```

DESIGN.md
```diff
@@ -0,0 +1,756 @@
...
+
+The tokio-uring crate will provide its buffer type, `IoBuf`, to use when reading
+and writing.
+
+```rust
+pub struct IoBuf {
+    kind: Kind,
+}
+
+enum Kind {
```
seanmonstar, line 285: Is this just for concept purposes? Do you plan to instead actually use like a vtable design kind of like in `bytes`? Otherwise accessing the slice will require a match every single time, which means the performance downsides of SBO.

carllerche: Might eventually, but I am not worrying about it now. It could be changed to avoid the branch on deref.

carllerche: I mean, eventually, we might "just" use bytes::Bytes.

withoutboats:
Overall there's a lot here that's similar to `ringbahn`. I'll leave some notes about the differences.

ringbahn is specifically designed to be externally extensible in a number of ways:
1. A user of ringbahn can define how `io_uring_enter` gets called; this is done via what's called the `Drive` trait. ringbahn's code just requests from the implementation of `Drive` a certain number of SQEs, and then expects the drive implementer to somehow call ringbahn's `complete` function to wake tasks or clean up resources as events are completed. This is how ringbahn could allow both multi-threaded and single-threaded implementations (note that this means there is a `Mutex` internal to ringbahn, basically around its equivalent to this document's `Lifecycle` type, though it should be very low contention). Obviously, end users are not expected to implement this, but its designed to allow multiple other frameworks built on top of it to share its code.
2. A user of ringbahn can define their own operations; there's no centralized enums of all the possible operations as appears here in the `State` enum. This only really requires one modification to the core code: instead of storing the buffers in the `State` enum, the future types owns any memory associated with the operation & if its dropped, the equivalent of the `Ignored` enum takes ownership and holds a callback to clean up that state. This is not only beneficial because it the set of primitive ops supported by io-uring is frequently growing, but because it could in theory allow users of ringbahn to define their own complex multi-step "operations" that submit multiple SQEs that are linked together and then tying that state to the completion of the whole chain.
3. ringbahn's equivalent of TcpStream etc is not extensible in terms of buffer strategy (i.e. there's no realistic way to use automatic buffer selection with ringbahn), but it is supposed to be if I were able to work on it again in the future. So rather than enumerating the kinds of buffers, it would ideally support arbitrary buffers that implement some trait to be called at the right times. I'm certain this is possible, my dayjob keeps me away from completing the work.

My impression is extensibility in this manner has never been a priority for tokio, so I'm not surprised to see the differences. Its certainly easier if you don't do this.

The other big difference is that ringbahn's provided IO handles use AsyncBufRead instead of passing ownership in and out. I think any good implementation should support both (ringbahn doesn't provide the ownership passing API, but I believe it would be straightforward to implement on top of its existing APIs).

DESIGN.md
```diff
@@ -0,0 +1,756 @@
...
+The queue's single-producer characteristic optimizes for a single thread to own
+a given submission queue. Supporting a multi-threaded runtime requires either
+synchronizing pushing onto the submission queue or creating a queue pair per
+thread. The tokio-uring runtime uses an isolated thread-per-core model,
+differing from Tokio's other runtimes. Unlike Tokio's primary multi-threaded
+runtime, there is no work-stealing. Tasks remain on the thread that spawned them
+for the duration of their lifecycle. Each runtime thread will own a dedicated
+submission and completion queue pair, and operations are submitted using the
+submission queue associated with the current thread. Operation completion
+futures will not implement `Send`, guaranteeing that they remain on the thread
```
Diggsey, line 138: This would have quite a large impact on the ecosystem. Is there a possibility to make the futures be `Send`, but only incur a significant performance penalty if polled on another thread?

For example, the source of the `!Send` bound appears to be the necessity to access the driver state when polling:
https://github.com/tokio-rs/tokio-uring/blob/master/src/driver/op.rs#L15

I could imagine a solution where the `Op` stores an `Arc<MagicCell<driver::Inner>>`. The idea behind `MagicCell` is that it would perform a runtime `Send` check (ie. check that the accessing thread is the same as the thread it was created on) before allowing access to the interior.

If this check fails, the `Op` would do something like this pseudocode:
```
let (tx, rx) = Proxy::new();
let original_op = mem::replace(self, rx);

original_thread.send(original_op.and_then(tx));
```

As long as tasks are fairly sticky, then the overhead would only be incurred for in-progress ops when a task is actually moved. Of course, there's the question of whether the overhead of the `Arc` and `MagicCell` would outweigh the benefits.

Diggsey:
How do you see this being used? Is the idea that application authors would directly build on top of `tokio-uring` instead of `tokio`?

If so, would it make sense to split the new public API (and associated execution model) from the underlying uring-based implementation, so that applications are still building against a platform-independent interface? (Even if the only back-end is linux specific right now)

AIUI io-uring is basically identical to IO completion ports on windows, with the innovation that operations can be submitted without a system-call, so it's plausible that there would be more than one "backend" that would benefit from this execution model. And it would also be desirable to have a fallback implementation based on mio for platforms that cannot support io-uring or other backend directly.

nikomatsakis (changes requested):
one quick nit

Kestrer:
io_uring is similar to IOCP in that they're both completion-based, but it has one major difference that makes it really hard to abstract over both: the threading model. io_uring naturally lends itself to a ring-per-thread model where each thread is self-contained and manages its own I/O. For Rust, this means thread-local runtimes, little synchronization and lots of `!Send` types. On the other hand IOCP is designed to work in work-stealing systems: there is typically one global IOCP instance that executes all I/O, and wakes up an essentially arbitrary thread when a completion status is received. For Rust, this means a single multithreaded work-stealing runtime, lots of synchronization and `Send` types.

One way to reconcile this difference is to always use a runtime-per-thread model, so on Windows there would be an IOCP instance for each thread (I don't know enough about IOCP to say whether this is a bad idea). Alternatively, there could be one global io_uring instance protected by a mutex, or an io_uring instance for each thread where submission queues are shared with a mutex but completion queues are thread-local.

I really don't know which solution is best; they all have disadvantages when run on platforms that don't natively support the model. I think in an ideal world, io_uring would support multiple submission/completion queues on a single `io_uring` instance so we can use a multithreaded work stealing model everywhere, but that might not ever happen.

DESIGN.md
```diff
@@ -0,0 +1,756 @@
...
+
+```rust
+my_tcp_stream.fill_buf().await?;
+let buf: IoBuf = my_tcp_stream.take_read_buf();
+
+// Mutate `buf` if needed here.
+
+my_other_stream.place_write_buf(buf);
+my_other_stream.flush().await?;
+```
```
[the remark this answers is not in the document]
carllerche: Yes via inherent methods on relevant types. It should fit into the model proposed here.

DESIGN.md
```diff
@@ -0,0 +1,756 @@
...
+    
+    async fn write_at(&self, buf: buf::Slice, pos: u64) -> BufResult<usize>;
+}
+
+/// The caller allocates a buffer
+let buf = buf::IoBuf::with_capacity(4096);
+
+// Read the first 1024 bytes of the file, when `std::ops::Try` is
+// stable, the `?` can be applied directly on the `BufResult` type.
+let BufResult(res, buf) = file.read_at(0, buf.slice(0..1024)).await;
```
pickfire, line 210: Is there any reason why we don't use `Index` here rather than having a separate slice function? We could do like `buf[0..1024]` and both should panic when the start or end index is out of the capacity.

raftario:
I feel like keeping io_uring and IOCP runtimes separate would be the best choice. There's no question about a thread per core architecture being a better fit for io_uring, and while IOCP can be used that way too (and pretty efficiently afaik if you create the completion port with a concurrency value of one), it would still be a bunch of additional code to maintain and logic to generalise. Note that IOCP is also typically used with the threadpool provided by the Windows kernel itself, as [illustrated by the .NET TPL](https://github.com/dotnet/runtime/blob/main/src/coreclr/vm/win32threadpool.cpp) (even though [that will change in .NET 6](https://github.com/dotnet/runtime/pull/43841)).

The way I see it, "vanilla" tokio already does an amazing job at being an efficient and portable runtime. If you're willing to pay the complexity cost of using a thread per core architecture and dealing with completion based IO for the extra performance, your probably already know where that code is going to run. Separate, targeted runtimes seem like the obvious choice in that scenario, especially since depending on io_uring itself isn't particularly portable (a portable runtime without support for BSD and Darwin, and 90% of Linux machines out there (just throwing random numbers here), doesn't sound like the best way to build portable stuff).

DESIGN.md
```diff
@@ -0,0 +1,756 @@
...
+    // Resources referenced by the kernel, this must stay
+    // available until the operation completes.
+    state: State,
+    lifecycle: Lifecycle,
+}
+
+enum State {
+    // An in-flight read operation, `None` when reading into
+    // a kernel-owned buffer pool
+    Read { buffer: Option<buf::Slice> },
```
pickfire, line 576: Is it better to have two different State variations where we have a generic for kernel-owned buffer pool and Vec (if they are always the same)? Will one state be `None` and another state be `Some` for `Read`?

DESIGN.md
Kestrer: It's probably because `Index` is required to return a reference but `.slice` has to return an owned value.

DESIGN.md
```diff
@@ -0,0 +1,756 @@
...
+let buf = buf::IoBuf::with_capacity(4096);
+
+// Read the first 1024 bytes of the file, when `std::ops::Try` is
+// stable, the `?` can be applied directly on the `BufResult` type.
+let BufResult(res, buf) = file.read_at(0, buf.slice(0..1024)).await;
+// Check the result.
+res?;
+
+// Write some data back to the file.
+let BufResult(res, _) = file.write_at(1024, buf.slice(0..512).await;
```
pythondude325, line 215: Missing a parenthesis here
```suggestion
let BufResult(res, _) = file.write_at(1024, buf.slice(0..512)).await;
```

DESIGN.md
```diff
@@ -0,0 +1,636 @@
...
+```rust
+let buf = IoBuf::with_capacity(4096);
+
+// This read will be canceled.
+select! {
+    _ = my_tcp_stream.read(buf.slice(..)) => unreachable!(),
+    _ = future::ready(()) => {}
+}
+
+let buf = IoBuf::with(capacity(4096))
```
Alphix: Shouldn't that be with_capacity?

DESIGN.md
```diff
@@ -0,0 +1,756 @@
...
+
+// Positional read and write function definitions
+impl File {
+    async fn read_at(&self, buf: buf::Slice, pos: u64) -> BufResult<usize>;
+    
+    async fn write_at(&self, buf: buf::Slice, pos: u64) -> BufResult<usize>;
+}
+
+/// The caller allocates a buffer
+let buf = buf::IoBuf::with_capacity(4096);
```
Alphix, line 206: How about letting IoBuf::with_capacity return a `Result` so that it can support [fallible allocations](https://rust-lang.github.io/rfcs/2116-alloc-me-maybe.html) with having to provide separate `try_*` methods?

DESIGN.md
```diff
@@ -0,0 +1,756 @@
...
+```rust
+let listener = Arc::new(TcpListener::bind(...));
+
+spawn_on_each_thread(async {
+    loop {
+        // Wait for this worker to have capacity. This
+        // ensures there are a minimum number of workers
+        // in the runtime that are flagged as with capacity
+        // to avoid total starvation.
+        current_worker::wait_for_capacity().await;
```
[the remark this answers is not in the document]
stuhood: Does this imply that "worker" is a new noun in this API? If so, is `spawn_on_each_thread` more like "spawn on each worker?" Or should this be `current_thread::wait_for_capacity` instead?

But also, if the implication is that the only way to saturate your runtime is to always be accepting connections on all threads that are dedicated to it, should `spawn_on_each_thread` take a `FnMut() -> impl Future` instead to remove the `loop`? The `loop` seems error prone, because if it ever exited (due to panic/etc) you'd have an idle thread spawned by the runtime, whereas with a function the runtime could restart threads.

DESIGN.md
```diff
@@ -0,0 +1,756 @@
...
+        
+        let socket = listener.accept().await;
+        
+        spawn(async move { ... });
```
stuhood, line 171: The `socket: TcpStream` is not itself a uring "operation", afaict, so it could be `Send`. Does that imply that load balancing could still happen between accepting connections and servicing them?

Load balancing of tasks to threads would then happen during `spawn`. You would `spawn` N tasks to `accept` (having all threads `accept` seems like overkill?), and upon `accept` they would `spawn` new tasks (which would likely go to other threads) to service the connections. The runtime would need to decide which thread to target each time it spawned a task, which could be tricky (long lived cheap tasks vs short lived expensive tasks, etc).

From an API perspective, the task would still be `!Send`, but the `spawn` function _would_ require `Send` for its argument(s) in order to move the inputs to the task onto its permanent home thread.

----

This question is particularly important I think because the effect of `current_worker::wait_for_capacity` seems to be that users end up creating their own thread pools (i.e. "a pool of threads fed by a queue") with the assistance of the `spawn_on_each_thread` function. In the accept case, the `accept` call is what waits on your queue: but other usecases that spawn work recursively would need an explicit queue of work after their `wait_for_capacity` call (if they actually wanted to balance across all threads that were available, rather than having recursion pinned at the top level).

It's not clear that any extra flexibility provided by this API justifies having applications need to implement their own load-balancing around `spawn` calls.

stuhood: Ah, sorry. Just noticed that this is addressed at the end in https://github.com/tokio-rs/tokio-uring/blob/design-doc/DESIGN.md#load-balancing-spawn-function . The idea that you might have load-balanced vs non-load-balanced spawn functions is interesting.

DESIGN.md
```diff
@@ -0,0 +1,756 @@
...
+`IoBuf::with_capacity` or checking-out an entry from a pool. 
+
+```rust
+// Individually heap-allocated
+let my_buf = IoBuf::with_capacity(4096);
+
+// Checked-out from a pool
+match my_buffer_pool.checkout() {
+    Ok(io_buf) => ...,
+    Err(e) => panic!("buffer pool empty"),
```
jonhoo, line 305: If "the buffer pool is empty" is the only possible error here, I'd prefer `Option` over `Result`, since it doesn't really feel like an error condition.

DESIGN.md
```diff
@@ -0,0 +1,756 @@
...
+}
+```
+
+On drop, if the `IoBuf` is a buffer pool member, it is checked back in. If the
+kernel initially checked out the buffer as part of a read operation, an io-uring
+operation is issued to return it. Submitting the io-uring operation requires the
+buffer to remain on the same thread that checked it out and is enforced by
+making the `IoBuf` type `!Send`.
```
jonhoo, line 313: Does this also mean that the pools themselves are `!Send`? I suspect not, but it wasn't clear from this paragraph.

DESIGN.md
```diff
@@ -0,0 +1,756 @@
...
+let mut buf = IoBuf::with_capacity(4096);
+let slice = buf.slice(0..100);
+
+assert_eq!(slice.len(), 0);
+assert_eq!(slice.capacity(), 100);
+
+// Read data from a file into the buffer
+let BufResult(res, slice) = my_file.read_at(slice, 0);
+
+assert_eq!(slice.len(), 100);
```
jonhoo, line 347: This is nitpicking, but this assert seems overly aggressive given that the file may contains fewer than 100 bytes.

DESIGN.md
```diff
@@ -0,0 +1,756 @@
...
+sufficient to satisfy safety.
+
+## Closing resources
+
+Idiomatically with Rust, closing a resource is performed in the drop handler,
+and within an asynchronous context, the drop handler should be non-blocking.
+Closing an io-uring resource requires canceling any in-flight operations, which
+is an asynchronous process. Consider an open TcpStream associated with
+file-descriptor (FD) 10. A task submits a read operation to the kernel, but the
+`TcpStream` is dropped and closed before it sees it. A new `TcpStream` is
```
jonhoo, line 381: > before _the kernel_ sees it

?

DESIGN.md
carllerche: I "know" the file contains more than 100 bytes :laugh:

DESIGN.md
```diff
@@ -0,0 +1,636 @@
...
+synchronizing pushing onto the submission queue or creating a queue pair per
+thread. The tokio-uring runtime uses an isolated thread-per-core model,
+differing from Tokio's other runtimes. Unlike Tokio's primary multi-threaded
+runtime, there is no work-stealing. Tasks remain on the thread that spawned them
+for the duration of their lifecycle. Each runtime thread will own a dedicated
+submission and completion queue pair, and operations are submitted using the
+submission queue associated with the current thread. Operation completion
+futures will not implement `Send`, guaranteeing that they remain on the thread
+to receive the operation result. Interestingly, the resources, e.g.,
+`TcpListener`, can be `Send` as long as they do not hold operation futures
```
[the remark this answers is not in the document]
carllerche: I added a paragraph in "future work" mentioning registering file descriptors and how the `Send` aspect impacts it.

DESIGN.md
```diff
@@ -0,0 +1,636 @@
...
+///
+/// This is implemented as a new type to implement std::ops::Try once
+/// the trait is stabilized.
+type BufResult<T> = (std::io::Result<T>, buf::Slice);
+
+// Positional read and write function definitions
+impl File {
+    async fn read_at(&self, buf: buf::Slice, pos: u64) -> BufResult<usize>;
+    
+    async fn write_at(&self, buf: buf::Slice, pos: u64) -> BufResult<usize>;
```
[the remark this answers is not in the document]
carllerche: `Bytes` is possible, but probably better supported as an `IoBuf` backend. Supporting a trait is *possible* but adds complexity/overhead as the trait would probably need to be boxed into a trait object and stored internally while the operation is in-flight.

DESIGN.md
```diff
@@ -0,0 +1,636 @@
...
+operation future. The runtime loads the in-flight operation state, stores the
+result, transitions the lifecycle to `Completed`, and notifies the waker. The
+next time the caller's task executes, it polls the operation's future which
+completes and returns the stored result.
+
+If the operation's future drops before the operation completes, it sets the
+lifecycle to `Ignored` and submits a cancellation request to the kernel. The
+cancellation request will attempt to terminate the operation, causing it to
+complete immediately with an error. Cancellation is best-effort; the operation
+may or may not terminate early. If the operation does complete, the runtime
```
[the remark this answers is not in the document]
carllerche: :+1: I don't *think* I need to update anything in the design doc to reflect this, but yes :)

DESIGN.md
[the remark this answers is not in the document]
carllerche: It could be `current_thread`, this snippet is just to illustrate the concept and not to include within tokio-uring. Either way, we can explore the options outside of the crate.

DESIGN.md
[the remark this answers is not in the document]
carllerche: Yes, `result()?` is better.

DESIGN.md
```diff
@@ -0,0 +1,756 @@
...
+        let socket = listener.accept().await;
+        
+        spawn(async move { ... });
+    }
+})
+```
+
+## Reading and writing
+
+Read and write operations require passing ownership of buffers to the kernel.
```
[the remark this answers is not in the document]
carllerche: :+1: you are correct.  I do cover the requirement to keep the file descriptor open later. I will keep this section focused on buffers.

DESIGN.md
carllerche: Odds are, `tokio-uring` will not be used when implementing a work-stealing backed scheduler on top. At that point, it would make more sense to add support for io-uring within tokio proper.

DESIGN.md
carllerche: yes, thanks

DESIGN.md
carllerche: It is *possible* but, part of the goal for tokio-uring, is to provide a synchronization-free option.

Related to the "ecosystem", there already are `!Send` options out there (e.g. glommio), so it already is an issue to deal with.

DESIGN.md
```diff
@@ -0,0 +1,756 @@
...
+argument.
+
+```rust
+async fn read<T: AsMut<[u8]>>(&self, dst: T) { ... }
+
+struct MyUnsableBuf {
+    mem: [u8; 10],
+}
+
+impl AsMut<[u8]> {
```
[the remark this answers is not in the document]
carllerche: Woops, thanks. It should be `MyUnstableBuf`. I will fix this :)

DESIGN.md
carllerche: It might be, but before making that tweak I would like to measure the impact. It doesn't materially change the design though so we can explore it in the future.

DESIGN.md
carllerche: Yes, thanks for the catch.

DESIGN.md
carllerche: I assume you mean "without having to provide separate `try_*` methods? I would default to following `std`'s lead here w/ `Vec`. The question is whether or not supporting fallible allocation should be the primary use case.

DESIGN.md
```diff
@@ -0,0 +1,756 @@
...
+The resource must still tolerate the caller dropping it without being explicitly
+closed. In this case, tokio-uring will close the resource in the background,
+avoiding blocking the runtime. The drop handler will move ownership of the
+resource handle to the runtime and submit cancellation requests for any
+in-flight operation. Once all existing in-flight operations complete, the
+runtime will submit a close operation.
+
+If the drop handler must process closing a resource in the background, it will
+notify the developer by emitting a warning message using [`tracing`]. In the
+future, it may be possible for Rust to provide a `#[must_not_drop]` attribute.
```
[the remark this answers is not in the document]
carllerche: You are correct re: panicking, which means drop must always be safe and correct, though not necessarily ideal. The idea of `#[must_not_drop]` is that it would be a lint and not enforced.

Just like with `#[must_use]` lints, if the developer means to drop and let cleanup happen in the background, they can silence the error (via some TBD mechanism).

DESIGN.md
```diff
@@ -0,0 +1,756 @@
...
+to the future via the `Completed` variant.
+
+When a task starts a read operation, the runtime allocates an entry in the
+in-flight operation store, storing the buffer and initializing the lifecycle to
+`Submitted`. The runtime then pushes the operation to the submission queue but
+does not synchronize it with the kernel. Synchronization happens once the task
+yields to the runtime, enabling the task to submit multiple operations without
+synchronizing each one. Delaying synchronization can add a small amount of
+latency. While not enforced, tasks should execute for no more than 500
+microseconds before yielding.
```
jonhoo, line 618: `coop` might be able to help here.

DESIGN.md
carllerche: Perhaps, there are conflicting patterns in std. I don' have a huge opinion there.

DESIGN.md
carllerche: They are `!Send` as they are registered w/ a specific uring.

DESIGN.md
```diff
@@ -0,0 +1,756 @@
...
+
+Dropping the first read operation cancels it, and because it never completes,
+the second read operation should read the first packet of data from the TCP
+stream. However, the kernel may have already completed the first read operation
+before seeing the cancellation request. A naive runtime implementation would
+drop the completion result of any canceled operation, which would result in
+losing data. Instead, the runtime could preserve the data from the first read
+operation and return it as part of the second read. The process would not lose
+data, but the runtime would need to perform an extra copy or return the caller a
+different buffer than the one it submitted.
```
jonhoo, line 727: I think it'd be useful here to explain why this isn't a problem with the proposed API.

DESIGN.md
jonhoo: My general observation has been that `Option` should be used if the alternative case a) carries no information; b) is common, not exceptional; and c) is unlikely to later be propagated as an `Err`. I _think_ all of those apply here, and therefore that `Option` is the better choice. The exception would be if you anticipate that `checkout()` can fail for _different_ reasons as well. But even in that case, I'd prefer `Result<Option<Pool>, E>`.

DESIGN.md
jonhoo: It'd be worthwhile to call that out explicitly in the doc I think.

DESIGN.md
carllerche: Done

DESIGN.md
```diff
@@ -726,6 +728,10 @@ operation and return it as part of the second read. The process would not lose
 data, but the runtime would need to perform an extra copy or return the caller a
 different buffer than the one it submitted.
 
+
+The proposed API is not vulnerable to this issue as resources track their
+operations, preventing them from being dropped as long as the resource is open.
```
jonhoo, line 733: I don't think that's the reason this problem wouldn't occur? If someone dropped a call to `consume` in a select, what would happen?

carllerche: consume is a future, but it is not an **operation** future.

jonhoo: Sure, but dropping the `consume` (just like dropping the `read`) might still lead to bytes being lost, no?

DESIGN.md
```diff
@@ -0,0 +1,773 @@
...
+    .timeout(Duration::from_millis(100))
+    .await?
+```
+
+The cancel and timeout function are inherent methods on operation future types.
+
+If the operation's future drops before the operation completes, the runtime will
+submit a cancellation request for the operation. However, cancellation is
+asynchronous and best-effort, the operation may still complete. In that case,
+the runtime discards the operation result.
```
gardnervickers, line 127: Does this imply that the runtime will also discard/close resources which were created by the kernel? For example `openat` or a `read` with registered buffers might result in an open FD or a buffer removed from the buffer pool. I'm guessing the runtime will need to look at the `State` enum associated with the cancelled request to figure out the request type, and use that to figure out if the `cqueue::Entry` has some resources which need to be cleaned up?

DESIGN.md
```diff
@@ -0,0 +1,773 @@
+# Summary
+
+The RFC proposes a new asynchronous Rust runtime backed by [io-uring] as a new
+crate: tokio-uring. The API aims to be as close to idiomatic Tokio, but will
+deviate when necessary to provide full access to io-uring's capabilities. It
+also will be compatible with existing Tokio libraries. The runtime will use an
+isolated thread-per-core model, and many types will be `!Send`.
+
+[io-uring]: https://kernel.dk/io_uring.pdf?source=techstories.org
```
skyzh, line 9: Is `?source=techstories.org` intended here?

DESIGN.md
```diff
@@ -0,0 +1,773 @@
...
+    
+    async fn write_at(&self, buf: buf::Slice, pos: u64) -> BufResult<usize>;
+}
+
+/// The caller allocates a buffer
+let buf = buf::IoBuf::with_capacity(4096);
+
+// Read the first 1024 bytes of the file, when `std::ops::Try` is
+// stable, the `?` can be applied directly on the `BufResult` type.
+let BufResult(res, buf) = file.read_at(0, buf.slice(0..1024)).await;
```
skyzh, line 210: I have a question on this. What if a future is canceled without being polled to complete? In this case,
* The kernel is still handling the request.
* The slice is dropped (as part of the future). (Not sure about this, point me out if I'm wrong)
* When kernel completes the request, memory reference of this slice might be no longer valid, which causes a segment fault.

skyzh: I've got the answer in this proposal. Thanks for your work!

> In this case, tokio-uring will close the resource in the background, avoiding blocking the runtime. 

But it seems that this part has not been implemented yet. Current implementation might block the runtime by using a spin loop. https://github.com/tokio-rs/tokio-uring/blob/master/src/driver/op.rs#L210

DESIGN.md
```diff
@@ -0,0 +1,773 @@
...
+for the duration of their lifecycle. Each runtime thread will own a dedicated
+submission and completion queue pair, and operations are submitted using the
+submission queue associated with the current thread. Operation completion
+futures will not implement `Send`, guaranteeing that they remain on the thread
+to receive the operation result. Interestingly, the resources, e.g.,
+`TcpListener`, can be `Send` as long as they do not hold operation futures
+internally. It is possible to have two operations in-flight from a single
+resource associated with different queues and threads.
+
+Because operations are not `Send`, tasks awaiting these operations are also not
```
bIgBV, line 144: Nit: I "not send" should have a consistent symbol across the doc, and I personally think using `!Send` makes it easier to read.

DESIGN.md
```diff
@@ -0,0 +1,773 @@
...
+    .buffer_size(16_384)
+    .num_buffers(256)
+    .build();
+
+// Create the runtime
+let mut rt = tokio_uring::runtime::Runtime::new()?;
+
+// Provide the buffer pool to the kernel. This passes
+// ownership of the pool to the kernel.
+let pool_token = rt.provide_buffers(my_pool)?;
```
bIgBV, line 261: Maybe the runtime should have a method for taking a list of pools, something like

```rust
let mut rt = tokio_uring::runtime::Runtime::new().
    .with_pools([netpool, fspool,...])
    .build()?;
```

Or

```rust
let mut rt = tokio_uring::runtime::Runtime::new().
    .with_pool(netpool)
    .with_pool(fspool)
    //...
    .build()?;
```

kaimast:
> The way I see it, "vanilla" tokio already does an amazing job at being an efficient and portable runtime. If you're willing to pay the complexity cost of using a thread per core architecture and dealing with completion based IO for the extra performance, you probably already know where that code is going to run. Separate, targeted runtimes seem like the obvious choice in that scenario, especially since depending on io_uring itself isn't particularly portable (a portable runtime without support for BSD and Darwin, and 90% of Linux machines out there (just throwing random numbers here), doesn't sound like the best way to build portable stuff).

I think there's a very common use case where you have a client-server architecture and want to leverage `io_uring` on the server-side. 
Unless I am missing something, having separate runtimes in this case would make it really hard to share code between client and server.

DESIGN.md
```diff
@@ -0,0 +1,773 @@
...
+[io-uring]: https://github.com/tokio-rs/io-uring/
+
+## Future work
+
+The main Tokio crate will likely adopts concepts from tokio-uring in the future.
+The most obvious area is Tokio's file system API, currently implemented using a
+thread-pool. The current API would remain, and the implementation would use
+io-uring when supported by the operating system. The tokio-uring APIs may form
+the basis for a Tokio 2.0 release, though this cannot happen until 2024 at the
+earliest. As an intermittent step, the tokio-uring crate could explore
```
shikhar, line 672: `s/intermittent/interim/` :)

DESIGN.md
mzabaluev: There needs to be a way to identify a pool (i.e. the buffer group in io-uring API) to read operations that use it.

DESIGN.md
```diff
@@ -0,0 +1,773 @@
...
+### Expose a raw io-uring operation submission API
+
+The proposed tokio-uring API does not include a strategy for the user to submit
+custom io-uring operations. Any raw API would be unsafe or would require the
+runtime to support taking opaque data as a trait object. Given that io-uring has
+a well-defined set of supported options, tokio-uring opts to support each
+operation explicitly. The application also can create its own set of io-uring
+queues using the io-uring crate directly.
+
+### Load-balancing spawn function
```
inevity, line 744: About load balance, If we do clientside load balancing, the server tokio-uring should support which feature?   We should mention is .

DESIGN.md
```diff
@@ -0,0 +1,773 @@
...
+
+If the operation's future drops before the operation completes, the runtime will
+submit a cancellation request for the operation. However, cancellation is
+asynchronous and best-effort, the operation may still complete. In that case,
+the runtime discards the operation result.
+
+The queue's single-producer characteristic optimizes for a single thread to own
+a given submission queue. Supporting a multi-threaded runtime requires either
+synchronizing pushing onto the submission queue or creating a queue pair per
+thread. The tokio-uring runtime uses an isolated thread-per-core model,
```
ileixe, line 132: Does 'thread-per-core' model here imply CPU affinity as well? I wonder `sched_setaffinity()` will be planned or not as I could not see related implementation code yet.

Noah-Kennedy (approved):
Approving to get this in. We can have further discussion in new PRs to make changes.

PR #43, opened 2021-08-09
buf: impl IoBuf/IoBufMut for bytes::Bytes/BytesMut

### LINKED ISSUE #42: Impl `IoBuf`/`IoBufMut` for `bytes::Bytes`/`BytesMut`
It will be very useful if implement `IoBuf`/`IoBufMut` for [`bytes::Bytes`](https://docs.rs/bytes/1.0.1/bytes/struct.Bytes.html)/[`BytesMut`](https://docs.rs/bytes/1.0.1/bytes/struct.BytesMut.html).
[`bytes`](https://crates.io/crates/bytes) can be an optional dependency.

songzhi:
import bytes as optional dependency.

Fixes #42

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -22,6 +22,7 @@ scoped-tls = "1.0.0"
 slab = "0.4.2"
 libc = "0.2.80"
 io-uring = { version = "0.5.0", features = [ "unstable" ] }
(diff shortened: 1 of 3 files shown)


carllerche (approved):
Looks good, thanks!

PR #44, opened 2021-08-09
fs: impl `File::read_exact_at` and `File::write_all_at`

### LINKED ISSUE #41: Impl `File::read_exact_at` and `File::write_all_at`
Same as [`std::io::Read::read_exact`](https://doc.rust-lang.org/std/io/trait.Read.html#method.read_exact) and [`std::io::Write::write_all`](https://doc.rust-lang.org/std/io/trait.Write.html#method.write_all).

Here's my simple solution:
```rust
impl File {
    pub async fn read_exact_at<T: IoBufMut>(&self, mut buf: T, mut pos: u64) -> BufResult<(), T> {
        let mut bytes_read = 0;
        let buf_len = buf.bytes_init();
        while bytes_read < buf_len {
            let slice = buf.slice(bytes_read..);
            let (res, slice) = self.read_at(slice, pos).await;
            buf = slice.into_inner();
            match res {
                Ok(0) => {
                    return (
                        Err(io::Error::new(
                            io::ErrorKind::UnexpectedEof,
                            "failed to fill whole buffer",
                        )),
                        buf,
                    )
                }
                Ok(n) => {
                    bytes_read += n;
                    pos += n as u64;
                }
                Err(ref e) if e.kind() == io::ErrorKind::Interrupted => {}
                Err(e) => return (Err(e), buf),
            };
        }

        (Ok(()), buf)
    }

    pub async fn write_all_at<T: IoBuf>(&self, mut buf: T, mut pos: u64) -> BufResult<(), T> {
        let mut bytes_written = 0;
        let buf_len = buf.bytes_init();
        while bytes_written < buf_len {
            let slice = buf.slice(bytes_written..);
            let (res, slice) = self.write_at(slice, pos).await;
            buf = slice.into_inner();
            match res {
                Ok(0) => {
                    return (
                        Err(io::Error::new(
                            io::ErrorKind::WriteZero,
                            "failed to write whole buffer",
                        )),
                        buf,
                    )
                }
                Ok(n) => {
                    bytes_written += n;
                    pos += n as u64;
                }
                Err(ref e) if e.kind() == io::ErrorKind::Interrupted => {}
                Err(e) => return (Err(e), buf),
            };
        }

        (Ok(()), buf)
    }
}
```

songzhi:
Current API and abstractions make it possible to write these helper methods.

Fixes #41


src/fs/file.rs
```diff
@@ -170,6 +170,89 @@ impl File {
...
+    /// ```
+    ///
+    /// [`ErrorKind::Interrupted`]: std::io::ErrorKind::Interrupted
+    /// [`ErrorKind::UnexpectedEof`]: std::io::ErrorKind::UnexpectedEof
+    pub async fn read_exact_at<T: IoBufMut>(
+        &self,
+        mut buf: T,
+        pos: u64,
+    ) -> crate::BufResult<(), T> {
+        let buf_len = buf.bytes_init();
```
carllerche, line 228: I think we want to use `bytes_total()` here to read in uninitialized memory? Though, this would change the API. E.g. passing in `Vec::with_capacity(100)` would read 100 bytes.

The `slice(..)` API could be used for limiting the length to read.

carllerche (changes requested):
Looks good, thanks. I had one thought inline.

src/fs/file.rs
songzhi: I did recognize that the `bytes_total()` instead of `bytes_init()` is used in the read operation in `tokio-uring`. But the `std::io::Read::read_exact`'s parameter is a `&mut [u8]`, whose length is immutable.
Maybe it's time to choose  consistency (in this crate) or conveyance  (between std and other libraries).
I think it's good to keep consistency in this crate and inform users the differences.

tsturzl:
Just giving this a bump. I'm really interested in using the feature.

tests/fs_file.rs
```diff
@@ -33,6 +40,17 @@ fn basic_read() {
     });
 }
 
+#[test]
```
Noah-Kennedy, line 43: Might wanna try this with some really big generated array of bytes, possibly with something like quickcheck, just to be sure.

Noah-Kennedy (changes requested):

Noah-Kennedy (approved):

src/fs/file.rs
```diff
@@ -171,6 +171,89 @@ impl File {
...
+    /// ```
+    ///
+    /// [`ErrorKind::Interrupted`]: std::io::ErrorKind::Interrupted
+    /// [`ErrorKind::UnexpectedEof`]: std::io::ErrorKind::UnexpectedEof
+    pub async fn read_exact_at<T: IoBufMut>(
+        &self,
+        mut buf: T,
+        pos: u64,
+    ) -> crate::BufResult<(), T> {
+        let buf_len = buf.bytes_total();
```
jiangliu, line 229: Should we check for wrapping around of `buf_len + pos` here?

src/fs/file.rs
```diff
@@ -222,6 +305,78 @@ impl File {
...
+    ///         file.close().await?;
+    ///         Ok(())
+    ///     })
+    /// }
+    /// ```
+    ///
+    /// [`write_at`]: File::write_at
+    /// [`ErrorKind::Interrupted`]: std::io::ErrorKind::Interrupted
+    pub async fn write_all_at<T: IoBuf>(&self, mut buf: T, pos: u64) -> crate::BufResult<(), T> {
+        let buf_len = buf.bytes_init();
```
jiangliu, line 352: validate `pos + buf_len` here too:)

src/fs/file.rs
songzhi: Not sure about this, is there any similar checks in other functions?

src/fs/file.rs
```diff
@@ -222,6 +316,89 @@ impl File {
...
+                            "failed to write whole buffer",
+                        )),
+                        buf,
+                    )
+                }
+                Ok(n) => {
+                    bytes_written += n;
+                }
+                Err(ref e) if e.kind() == io::ErrorKind::Interrupted => {}
+                Err(e) => return (Err(e), buf),
```
bnoordhuis, line 395: Drive-by review: wouldn't it be more appropriate to return success if `bytes_written > 0`? Returning an error means the caller loses track of how much data was written out. In some scenarios that might be a security liability.

Same line of reasoning applies to the `Ok(0)` case.

songzhi: It's same as `std::fs::File::write_all` implementation, so I don't think this is a problem.

littledivy (approved):

Noah-Kennedy (approved):

Noah-Kennedy:
@songzhi can you resolve the conflicts? I'll merge once that's done.

Noah-Kennedy (approved):

tests/fs_file.rs
FrankReh: @Noah-Kennedy Is this still a concern?

src/fs/file.rs
FrankReh: Let's resolve this unless someone has an example in std or tokio where such a wrap around check is made?

FrankReh (approved):
I'm going to say, let's try to get this committed. It's been out a long time.

src/fs/file.rs
FrankReh: @carllerche Does the current form of the code satisfy your concern from Aug 2021?

FrankReh:
@carllerche Waiting for your okay. You had requested changes so the page wants your blessing.

src/fs/file.rs
FrankReh: @Noah-Kennedy Looking at the latest version of this code, it does use bytes_total() for the read and only the bytes_init() for the write. So I think the concern was taken to heart. Want to have this PR ignore a further review by @carllerche? Then it can be merged.

PR #51, opened 2021-08-26
fix panic when sq is full

ihciah:
In `src/fs/file.rs:169`, it assume the `read_at` must be success. However, in `src/driver/op.rs:70` it may fail.

In my experiment, when under high presure, the sq may be full which will cause the `submit` returns ResourceBusy. In this situation maybe we should retry.


jiangliu:
Seems this issue has been solved in the latest code, so should we close this PR?

PR #52, opened 2021-09-05
Fix panic in Deref/DerefMut for Slice extending into uninitialized part of the buffer

### LINKED ISSUE #45: [Bug] Wrong `Slice::deref/deref_mut` implementation
Here's a minimal reproducible example:
```rust
let v = Vec::with_capacity(32);
let slice = v.slice(..);
let x = &*slice;
// raise error 'range end index 32 out of range for slice of length 0'
```

The reason is that `Slice` creation uses `bytes_total()`, `Slice::deref` uses `bytes_init()`.
The module level `deref/deref_mut` or the `Slice::deref/deref_mut` implementation, I'm not sure which one should be modified.

https://github.com/tokio-rs/tokio-uring/blob/2c9fbeab9694b74a21588ff398767ee8b5619032/src/buf/io_buf.rs#L79-L83
https://github.com/tokio-rs/tokio-uring/blob/2c9fbeab9694b74a21588ff398767ee8b5619032/src/buf/mod.rs#L16-L20
https://github.com/tokio-rs/tokio-uring/blob/2c9fbeab9694b74a21588ff398767ee8b5619032/src/buf/slice.rs#L126-L132

mzabaluev:
The `Deref` and `DerefMut` impls for `Slice` did not ensure that the end of the returned byte slice is clipped to the end of initialized data in the buffer.

Fixes #45.

### DIFF
--- a/src/buf/slice.rs
+++ b/src/buf/slice.rs
@@ -165,26 +165,3 @@ unsafe impl<T: IoBufMut> IoBufMut for Slice<T> {
         self.buf.set_init(self.begin + pos);
     }
 }
-
-#[cfg(test)]
-mod tests {
-    use crate::buf::{IoBuf, IoBufMut};
-    use std::mem;
-
-    #[test]
-    fn can_deref_slice_into_uninit_buf() {
-        let buf = Vec::with_capacity(10).slice(..);
-        assert!(buf[..].is_empty());
-        let _ = buf.stable_ptr();
-        assert_eq!(buf.bytes_init(), 0);
-        assert_eq!(buf.bytes_total(), 10);
-
-        let mut v = Vec::with_capacity(10);
-        v.push(42);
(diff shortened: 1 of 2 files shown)


dignifiedquire:
Bump, seems like an important fix to merge

meyerzinn:
Didn't see this, but it is the root cause of #82

Noah-Kennedy (approved):

FSMaxB:
I just ran into this. @Noah-Kennedy can you please make a release that incorporates that fix?

FrankReh:
> I just ran into this. @Noah-Kennedy can you please make a release that incorporates that fix?

We should put a new release out very soon.

FrankReh:
@meyerzinn 
> Didn't see this, but it is the root cause of #82

With this merged, are you comfortable closing #82 as you are its author?

PR #53, opened 2021-09-07
Improve the ergonomics of .slice with IntoSlice trait

mzabaluev:
The owned slice API has a flaw: `b"hello".to_vec().slice(1..).slice(..3)` produces a
`Slice<Slice<Vec<u8>>>`.
It is easy to inadvertently end up with nested Slice wrappers, which adds unnecessary overhead to the buffer accessor methods and makes the results of `.begin()` and `.end()` difficult to interpret.

Move the `slice` method to a new trait `IntoSlice`, which is used to uniformly convert buffer parameters to a `Slice<T>`, with the impl of `InfoSlice` for `Slice<T>` closing onto `Slice<T>`.

Instead of `IoBuf` which can no longer be implemented for `Slice`, the buffer parameter type in the generic I/O operation methods is bound with `IntoSlice` to keep the convenient polymorphism, while reducing code bloat in monomorphization when the application passes both buffers and their slices to tokio-uring operations. Consequently, all buffers returned from the operations are wrapped into `Slice`.

This PR incorporates #52 in order to get tests to pass.


src/buf/slice.rs
```diff
@@ -127,36 +172,135 @@ impl<T: IoBuf> ops::Deref for Slice<T> {
...
 
-unsafe impl<T: IoBuf> IoBuf for Slice<T> {
-    fn stable_ptr(&self) -> *const u8 {
-        ops::Deref::deref(self).as_ptr()
+impl<T: IoBuf> Slice<T> {
+    /// Returns a raw pointer to the vector’s buffer at the slice offset.
+    ///
+    /// This method is to be used by the `tokio-uring` runtime and it is not
+    /// expected for users to call it directly.
+    pub fn stable_ptr(&self) -> *const u8 {
```
mzabaluev, line 194: I'm not sure it's a good idea to combine inherent methods with the `Deref` impl, as per the API guidelines [C-DEREF], [C-SMART-PTR].

[C-SMART-PTR]: https://rust-lang.github.io/api-guidelines/predictability.html#c-smart-ptr
[C-DEREF]: https://rust-lang.github.io/api-guidelines/predictability.html#c-deref

As these methods are not meant to be called by the API user, perhaps they should be turned into associated functions and/or made private.

On the other hand, the `Deref`/`DerefMut` impls for `Slice` could be considered for removal in favor of explicit accessor methods like `.as_bytes()`/`.as_mut_bytes()`. Especially if we'd want to equip `Slice` with more non-slicey methods for appending to buffers, like `.put_bytes`, `.put_u32_le`, etc.

mzabaluev: > explicit accessor methods like `.as_bytes()` / `.as_mut_bytes()`

... or a `std::ops::Index` impl, as this is what the users would need the most from a buffer, and `[..]` is a concise adapter syntax for anything else where buffer contents would be used as a byte slice.

Noah-Kennedy:
@mzabaluev any interest in completing this?

mzabaluev:
@Noah-Kennedy Yes, I should update the PR at least, will get to it soon.

mzabaluev:
@Noah-Kennedy I have rebased the branch against master and updated the newly added socket ops `SendTo` and `RecvFrom` along with their user-facing methods to the `IntoSlice` API.

FrankReh:
@mzabaluev I know you were asked this before. Any interest in merging with master and pushing up changes again?

This looks like it was going to be important and  @Noah-Kennedy was happy with the direction but was left requiring another review. I would like to help if I can.

FrankReh:
@Noah-Kennedy Should this go in? I can work on rebasing.

mzabaluev:
I will try to update this over the weekend.

FrankReh:
@mzabaluev will you have a look at these CI errors? I suspect they come from documentation examples that were added since you started this PR. If you don't have time, I can take a look today.

FrankReh:
@mzabaluev I have fixed up the CI issues locally. Let me know if its okay with you that I push them up to this branch.

ollie-etl:
All, I'll add I'm also looking at this, although my aims are a little broader. I think `struct Slice` + `Pin`can completely replace the existing `IoBuf` and `IoMut`traits, giving both the very nice simplification shown in this PR, but also generalizing existing methods over any owned buffer which derefs to bytes. I will push as a PR for comment

mzabaluev:
> @mzabaluev I have fixed up the CI issues locally. Let me know if its okay with you that I push them up to this branch.

@FrankReh Go ahead. I suppose the newly added methods like `write_all` need to be converted to the new API.

FrankReh:
@mzabaluev Would you mind reviewing the changes I made to the write_all example and write_all method for tcp, and perhaps the doc examples that use write_all too? Please check that the signatures look right, to be conforming with the others you had changed, that my logic for making progress in the loop inside write_all looks right to you and perhaps also that the progress method seems idiomatic, and finally, my use of in_begin and  in_end to track what the original slice looked like coming in so it could be returned.

FrankReh:
We expect the test to fail today for reasons not related to this commit.

mzabaluev:
@FrankReh The `write_all` changes look fine, thank you.

FrankReh:
@mzabaluev Then this is ready. When master is ready for such a change, @Noah-Kennedy will likely merge. Thank you. It's surprising the old Slice of Slice ... lived here for the last year.

src/buf/slice.rs
mzabaluev: I have realized that the `IoBuf`/`IoBufMut`-cum-inherent-`Slice` methods do not need to be public, so I've made them `pub(crate)`. But the above comments still apply to the other inherent methods of `Slice`.

mzabaluev:
@FrankReh
> I thought they should be pub(crate) but then I found some unit tests failing and decided to leave well-enough alone.

You are right, I've now fixed that.

FrankReh:
Okay. Will wait on @Noah-Kennedy as he had something he wanted to get in first. This still looks good to me.

src/buf/slice.rs
```diff
@@ -29,8 +29,52 @@ pub struct Slice<T> {
...
+    /// Converts the buffer into a [`Slice`] covering its full extent.
+    ///
+    /// This method is used internally by the `io-uring` operations and the
+    /// end user does not need to call it directly.
+    /// The implementation should be equivalent to `self.slice(..)`, but
+    /// can be optimized by skipping range checks.
+    fn into_full_slice(self) -> Slice<Self::Buf> {
+        self.slice(..)
+    }
+}
```
ollie-etl, line 74: Could we slice become a member function of  `impl Slice`, and this trait be replaced with an instance of `From<T: IoBuf> for Slice`. I prefer that because in the absence of specialization, we can then repeatedly chain calls to slice, and they would then operate without creating nested slice instances.

```rust
   let buf = vec![0u8;4096];
   // Creates a slice over the full range
   let mut my_slice: Slice<_> = buf.into();
   // Slice over 32 bytes offset at 16, and also demonstrate that post slice, the indeces are into the visible region of the slice
   my_slice.slice(16..48).slice(0..);
```

ollie-etl: And also because slice is implments IoBuf, inputs to API calls don't need to be slice, but returned values on read-like function will be.

ollie-etl: I did start implmenting this, but got a bit carried away replacing IoBuf with Pin. I think that is a more involved change

mzabaluev: It's an interesting idea; the API methods always returning `Slice<T>` is a bit of a bump when you actually need `T`.

Would `IoBuf` lose its `slice` method then? If not, `<Slice<T> as IoBuf>::slice` remains a wart that could still get called inadvertently in some generic code.

I've started exploring this and came up with #163.

mzabaluev: > `<Slice<T> as IoBuf>::slice` remains a wart

I think this can be fixed by adding an associated type to `IoBuf`. So I'm going to make yet another PR to explore that.

mzabaluev: > I think this can be fixed by adding an associated type to `IoBuf`.

I could not make that work. However, I favour #163 over this PR now.

mzabaluev:
Superseded by #163.

PR #54, opened 2021-09-09
Fixed buffers to support ReadFixed and WriteFixed ops

mzabaluev:
Add infrastructure to manage pre-registered buffers, and operation methods `File::read_fixed_at` and `File::write_fixed_at` to make use of them. Exclusive access to buffer data between the application and in-flight ops is controlled at runtime.

This is initial API to enable fixed buffers. Future developments may include:

- :white_check_mark: Improved parameter polymorphism with ~#53~ #172;
- An internal linked list of free buffers, to be able to check out the next available buffer from `FixedBufRegistry` without enumerating or keeping track of indices;
- Support `IORING_REGISTER_BUFFERS2`/`IORING_REGISTER_BUFFERS_UPDATE`.

Needs #52 to get tests to pass.


src/fs/file.rs
```diff
@@ -170,6 +171,56 @@ impl File {
...
+    ///
+    /// # Examples
+    ///
+    /// ```no_run
+    /// use tokio_uring::fs::File;
+    /// use tokio_uring::buf::fixed::BufRegistry;
+    /// use tokio_uring::buf::IoBuf;
+    /// use std::iter;
+    ///
+    /// fn main() -> Result<(), Box<dyn std::error::Error>> {
```
zeenix, line 194: Might be good to hide the details not specific to this example so user can just read and focus on the relevant details:

```suggestion
    ///# fn main() -> Result<(), Box<dyn std::error::Error>> {
```

src/fs/file.rs
```diff
@@ -170,6 +171,56 @@ impl File {
...
+    ///         let (res, buffer) = f.read_fixed_at(buffer.slice(..), 0).await;
+    ///         let n = res?;
+    ///
+    ///         println!("The bytes: {:?}", &buffer[..n]);
+    ///
+    ///         // Close the file
+    ///         f.close().await?;
+    ///         Ok(())
+    ///     })
+    /// }
```
zeenix, line 212: Might be good to hide the details not specific to this example so user can just read and focus on the relevant details:

```suggestion
    ///# fn main() -> Result<(), Box<dyn std::error::Error>> {
    /// tokio_uring::start(async {
    ///     let registry = BufRegistry::new(iter::repeat(vec![0; 10]).take(10));
    ///     registry.register()?;
    ///
    ///     let f = File::open("foo.txt").await?;
    ///     let buffer = registry.check_out(2).unwrap();
    ///
    ///     // Read up to 10 bytes
    ///     let (res, buffer) = f.read_fixed_at(buffer.slice(..), 0).await;
    ///     let n = res?;
    ///
    ///     println!("The bytes: {:?}", &buffer[..n]);
    ///
    ///     // Close the file
    ///     f.close().await?;
    ///     Ok(())
    /// })
    ///# }
```

mzabaluev: Good idea! I copy-pasted this from the other examples, so perhaps they should be changed likewise.

mzabaluev:
@Noah-Kennedy Yes, I should update the PR at least, will get to it soon.

mzabaluev:
I have rebased the branch. The tests (after merging with #52) sporadically fail, but I see the same flakeyness in master.

src/driver/register.rs
```diff
@@ -0,0 +1,130 @@
...
+                iov_base: buf.as_mut_ptr() as *mut _,
+                iov_len: buf.capacity(),
+            });
+            buf_states.push(BufState::Free {
+                init_len: buf.len(),
+            });
+            mem::forget(buf);
+        }
+        debug_assert_eq!(iovecs.len(), buf_states.len());
+        let raw_vecs = unsafe { ptr::NonNull::new_unchecked(iovecs.as_mut_ptr()) };
```
ollie-etl, line 40: Is the assumption that iovecs is contiguous not erroneous? I don't believe this is enforced by `impl Iterator`

ollie-etl: ^ Please disregard this

FrankReh:
@mzabaluev Can we try again and get this merged? You put a lot of work into this. I can get up to speed on it shortly if you wanted to rebase with master. Perhaps wait for the other merge we might pick up first so the rebase only has to be done once more.

mzabaluev:
@FrankReh I'll try to rebase the branch this week.

mzabaluev:
@ollie-etl
> I believe the api for Buffers::new is flawed, and should instead take a buffer and fragment it internally.

This API allows pre-registering an arbitrary set of pre-allocated buffers. While it's true that each buffer needs to be allocated separately, registering is meant to be a one-time operation for the lifetime of `Runtime`, and with a good allocator and reasonably sized buffers, the overhead should be negligible. If `bytes` was a non-optional dependency, it would be possible to take a sequence of `BytesMut` handles into fragments of a larger buffer.

FrankReh (changes requested):
The API provided by this PR needs to be evaluated in light of the pbuf_ring support that is going to this crate in the next week or so.

FrankReh:
Okay, here's the thing. In my defense, I assumed the PR by the same OP was going to be worked on first and I would have time to look through this and fully understand what's going on. I think the feature this PR addresses should be handled with a trait, perhaps named FixedBuffers, so the read_fixed and write_fixed operations are not tied to the opinions of buffer management.

The PBUF_RING feature I am introducing in the next few days uses a trait, BufferGroup, for the same reasons.

I would like to start a fresh review of this PR in the morning. If @mzabaluev, who started this PR over a year ago, would like me to make modifications based on the review, that would be fine and very understandable.

Or if @Noah-Kennedy or @carllerche would like to see this go in, as is, and then wait for my own PR that proposes making it generic, that would be fine too.

The uring interface has evolved a lot since its introduction, I guess with kernel 5.1, but mostly in the area of reading or receiving into buffers. The fixed buffer approach may be the only shared buffer mechanism introduced for writing so far. That's interesting and not something I appreciated until I started looking at this.

src/driver/mod.rs
```diff
@@ -37,19 +42,26 @@ mod util;
...
 
 pub(crate) struct Driver {
     /// In-flight operations
     ops: Ops,
 
     /// IoUring bindings
     pub(crate) uring: IoUring,
+
+    /// Reference to the currently registered buffers
+    buffers: Option<Rc<RefCell<Buffers>>>,
```
FrankReh, line 64: This is one of the big deals. Adding to the core Driver.

It's not needed for using the registry, as there are no general functions yet for checking out a buffer that don't require the registry itself.

But it probably provides an important safety guaranteeing the registery won't be dropped without the entire driver being dropped.

So would you mind naming it fixed_buffers, or fixed_registry, or something similar? The term 'buffers' is going to be all over the place in this crate.

src/driver/read_fixed.rs
```diff
@@ -0,0 +1,64 @@
...
+use std::io;
+
+pub(crate) struct ReadFixed {
+    /// Holds a strong ref to the FD, preventing the file from being closed
+    /// while the operation is in-flight.
+    #[allow(dead_code)]
+    fd: SharedFd,
+
+    /// The in-flight buffer.
+    buf: Slice<FixedBuf>,
```
FrankReh, line 16: This is very cool that the Slice of a FixedBuf can be used like this because the uring interface expressly allows the fixed read (and fixed write) to be done from addresses not necessarily at the start of the registered fixed buffer.

src/driver/read_fixed.rs
```diff
@@ -0,0 +1,64 @@
...
+        // Convert the operation result to `usize`
+        let res = cqe.result.map(|v| v as usize);
+        // Recover the buffer
+        let mut buf = self.buf;
+
+        // If the operation was successful, advance the initialized cursor.
+        if let Ok(n) = res {
+            // Safety: the kernel wrote `n` bytes to the buffer.
+            unsafe {
+                buf.set_init(n);
```
FrankReh, line 58: Very cool that setting init on a slice works.

src/fs/file.rs
```diff
@@ -290,6 +291,56 @@ impl File {
         op.await
     }
 
+    /// Like [`read_at`], but using a pre-mapped buffer
+    /// registered with [`BufRegistry`].
```
FrankReh, line 295: Should this comment point out a slice of the registered buffer is also allowed? The signature does make it obvious but I'm still confused myself whether the slicing has to be made from front to back or could an app start from the back, or even the middle if they wanted? Well, maybe not a question for this review.

tests/fixed_buf.rs
```diff
@@ -0,0 +1,95 @@
...
+        // Meanwhile, we replace buffer registration in the kernel:
+        buffers.unregister().unwrap();
+        let buffers = BufRegistry::new([Vec::with_capacity(1024)]);
+        buffers.register().unwrap();
+
+        // The old buffer's index no longer matches the memory area of the
+        // currently registered buffer, so the read operation using the old
+        // buffer's memory should fail.
+        let (res, _) = file.read_fixed_at(fixed_buf.slice(..), 0).await;
+        assert_err!(res);
```
FrankReh, line 83: A very nice test and nice explanation.

FrankReh:
@ollie-etl 
> I believe the api for Buffers::new is flawed, and should instead take a buffer and fragment it internally.

Doesn't the caller have the option of allocating a buffer and passing in an iterator over that buffer if they want? The crate could even provide a helper function or two to do that using that, perhaps based on std::alloc::alloc_zeroed(layout).

src/driver/buffers.rs
```diff
@@ -0,0 +1,132 @@
...
+    orig_cap: usize,
+}
+
+enum BufState {
+    Free { init_len: usize },
+    CheckedOut,
+}
+
+impl Buffers {
+    pub(crate) fn new(bufs: impl Iterator<Item = Vec<u8>>) -> Self {
```
FrankReh, line 24: Is this Item of type Vec<u8> the only and best choice? Is there a more primitive type that would serve the same purpose? Maybe its a good choice because the std library comes with so many ways to unsafely create a Vec from a larger allocation?

FrankReh (changes requested):
This is really awesome. It is much less opinionated than I had assumed it had to be.

Taking ownership of the buffer iterator by new so it can all be dropped when the last reference to the registry is dropped is very elegant.

Lots of review comments, most just comments.

src/buf/fixed.rs
```diff
@@ -0,0 +1,184 @@
...
+/// the kernel for `io-uring` operations.
+///
+/// `FixedBuf` handles can be obtained from a [`BufRegistry`] collection.
+/// For each buffer, only a single `FixedBuf` handle can be either used by the
+/// application code or owned by an I/O operation at any given time,
+/// thus avoiding data races between `io-uring` operations in flight and
+/// the application accessing buffer data.
+pub struct FixedBuf {
+    registry: Rc<RefCell<Buffers>>,
+    buf: ManuallyDrop<Vec<u8>>,
```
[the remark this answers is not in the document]
oliverbunting: `Box<[T]>`?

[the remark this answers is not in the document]
FrankReh: @oliverbunting I don't think it's  possible to create a Box around a variable sized slice because trait Sized is not implemented for [u8]. Did you mean something else?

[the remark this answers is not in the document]
FrankReh: A Vec also has the nice properties of len and cap, not just len, and I think this PR was taking advantage of an initial length, perhaps to avoid any initialization of the data being given to the kernel, or maybe to pre-seed the buffers with some values - perhaps to be used for a fixed buffer write. It's just the growable that I don't like but lots of crates seem to have decided to just not let their types grow by not calling things like push.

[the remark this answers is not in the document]
ollie-etl: @FrankReh I've long lamented that (as far as I know) there is no way of specifying that you _only_ want a generic `T` to be by value, and not by reference. `AsRef<[T]>` and `Deref` are just too permissive here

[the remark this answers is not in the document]
FrankReh: @ollie-etl Your point is that the API this feature introduces doesn't really give ownership of the memory to the driver?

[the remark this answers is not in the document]
ollie-etl: No, it does. My point is I'd very much like to be able to generalize over any type T which can be `Deref`'d to a `[u8]`, but unfortunately I can't exclude `&[u8]` from that set, which prevents the generalization.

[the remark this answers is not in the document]
ollie-etl: Hmm, I have a suggestion - I'll do a full review and see if it sticks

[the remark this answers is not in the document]
ollie-etl: > Hmm, I have a suggestion - I'll do a full review and see if it sticks

I've looked into this, its much broader than the scope of this PR. This PR should not be blocked on my musings, it is useful as is, and can always be changed to a more general form in the future.

src/buf/fixed.rs
```diff
@@ -0,0 +1,184 @@
...
+/// A `BufRegistry` value is a lightweight handle for a collection of
+/// allocated buffers. Cloning of a `BufRegistry` creates a new reference to
+/// the same collection of buffers.
+///
+/// The buffers of the collection are not deallocated until:
+/// - all `BufRegistry` references to the collection have been dropped;
+/// - all [`FixedBuf`] handles to individual buffers in the collection have
+///   been dropped, including the buffer handles owned by any I/O operations
+///   in flight;
+/// - every `tokio-uring` runtime the buffers have been registered in
```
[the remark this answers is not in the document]
mzabaluev: > Presumably the uring interface will report an error if a buffer address is used in a fixed read or write that doesn't fall into the range mapped by the buffer index.

There is a test that confirms this behavior, albeit in the specific case of a buffer having been unregistered while its `FixedBuf` handle is still alive.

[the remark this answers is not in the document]
mzabaluev: > Do we think the same buffer collection can be shared between rings?

It's technically possible, but with `Rc` you'd need to have more than one `Runtime` in the same thread and I don't know if such a setup is usable.

src/buf/fixed.rs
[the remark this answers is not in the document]
mzabaluev: `Vec` is used because we need to pass the original capacity on deallocation. Once the buffers are owned by the registry instance, they are not grown or shrunk.

The discussion made me think of adding an allocator type parameter so we can also accommodate `Vec` with custom allocators.

src/driver/buffers.rs
```diff
@@ -0,0 +1,132 @@
...
+use std::slice;
+
+pub(crate) struct Buffers {
+    raw_vecs: ptr::NonNull<iovec>,
+    buf_states: Vec<BufState>,
+    orig_cap: usize,
+}
+
+enum BufState {
+    Free { init_len: usize },
```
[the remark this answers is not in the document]
mzabaluev: It's needed to preserve the length of the initial**ized** part of a buffer. The information on the raw buffer goes into the `iovec`, but to recover access to the initialized bytes through the `FixedBuf`, we also need the length of the initialized part for the `Vec::from_raw_parts` call. Operations writing into the buffer can also extend the initialized part through `IoBufMut::set_init`.

I can add some documentation on management of initialized data in fixed buffers.

tests/fixed_buf.rs
mzabaluev: I'm thinking of putting these explanations into assert failure messages, but this is just stylistics.

src/driver/buffers.rs
```diff
@@ -0,0 +1,132 @@
...
+    CheckedOut,
```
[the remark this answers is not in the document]
mzabaluev: When the buffer is checked out, the length of the initialized part is recorded in the `FixedBuf`. It can also be changed by operations, so it's best to have it accounted in a single place.

src/driver/buffers.rs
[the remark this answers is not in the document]
FrankReh: Thank you. Some documentation on the management of the initialized data in fixed buffers would go a long way in letting someone know what they are dealing with when they are considering creating a fixed buffer pool and then using it in fixed read and write operations.

You don't have to do it, but some unit tests, or example code in the document some day is called for too. We can leave it out for now.

src/driver/buffers.rs
mzabaluev: I used `Vec<u8>` as a safe representation of an allocated buffer. I'm not aware of another primitive type to provide the semantics of memory allocation and management of initialized vs. uninitialized data. It could be made more convenient if the initializing iterator items were generic and bound with `Into<Vec<u8>>`, but I do not want to hide the act of providing a number of allocated buffers.

As mentioned above, the registry API can be easily made generic around the allocator the same way `Vec` is.

Otherwise, I have not yet thought of making the fixed buffers API more generic. I think the primary use case here is to allocate one or several flat byte buffers, register them with the kernel, and then use the buffers repeatedly with I/O operations, filling or consuming buffer data as needed by the application with its program-friendly memory structures living elsewhere.

src/driver/buffers.rs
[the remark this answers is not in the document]
FrankReh: Makes sense. Would you mind adding a one line comment to that effect for this invariant?

FrankReh (changes requested):
Sorry, I seem to have appointed myself the king of asking for one line  comments here and there. I can see myself being dethroned very easily if anyone wants to say the code speaks for itself better than a comment can.

src/buf/fixed.rs
```diff
@@ -0,0 +1,184 @@
...
+///   has terminated.
+#[derive(Clone)]
+pub struct BufRegistry {
```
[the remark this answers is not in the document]
mzabaluev: The path in the public API is already `buf::fixed::BufRegistry`, and I'm of the school of thought that [instructs](https://github.com/rust-lang/rfcs/blob/master/text/0356-no-module-prefixes.md) to use the module names in preference to prefixes. Then again, I've got `buf::fixed::FixedBuf`, but that has not been named `Buf` to avoid clashes with `bytes::Buf`.

Perhaps we could justify the prefix as describing the distinctive attribute of these objects, so `FixedBufRegistry` it should be.

FrankReh:
Looks like a name in a documented example didn't get changed.

src/buf/fixed/registry.rs
```diff
@@ -0,0 +1,114 @@
...
+    }
+
+    /// Returns a buffer identified by the specified index for use by the
+    /// application, unless the buffer is already in use.
+    ///
+    /// The buffer is released to be available again once the
+    /// returned `FixedBuf` handle has been dropped. An I/O operation
+    /// using the buffer takes ownership of it and returns it once completed,
+    /// preventing shared use of the buffer while the operation is in flight.
+    pub fn check_out(&self, index: usize) -> Option<FixedBuf> {
```
mzabaluev, line 105: Perhaps this should be renamed to just `get`: the API user is not much interested in the internal mechanics of checking buffers in and out.

src/fs/file.rs
```diff
@@ -290,6 +291,56 @@ impl File {
...
+    ///     let (res, buffer) = f.read_fixed_at(buffer.slice(..), 0).await;
+    ///     let n = res?;
+    ///
+    ///     println!("The bytes: {:?}", &buffer[..n]);
+    ///
+    ///     // Close the file
+    ///     f.close().await?;
+    ///     Ok(())
+    /// })
+    ///# }
```
FrankReh, line 332: Nice example. I think we should get something in, with the understanding that some things will change as we try to support Read and Write traits.

src/fs/file.rs
```diff
@@ -290,6 +291,56 @@ impl File {
...
+    /// ```
+    pub async fn read_fixed_at(
+        &self,
+        buf: Slice<FixedBuf>,
+        pos: u64,
+    ) -> crate::BufResult<usize, Slice<FixedBuf>> {
```
FrankReh, line 338: This signature has changed right? I thought it had been taking the id. This is probably better.

src/driver/read_fixed.rs
```diff
@@ -0,0 +1,64 @@
+use crate::buf::fixed::FixedBuf;
+use crate::buf::{IoBuf, IoBufMut, Slice};
+use crate::driver::op::{self, Completable};
+use crate::driver::{Op, SharedFd};
+use crate::BufResult;
+
+use std::io;
+
+pub(crate) struct ReadFixed {
```
FrankReh, line 9: Now I see a little more, I'm confused a little more. 

The fixed read and write operations should work on sockets too, although I don't find liburing tests that expressly prove that. The man pages for fixed read and write say they are like the read and write operations, and those work for socket fds.

This morning, I'm confused by the existing driver in this crate having a 'struct Read<T>' in driver/read.rs and that struct has a buf: T field, but there is also a struct Socket in driver/socket.rs that does not have a buf field.

So at a minimum, I'm wondering if these read_fixed_at and write_fixed_at should be offered in the socket.rs Socket impl too. Or at least a comment saying once this is in for files, we'll want to do the same for sockets at a later time.

src/driver/register_buffers.rs
```diff
@@ -0,0 +1,37 @@
+use crate::buf::fixed::FixedBuffers;
+use crate::runtime::CONTEXT;
+
+use std::cell::RefCell;
+use std::io;
+use std::rc::Rc;
+
+pub(crate) fn register_buffers(buffers: &Rc<RefCell<FixedBuffers>>) -> io::Result<()> {
```
FrankReh, line 8: Nice file.

src/buf/fixed/registry.rs
FrankReh: I could see either way. I like specific names like check_out, but over the long haul, packages like tokio seem to go with short meaning names, even if those names are used in other modules for other purposes too. I'll let you and @Noah-Kennedy figure this out.

FrankReh (changes requested):
Very nice structure and names. I don't think I found anything that seems to be a problem. I'll admit I didn't try to follow the unit tests this time because I'm assuming they were like before, just in different files now.

I propose this merge wait both for @Noah-Kennedy 's review and for the Slice PR to be merged first.

Also I would vote that we get this in soon, when we can, even though we know slices and signatures and traits and builder op APIs are coming down the road. While we have active participants, the more we can get into the master branch that compiles and passes unit tests, the more we have to work with when trying to pull it all together into a tight API.

src/driver/read_fixed.rs
mzabaluev: I've been wondering about how these operatoins apply to sockets too.
I guess for sockets the reads/writes can't have a meaningful offset, so the methods will be named `read_fixed` and `write_fixed`.

src/fs/file.rs
mzabaluev: Indeed: we have to pass valid buffer data to the op (the pointer, the length and the id), but the caller might want to limit the read to an offset portion of the buffer (e.g. to implement something like `read_exact`). With #163, the argument could be also generalized to `impl Into<Slice<FixedBuf>>`.

src/buf/fixed/registry.rs
mzabaluev: I recall now why I decided against `get`: this name is commonly given to idempotent operations, but here you can't "get" the buffer under the same index a second time unless you drop it first.

FrankReh: Okay. Good enough reason for me. Also, I have to remind myself too that none of this is cast in store. Getting it in and letting people use it is the most important thing. Everyone should know APIs can change this year and probably into next year.

src/fs/file.rs
FrankReh: Yes. If we end up having a choice, which would you like to see be merged first?

src/driver/read_fixed.rs
FrankReh: Yes. Very reasonable. Can be a later PR if you like.

FrankReh (commented):
Nice. I'm going to give these a spin.

FrankReh:
@mzabaluev Running into this compile error. Does this not have your latest slice branch merged in? Does that explain the Slice<Slice<FxedBuf>> error?

```rust
$ cat tcp_listener_fixed_buffers.rs
use std::{env, iter, net::SocketAddr};

use tokio_uring::buf::fixed::FixedBufRegistry;
use tokio_uring::net::TcpListener;

fn main() {
    let args: Vec<_> = env::args().collect();

    let socket_addr = if args.len() <= 1 {
        "127.0.0.1:0"
    } else {
        args[1].as_ref()
    };
    let socket_addr: SocketAddr = socket_addr.parse().unwrap();

    tokio_uring::start(async {
        let listener = TcpListener::bind(socket_addr).unwrap();

        println!("Listening on {}", listener.local_addr().unwrap());

        loop {
            let (stream, socket_addr) = listener.accept().await.unwrap();
                // implement ping-pong loop

                println!("{} connected", socket_addr);
                let mut n = 0;

                let registry = FixedBufRegistry::new(iter::repeat(vec![0; 4096]).take(1));
                registry.register().unwrap();

                let buffer = registry.check_out(0).unwrap();

                use tokio_uring::buf::IoBuf;
                loop {
                    let (result, nslice) = stream.read_fixed(buffer.slice(..)).await;
                    let read = result.unwrap();
                    if read == 0 {
                        println!("{} closed, {} total ping-ponged", socket_addr, n);
                        break;
                    }

                    // Ignore fact that this isn't a write_all with fixed yet. My first tests are
                    // with short texts from nc anyway.
                    let (res, nslice) = stream.write_fixed(nslice.slice(..read)).await;
                    let _ = res.unwrap();
                    buffer = nslice.into_inner();
                    println!("{} all {} bytes ping-ponged", socket_addr, read);
                    n += read;
                }
            });
        }
    });
}
```

```sh
$ cargo build --release --example tcp_listener_fixed_buffers
  Compiling tokio-uring v0.4.0 (/home/frank/forks/tokio-uring)
error[E0308]: mismatched types
   --> examples/tcp_listener_fixed_buffers.rs:45:60
    |
45  |                     let (res, nslice) = stream.write_fixed(nslice.slice(..read)).await;
    |                                                ----------- ^^^^^^^^^^^^^^^^^^^^ expected struct `FixedBuf`, found struct `Slice`
    |                                                |
    |                                                arguments to this function are incorrect
    |
    = note: expected struct `Slice<FixedBuf>`
               found struct `Slice<Slice<FixedBuf>>`
note: associated function defined here
   --> /home/frank/forks/tokio-uring/src/net/tcp/stream.rs:114:18
    |
114 |     pub async fn write_fixed(
    |                  ^^^^^^^^^^^

For more information about this error, try `rustc --explain E0308`.
error: could not compile `tokio-uring` due to previous error
```

FrankReh:
Here's a version that works, albeit for just one connection at a time. It shows the buffer is being checked in when it gets dropped.

```rust
$ cat tcp_listener_fixed_buffers.rs
use std::{cell::RefCell, env, iter, net::SocketAddr, rc::Rc};

use tokio_uring::buf::fixed::FixedBufRegistry;
use tokio_uring::net::TcpListener;

fn main() {
    let args: Vec<_> = env::args().collect();

    let socket_addr = if args.len() <= 1 {
        "127.0.0.1:0"
    } else {
        args[1].as_ref()
    };
    let socket_addr: SocketAddr = socket_addr.parse().unwrap();

    tokio_uring::start(async {
        let listener = TcpListener::bind(socket_addr).unwrap();

        println!("Listening on {}", listener.local_addr().unwrap());

        let registry = FixedBufRegistry::new(iter::repeat(vec![0; 4096]).take(1));
        registry.register().unwrap();
        let registry = Rc::new(RefCell::new(registry));

        loop {
            let (stream, socket_addr) = listener.accept().await.unwrap();
            let registry = registry.clone();
                // implement ping-pong loop

                println!("{} connected", socket_addr);
                let mut n = 0;

                // Check out anew, letting it get dropped.
                // Sure, it means this server can only handle one connection at a time.
                let mut buffer = registry.borrow().check_out(0).unwrap();

                use tokio_uring::buf::IoBuf;
                loop {
                    let (result, nslice) = stream.read_fixed(buffer.slice(..)).await;
                    let read = result.unwrap();
                    if read == 0 {
                        println!("{} closed, {} total ping-ponged", socket_addr, n);
                        break;
                    }
                    buffer = nslice.into_inner();

                    // Ignore fact that this isn't a write_all with fixed yet. My first tests are
                    // with short texts from nc anyway.
                    let (res, nslice) = stream.write_fixed(buffer.slice(..read)).await;
                    let _ = res.unwrap();
                    buffer = nslice.into_inner();
                    println!("{} all {} bytes ping-ponged", socket_addr, read);
                    n += read;
                }
            });
        }
    });
}
```

mzabaluev:
> Does this not have your latest slice branch merged in? Does that explain the `Slice<Slice>` error?

Yes, I'm developing #53 independently.

There is no need to do this, `FixedBufRegistry` does it already internally and the registry handle can be cloned:

```rust
       let registry = Rc::new(RefCell::new(registry));
```

To resolve the "one buffer at a time" problem, I plan to add a linked list through free buffers later.

FrankReh:
@mzabaluev Right. I've got the example working without the Rc and RefCell now. Thanks for the pointer.

Having a check_out_next would be convenient if it can use the same space. Sounds like a slab in a way, but one that doesn't grow. Separate PR though right? I don't want this work to get even bigger while waiting for it to be merged.

tests/fixed_buf.rs
```diff
@@ -0,0 +1,95 @@
...
+
+        let fixed_buf = buffers.check_out(0).unwrap();
+        let (res, buf) = file.read_fixed_at(fixed_buf.slice(..), 0).await;
+        let n = res.unwrap();
+        assert_eq!(n, HELLO.len());
+        assert_eq!(&buf[..], HELLO);
+    });
+}
```
FrankReh, line 91: Could we have one unit test that checks a slice with non-zero offset to a FixedBuf works as expected for a read_fixed and write_fixed? It's really cool we can use a fixed buffer and not always start the reads or writes from the beginning.

mzabaluev:
I'd prefer #172 merged first so that the API of the new methods could be made less fugly.

mzabaluev:
I want to provide access to the initialized bytes in `FixedBuf` as an `&[u8]`.
What's your opinion on implementing `Deref`/`DerefMut`? `Slice` implements these traits, but it's potentially troublesome to have `Deref` together with inherent methods or traits with application-callable methods that do something other than trivially delegate to the target ([1](https://rust-lang.github.io/api-guidelines/predictability.html#c-deref) [2](https://rust-lang.github.io/api-guidelines/predictability.html#c-smart-ptr)). Perhaps it's more prudent to implement `Index`, `IndexMut`, and `AsRef<[u8]>`?

mzabaluev:
Thanks @FrankReh and @Noah-Kennedy for getting #172 merged. The API here is much nicer now, both `FixedBuf` and `Slice<FixedBuf>` can be passed to the operations.

FrankReh:
I'll try to review now. Took a quick glance and they did look cleaner. But @Noah-Kennedy will be anxious to get his moved files in so this may have to go through another rebase. From his move diffs though, mostly files were moved with no symbol name changes, but the odd *use* or *mod* change. I don't know if git will be making that easy for you to merge or rebase with or not.  But before doing too much manual work, probably want to wait for Noah to also bless this type of work. I don't think he's commented on it recently.

src/driver/read_fixed.rs
```diff
@@ -1,27 +1,30 @@
...
-use crate::buf::{IoBuf, IoBufMut, Slice};
+use crate::buf::BoundedBufMut;
 use crate::driver::op::{self, Completable};
 use crate::driver::{Op, SharedFd};
 use crate::BufResult;
 
 use std::io;
 
-pub(crate) struct ReadFixed {
+pub(crate) struct ReadFixed<T> {
```
FrankReh, line 9: Now it's generic over T, but before it doesn't have to be. It took a Slice<FixedBuf> buf. Wonder what the implications of that are.

src/driver/read_fixed.rs
```diff
@@ -1,27 +1,30 @@
...
 
     /// The in-flight buffer.
-    buf: Slice<FixedBuf>,
+    buf: T,
 }
 
-impl Op<ReadFixed> {
+impl<T> Op<ReadFixed<T>>
+where
+    T: BoundedBufMut<BufMut = FixedBuf>,
```
FrankReh, line 21: Interesting, the associated type is FixedBuf. And not a FixedBufMut?

FrankReh (approved):
I think it looks very good. As with all things so complex, there will be many color choices to paint the shed. I would love to see this go in, even if there are things to change down the road - although right now I'm not thinking of any. I can't tell you how many times I thought about doing something with indexed buffers only to then remember that they aren't in the library yet.

FrankReh:
Going to wait for @Noah-Kennedy to decide the timing of this merge, or to provide his own review too of course.

Noah-Kennedy:
I just merged #180, so this needs a rebase. I'll review once that happens.

tests/fixed_buf.rs
mzabaluev: Done in https://github.com/tokio-rs/tokio-uring/pull/54/commits/f0443d7a1c9cd6ce5ecdc67a69414579e6a484c9

src/driver/read_fixed.rs
mzabaluev: Now it will carry a `FixedBuf` if that's what the caller has passed, no need for a `Slice` wrapper in this case.

tests/fixed_buf.rs
```diff
@@ -90,6 +90,47 @@ fn unregister_invalidates_checked_out_buffers() {
     });
 }
 
+#[test]
```
FrankReh, line 93: This is very cool. Would you mind adding a few one line comments saying what the asserts are proving? I know what they should be proving, since it is called slicing, but a few comments can go a long way in getting the light bulb to go off for someone. Again. Slicing fixed buffers is way cool. I'm glad the uring driver folks thought of it and I'm glad this is going to support it.

FrankReh (approved):
Thanks. One comment.

src/driver/read_fixed.rs
mzabaluev: There is no `FixedBufMut`, a `FixedBuf` handle can be used as a mutable buffer.

src/driver/read_fixed.rs
```diff
@@ -42,8 +45,11 @@ impl Op<ReadFixed> {
     }
 }
 
-impl Completable for ReadFixed {
-    type Output = BufResult<usize, Slice<FixedBuf>>;
+impl<T> Completable for ReadFixed<T>
+where
+    T: BoundedBufMut<BufMut = FixedBuf>,
```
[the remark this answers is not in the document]
mzabaluev: Yes, exactly.

src/buf/fixed/handle.rs
```diff
@@ -0,0 +1,110 @@
...
+        &mut self.buf
+    }
+}
+
+impl Debug for FixedBuf {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        f.debug_struct("FixedBuf")
+            .field("buf", &*self.buf) // deref ManuallyDrop
+            .field("index", &self.index)
+            .finish_non_exhaustive()
```
mzabaluev, line 108: I'm not sure what the MSRV policy here is, but this requires Rust 1.53.

FrankReh: I think that's okay. We have other things that probably require an even newer compiler version already.

FrankReh (approved):
LGTM

Noah-Kennedy (changes requested):
I'm not sure about parts of this.

I'm a bit concerned about soundness in the case where buffers are checked out and then unregistered. If my understanding is correct, a user could safely cause undefined behavior by using the deref implementation of an existing buffer after the set of buffers has been unregistered.

I'm also uncomfortable with the idea of this being able to leak buffers this easily.

I don't necessarily have solutions to these issues right now, but they need solved before we can merge.

mzabaluev:
> I'm a bit concerned about soundness in the case where buffers are checked out and then unregistered. If my understanding is correct, a user could safely cause undefined behavior by using the deref implementation of an existing buffer after the set of buffers has been unregistered.

Unregistering the buffers does not get them deallocated, as long as there is a `FixedBuf` or a `FixedBufRegistry` handle holding a reference to the registry's inner data. And only these objects provide the safe API to access buffers' data.

My concern is rather that it's not necessarily easy to know the moment when the buffers get dropped. But:

1. The v1 io-uring API for fixed buffers is mostly suited for one time set up and tear down rather than juggling with successive generations of buffers.
2. The `IORING_REGISTER_BUFFERS2` API with tagged buffers in fact promotes the model of eventual dropping of replaced buffers as CQEs reporting their release by the kernel get received.

> I'm also uncomfortable with the idea of this being able to leak buffers this easily.

The only problem I see is that leaking a single `FixedBuf` handle, in general scenarios where such leaking is possible, would leak the entire buffer collection owned by its `FixedBufRegistry`.

FrankReh:
@Noah-Kennedy I think you raised two concerned about 12 hours ago and @mzabaluev gave some feedback 7 hours ago. Do you still believe holding a reference to a buffer after the kernel has had the buffers unregistered is a problem? I may just be repeating @mzabaluev , but I believe the memory has not been freed in that case so no UB there, and the kernel would reject further attempts at using that buffer (but it would have had to be owned, not just referenced). So maybe not a problem?

On the second concern, where do you see buffers being leaked? Or is that part of the first concern and @mzabaluev and I don't follow your point, or you've changed your mind on it?

Noah-Kennedy:
Sorry about the responsiveness, I'm traveling at moment.

@mzabaluev thanks for the clarification, I see that I misread how that is tracked.

I'm in favor of merging.

Noah-Kennedy (approved):

Noah-Kennedy:
@FrankReh I'll leave the merging to you as I'm getting on a plane in just a moment.

Noah-Kennedy:
I'm fully in agreement here.

mzabaluev:
Thanks for shepherding this!

> The fixed indexes begs for one or more implementations of sets of buffers. One where the _next_ is supported, as @mzabaluev has called out before.

I'm pretty sure I can work two `Option<u16>` linked list indices for the next and the previous free buffer into `BufState::Free`, so that the overall size of the enum does not increase. Updating the list when a buffer gets freed takes constant time. So the `try_next` operation could be added for `FixedBufRegistry` itself.

> I'm also wondering about one where _next_ would be a Future so one could _await_ the next free index. Let's say I create an index registry with 1000 buffers and I don't want to drop a service if another buffer isn't available, but rather I want to wait for one to free up.

I think it's also feasible to implement an async `next` method for `FixedBufRegistry`. Tasks pending on `next` can post their wakers internally at the registry, and dropping a `FixedBuf` will fire up all accumulated wakers at its registry. This feature has a small and constant runtime cost per each buffer check out/in, unless the tasks exhaust the pool of free buffers.

> Once it is behind a Future, it could also be behind a timed Future that does something different if a buffer isn't freed in a reasonable amount of time.

I can't think of anything that can't be achieved with existing composing APIs like `tokio::time::timeout`, but maybe there is something clever to do with the buffers internally when the timeout is elapsed?

FrankReh:
Not to confuse anyone, I totally saw using the tokio timeout future. Didn't mean it would be built into our library or an extras module. Just that waiting for the *next* would be useful and then it's easy to break out with a timeout too if the situation calls for it. Guess I was stating the obvious and that was confusing.

PR #73, opened 2022-02-14
Direct I/O

MarinPostma:
This PR adds support for Direct I/O for tokio-uring

# Todo
- [x] Support basic i/o operation on a DMA file
- [x] Handle buffer allocation alignment
- [ ] Retrieve device alignment requirements
- [ ] Better documentation
- [ ] More tests


MarinPostma:
i unfortunately don't have the time to work on that for now, closing this PR, and I'll maybe reopen when I have more time.

ahrens:
@MarinPostma We're interested in trying this out, specifically for doing directio to block devices.  Just wanted to check if you're aware of any issues with this code aside from the ones you mentioned above (determining the file's alignment, docs, tests).
(cc @sumedhbala-delphix @tonynguien)

Noah-Kennedy:
@MarinPostma I'm actually quite interested in getting this in sometime, I just don't have as much bandwidth right now.

If anyone is interested in driving this forward, I would be quite interested in helping and assisting.

MarinPostma:
@ahrens I'm not sure this is working, i haven't touched this or in a while.

@Noah-Kennedy happy continue this work, if any one is ready to help with API design

ahrens:
One thing that would be a nice improvement to the API would be for the read/write calls to take a generic trait equivalent to IoBuf/[IoBufMut](https://docs.rs/tokio-uring/latest/tokio_uring/buf/trait.IoBufMut.html), which DmaBuffer implemented.  That way consumers could implement their own memory management of the buffers.  We're doing that today with IoBufMut, so that we can interface with [Bytes](https://docs.rs/bytes/latest/bytes/struct.Bytes.html)/[BufMut](https://docs.rs/bytes/latest/bytes/buf/trait.BufMut.html) without copying data.

PR #74, opened 2022-02-27
Add unix stream & listener

jon-chuang:
Todo: 
- [X] convert all uses of OsSocket to socket2::Socket (to reduce dependencies)

Future:
- Unix datagram

Comments:
- Looking at https://github.com/tokio-rs/tokio-uring/pull/40, apparently I reverted on suggestion not to use `ToSocketAddress`. Personally, I like the use better, and also it is more consistent with tokio API. So I'm proposing to keep this change.

### DIFF
--- a/src/driver/socket.rs
+++ b/src/driver/socket.rs
@@ -92,10 +92,21 @@ impl Socket {
     }
 
     pub(crate) fn bind(socket_addr: SocketAddr, socket_type: libc::c_int) -> io::Result<Socket> {
-        let socket = Socket::new(socket_addr, socket_type)?;
-        let addr = SockAddr::from(socket_addr);
-        syscall!(bind(socket.as_raw_fd(), addr.as_ptr(), addr.len()))?;
-        Ok(socket)
+        let sys_listener =
+            socket2::Socket::new(get_domain(socket_addr).into(), socket_type.into(), None)?;
+        let addr = socket2::SockAddr::from(socket_addr);
+
+        sys_listener.set_reuse_port(true)?;
+        sys_listener.set_reuse_address(true)?;
+
+        // sys_listener.set_send_buffer_size(send_buf_size)?;
+        // sys_listener.set_recv_buffer_size(recv_buf_size)?;
+
+        sys_listener.bind(&addr)?;
+
+        let fd = SharedFd::new(sys_listener.as_raw_fd());
+
+        Ok(Self { fd })
     }
 
     pub(crate) fn bind_unix<P: AsRef<Path>>(
--- a/src/driver/socket.rs
+++ b/src/driver/socket.rs
@@ -2,6 +2,7 @@ use crate::{
     buf::{IoBuf, IoBufMut},
     driver::{Op, SharedFd},
 };
+use socket2::SockAddr;
 use std::{
     io,
     net::SocketAddr,
@@ -66,7 +67,7 @@ impl Socket {
         op.recv().await
     }
 
-    pub(crate) async fn accept(&self) -> io::Result<(Socket, SocketAddr)> {
+    pub(crate) async fn accept(&self) -> io::Result<(Socket, Option<SocketAddr>)> {
         let op = Op::accept(&self.fd)?;
         let completion = op.await;
         let fd = completion.result?;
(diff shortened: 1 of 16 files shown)


src/driver/socket.rs
```diff
@@ -92,6 +120,16 @@ impl Socket {
         Ok(socket)
     }
 
+    pub(crate) fn bind_unix<P: AsRef<Path>>(
```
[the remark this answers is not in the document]
jon-chuang: Reduced some of proliferation

jon-chuang:
Ready for review

src/driver/socket.rs
```diff
@@ -31,6 +32,14 @@ impl Socket {
         Ok(Socket { fd })
     }
 
+    pub(crate) fn new_unix(socket_type: libc::c_int) -> io::Result<Socket> {
+        let socket_type = socket_type | libc::SOCK_CLOEXEC;
+        let domain = libc::AF_UNIX;
+        let fd = syscall!(socket(domain, socket_type, 0))?;
```
Noah-Kennedy, line 38: You can probably do this with socket2

src/net/tcp/listener.rs
```diff
@@ -43,10 +46,19 @@ impl TcpListener {
     /// Binding with a port number of 0 will request that the OS assigns a port
     /// to this listener. The port allocated can be queried via the `local_addr`
     /// method.
-    pub fn bind(socket_addr: SocketAddr) -> io::Result<TcpListener> {
-        let socket = Socket::bind(socket_addr, libc::SOCK_STREAM)?;
-        socket.listen(1024)?;
-        Ok(TcpListener { inner: socket })
+    ///
+    /// [`ToSocketAddrs`]: trait@std::net::ToSocketAddrs
+    pub fn bind<A: ToSocketAddrs>(addr: A) -> io::Result<Self> {
```
Noah-Kennedy, line 51: Let's keep this as a SocketAddr for now, to prevent blocking.

src/net/tcp/stream.rs
```diff
@@ -49,12 +47,19 @@ impl TcpStream {
...
     /// connection attempt (the last address) is returned.
     ///
-    /// [`ToSocketAddrs`]: trait@tokio::net::ToSocketAddrs
-    pub async fn connect(socket_addr: SocketAddr) -> io::Result<TcpStream> {
-        let socket = Socket::new(socket_addr, libc::SOCK_STREAM)?;
-        socket.connect(socket_addr).await?;
-        let tcp_stream = TcpStream { inner: socket };
-        Ok(tcp_stream)
+    /// [`ToSocketAddrs`]: trait@std::net::ToSocketAddrs
+    pub async fn connect<A: ToSocketAddrs>(addr: A) -> io::Result<TcpStream> {
```
Noah-Kennedy, line 51: Same as with `bind`

src/net/unix/listener.rs
```diff
@@ -0,0 +1,78 @@
...
+    pub async fn accept(&self) -> io::Result<UnixStream> {
+        let (socket, _) = self.inner.accept().await?;
+        let stream = UnixStream { inner: socket };
+        Ok(stream)
+    }
+}
+
+impl std::ops::Drop for UnixListener {
+    fn drop(&mut self) {
+        // If the file could not be deleted, we tried our best
```
Noah-Kennedy, line 71: Let's not have a drop implementation. We should keep the behavior the same here as with tokio and std.

Noah-Kennedy:
This is almost there, just a couple more changes are needed.

Noah-Kennedy (changes requested):
I left some feedback comments for what needs addressed.

src/net/tcp/listener.rs
jon-chuang: I don't think bind is an operation on the critical path. Personally, I think use of the trait makes API more ergonomic.

Could you name a scenario where you are continually binding a new port frequently?

src/driver/socket.rs
Noah-Kennedy: @jon-chuang can we do this without the raw syscall here?

src/net/tcp/listener.rs
jon-chuang: Actually, after some reflection, I understand that this blocking call could block the whole thread. So I guess I am in agreement.

src/driver/socket.rs
jon-chuang: My bad, missed this one

Noah-Kennedy (approved):
Looks good, thanks for doing this!

Noah-Kennedy:
@jon-chuang run `cargo fmt` and push the formatting changes to appease the CI gods.

PR #78, opened 2022-03-14
driver: batch submit requests and add benchmark

Noah-Kennedy:


src/runtime.rs
```diff
@@ -78,12 +86,26 @@ impl Runtime {
                 }
             };
 
+            let wakeup = async {
```
carllerche, line 89: Why does this need to be in a timer and in park isn't sufficient? Is it because the `on_thread_park` callback isn't always invoked?

Noah-Kennedy: from discord: I added that to make sure that we don't get starved in the case where we have futures getting woken by things which aren't io_uring. We could be starved if there was always work to do and if we don't push enough entries to force a flush of the SQ.

Noah-Kennedy:
@carllerche can you take a look here?

src/runtime.rs
carllerche: 100ms is pretty high... Could we add alternative hooks? Maybe when a task is done being polled?

Noah-Kennedy: Splitting this into another PR

carllerche (approved):
Seems ok thanks.

PR #85, opened 2022-06-08
`from_std(...)` for UdpSocket

markmandel:
I wanted the ability to have finer grained control over which socket options where set when creating a UdpSocket, so implemented a similar API surface to that which exists on `tokio::net::UdpSocket` for the same reason.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,2 @@
 /target
 Cargo.lock
-.idea
-*.iml
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,4 @@
 /target
 Cargo.lock
+.idea
+*.iml
--- a/src/net/udp.rs
+++ b/src/net/udp.rs
@@ -92,6 +92,62 @@ impl UdpSocket {
         Ok(UdpSocket { inner: socket })
     }
 
+    /// Creates new `UdpSocket` from a previously bound `std::net::UdpSocket`.
+    ///
+    /// This function is intended to be used to wrap a UDP socket from the
+    /// standard library in the Tokio equivalent. The conversion assumes nothing
+    /// about the underlying socket; it is left up to the user to set it in
+    /// non-blocking mode.
+    ///
+    /// This can be used in conjunction with socket2's `Socket` interface to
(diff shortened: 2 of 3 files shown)


markmandel:
Doh. Editor wasn't set to format on save. Formatting fixes all fixed now.

.gitignore
```diff
@@ -1,2 +1,4 @@
 /target
 Cargo.lock
+.idea
```
Noah-Kennedy, line 3: Let's not add editor-specific stuff here. I'd suggest using a global gitignore on your device instead.

src/net/udp.rs
```diff
@@ -92,6 +92,62 @@ impl UdpSocket {
         Ok(UdpSocket { inner: socket })
     }
 
+    /// Creates new `UdpSocket` from a previously bound `std::net::UdpSocket`.
+    ///
+    /// This function is intended to be used to wrap a UDP socket from the
+    /// standard library in the Tokio equivalent. The conversion assumes nothing
+    /// about the underlying socket; it is left up to the user to set it in
+    /// non-blocking mode.
```
Noah-Kennedy, line 100: Nonblocking isn't as meaningful in the context of io-uring.

Noah-Kennedy (changes requested):

.gitignore
markmandel: TIL about global gitignore. On it.

src/net/udp.rs
markmandel: Took a stab at an updated description that isn't as specific - plesae let me know what you think.

Noah-Kennedy: So, the requirement to set things to nonblocking mode is more relevant in the context of epoll, where the strategy is to basically loop on attempting a non-blocking read/write up front, and, if that fails, registering with the IO driver's epoll loop to be woken once readiness is signaled, and to then go back to attempting an IO op again, which mail once again fail, and so on...

Noah-Kennedy (approved):
Sorry for leaving you hanging; this looks good. Thanks for the PR!

PR #87, opened 2022-07-05
Implement fs::File::readv_at()/writev_at()

jiangliu:
These two methods are very convenient to implement Fuse servers. 
Typical FUSE reply messages are composed up by <COMMON HEADER,  OP HEADER, PAYLOAD DATA>,
writev_at() helps to avoid memory copy operations.

### DIFF
--- a/src/driver/mod.rs
+++ b/src/driver/mod.rs
@@ -14,6 +14,8 @@ mod open;
 
 mod read;
 
+mod readv;
+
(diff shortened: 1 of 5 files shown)


Noah-Kennedy:
Sorry I forgot about this one. If this happens again, please ping me so that I come back to this.

Noah-Kennedy (approved):

Issue #48, opened 2021-08-19
Add missing filesystem operations


kaimast:
It would be great if we could add the most important functions from `tokio::fs`/`std::fs` to this crate. As far as I can tell `tokio-uring` only exposes `fs::File` right now and none of the ones below are supported yet.

Directories:
- [x] create_dir
- [ ] read_dir
- [x] remove_dir
- [ ] remove_dir_all

Files:
- [ ] metadata
- [x] rename
- [ ] copy
- [x] remove_file

Links (I don't really need these but would be great to have):
- [ ] symlink_metadata
- [ ] soft_link
- [ ] hard_link
- [ ] read_link

This should be fairly straightforward to implement, right? I can try to take a stab at some of these once I find some time.

kaimast:
Is there going to be another `io-uring` release soonish? I just added a `create_dir` call on [this branch](https://github.com/kaimast/tokio-uring/tree/create-dir), but it requires the git version of that crate.

Noah-Kennedy:
Yes, there should be a release soon.

Noah-Kennedy:
@kaimast version 0.2.0 has been released.

kaimast:
Thanks! I was talking about the `io-uring` crate though.

kaimast:
We should maybe keep this open until the others are implemented too?

FrankReh:
That close seemed to have been done automatically. Anyway, no harm. Definitely some file system operations left.

Icelk:
Is `tokio::fs::canonicalize` also an operation we can support?

FrankReh:
> Is `tokio::fs::canonicalize` also an operation we can support?

The io_uring API doesn't support anything like a canonicalize (or in Unix terms, a realpath) so to avoid using the std or the tokio canonicalize, this crate, or just as easily I think, a 3rd party package, could build the logic from this crate's `StatxBuilder`.

That's not to say we wouldn't accept a PR that essentially duplicated a realpath based on our statx. But any suggested additions would have to be compatible with our license and I don' know if even starting with a BSD version of realpath is an acceptable approach.

Icelk:
I understand. Thanks.

PR #96, opened 2022-07-29
fs: implement `rename`

arashsm79:
Implement the `rename` operation for `fs`.
This operation uses [`IORING_OP_RENAMEAT`](https://man.archlinux.org/man/io_uring_enter.2.en#IORING_OP_RENAMEAT) as the opcode for the submitted
SQE.

Both file descriptors are set to the special value `AT_FDCWD` so that
the given paths are interpreted relative to the current working
directory of the calling process. [[1]](#1)

The operation is submitted with a zero `flags` argument so that it would
be equivalent to the corresponding operation in [`tokio:fs`](https://docs.rs/tokio/latest/tokio/fs/fn.rename.html)/[`std:fs`](https://doc.rust-lang.org/std/fs/fn.rename.html). [[1]](#1)

Refs: #48

<a id="1">[1]</a> [`renameat2` syscall man page](https://man7.org/linux/man-pages/man2/renameat2.2.html)

### DIFF
--- a/src/driver/rename_at.rs
+++ b/src/driver/rename_at.rs
@@ -1,0 +1,40 @@
+use crate::driver::{self, Op};
+
+use std::ffi::CString;
+use std::io;
+use std::path::Path;
+
+/// Renames a file, moving it between directories if required.
+/// The given paths are interpreted relative to the current working directory
+/// of the calling process.
+pub(crate) struct RenameAt {
+    pub(crate) from: CString,
+    pub(crate) to: CString,
+}
+
+impl Op<RenameAt> {
+    /// Submit a request to rename a specified path to a new name with
+    /// the provided flags.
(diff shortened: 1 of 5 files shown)


src/driver/rename_at.rs
```diff
@@ -0,0 +1,40 @@
+use crate::driver::{self, Op};
+
+use std::ffi::CString;
+use std::io;
+use std::path::Path;
+
+/// Renames a file, moving it between directories if required.
+/// The given paths are interpreted relative to the current working directory
```
Noah-Kennedy, line 8: Add a newline between the first and second line.

Noah-Kennedy (changes requested):
One minor thing, but this is great!

arashsm79:
Done.

Noah-Kennedy (approved):

PR #97, opened 2022-07-31
io: add Submitted and Unsubmitted

Noah-Kennedy:
This adds support for "raw operations" to be used, and unblocks work on a more powerful builder-style API for provided buffers and attaching flags.


Noah-Kennedy:
Gonna back this out now that I am taking a slightly different approach here.

Issue #106, opened 2022-09-02
Implement local_addr for TcpListener and UnixListener


FrankReh:
I don't believe io_uring can be used to get the local_addr from a socket that is TCP listening or Unix listening, but the ability would be useful nonetheless.

Is it okay to provide blocking method calls to return the local_addr for the two types?

If so, new unit tests would not need to hardcode port numbers in their TCP tests. Existing unit tests could also be modified.

PR #107, opened 2022-09-04
net: add TcpListener.local_addr method

FrankReh:
The local_addr method is a blocking call.

### DIFF
--- a/src/net/tcp/listener.rs
+++ b/src/net/tcp/listener.rs
@@ -40,13 +40,41 @@ impl TcpListener {
     ///
     /// Binding with a port number of 0 will request that the OS assigns a port
     /// to this listener.
-    ///
-    /// In the future, the port allocated can be queried via a (blocking) `local_addr`
-    /// method.
     pub fn bind(addr: SocketAddr) -> io::Result<Self> {
         let socket = Socket::bind(addr, libc::SOCK_STREAM)?;
         socket.listen(1024)?;
         Ok(TcpListener { inner: socket })
+    }
+
+    /// Returns the local address that this listener is bound to.
+    ///
+    /// This can be useful, for example, when binding to port 0 to figure out
+    /// which port was actually bound. The implementation relies on the standard net
+    /// system call so is blocking.
+    ///
+    /// # Examples
+    ///
+    /// ```
+    /// use std::net::{Ipv4Addr, SocketAddr, SocketAddrV4};
+    /// use tokio_uring::net::TcpListener;
+    ///
+    /// let listener = TcpListener::bind("127.0.0.1:8080".parse().unwrap()).unwrap();
+    ///
+    /// let addr = listener.local_addr().expect("Couldn't get local address");
+    /// assert_eq!(addr, SocketAddr::V4(SocketAddrV4::new(Ipv4Addr::new(127, 0, 0, 1), 8080)));
+    /// ```
+    #[cfg(unix)]
+    pub fn local_addr(&self) -> io::Result<SocketAddr> {
+        use std::os::unix::io::{AsRawFd, FromRawFd};
+
+        let fd = self.inner.as_raw_fd();
+        // SAFETY: Our fd is the handle the kernel has given us for a TcpListener.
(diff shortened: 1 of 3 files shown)


FrankReh:
Fixes TCP portion of #106 

But maybe there is a more standard way to cast the fd to a net::TcpListener?

src/net/tcp/listener.rs
```diff
@@ -40,15 +40,43 @@ impl TcpListener {
...
-    /// method.
     pub fn bind(addr: SocketAddr) -> io::Result<Self> {
         let socket = Socket::bind(addr, libc::SOCK_STREAM)?;
         socket.listen(1024)?;
         Ok(TcpListener { inner: socket })
     }
 
+    /// Returns the local address that this listener is bound to.
+    ///
+    /// This can be useful, for example, when binding to port 0 to figure out
```
Noah-Kennedy, line 51: I wouldn't exactly call this blocking. Certain system calls (like this one) are fast enough that they can be considered nonblocking. This one will simply drop to kernelspace to grab some data about the socket and return back.

It's probably even less expensive than another syscall we generally agree to be nonblocking, the `recv` syscall (in nonblocking mode), which does far more work in kernelspace than this one!

src/net/tcp/listener.rs
```diff
@@ -40,15 +40,43 @@ impl TcpListener {
     ///
     /// Binding with a port number of 0 will request that the OS assigns a port
     /// to this listener.
-    ///
```
Noah-Kennedy, line 43: Yeah I think this comment was kinda wrong tbh anyways.

src/net/tcp/listener.rs
```diff
@@ -40,15 +40,43 @@ impl TcpListener {
...
+    /// ```
+    /// use std::net::{Ipv4Addr, SocketAddr, SocketAddrV4};
+    /// use tokio_uring::net::TcpListener;
+    ///
+    /// let listener = TcpListener::bind("127.0.0.1:8080".parse().unwrap()).unwrap();
+    ///
+    /// let addr = listener.local_addr().expect("Couldn't get local address");
+    /// assert_eq!(addr, SocketAddr::V4(SocketAddrV4::new(Ipv4Addr::new(127, 0, 0, 1), 8080)));
+    /// ```
+    #[cfg(unix)]
```
Noah-Kennedy, line 66: Not sure if the `#[cfg(unix)]` is necessary considering that this crate only works on unix in the first place.

Noah-Kennedy (changes requested):

Noah-Kennedy:
Some minor feedback, but overall this looks good!

src/net/tcp/listener.rs
FrankReh: Agreed.

src/net/tcp/listener.rs
FrankReh: I wasn't sure either but thought it safer to include so you could comment. Happy to remove it.

src/net/tcp/listener.rs
```diff
@@ -40,15 +40,43 @@ impl TcpListener {
...
+    pub fn local_addr(&self) -> io::Result<SocketAddr> {
+        use std::os::unix::io::{AsRawFd, FromRawFd};
+
+        let fd = self.inner.as_raw_fd();
+        // SAFETY: Our fd is the handle the kernel has given us for a TcpListener.
```
FrankReh, line 71: Is this SAFETY comment too much? Happy to replace if you want to suggest an alternative.

FrankReh:
Think I handled your first set of comments.

I added a question about the SAFETY comment. Is it too wordy?

And importantly, I've added the local_addr method for the UnixListener now but have only tested it with the document example.

src/net/tcp/listener.rs
Noah-Kennedy: Yeah I'd recommend removing it

src/net/tcp/listener.rs
Noah-Kennedy: This is great!

Noah-Kennedy (approved):

PR #113, opened 2022-09-15
driver: add Builder API as an option to start

### LINKED ISSUE #105: io_uring setup parameters
Currently, tokio_uring::start calls runtime::Runtime::new which calls the io_uring::IoUring::new with an entries count value of 256. But there are many settings besides the entries value that one may want to set when creating the kernel io_uring device.

The io_uring crate provides a builder API for setting those parameters that then get passed to the kernel's io_uring setup; but there is no API in tokio_uring to use that builder API for the parameters.

How should the tokio_uring support use of the io_uring builder api for setup parameters?

Can it reexport the io-uring Builder and pass that to another version of the tokio_uring::start (perhaps start_builder(b: io_uring::Builder)? Or should it make its own Builder that copies all the methods from the io-uring Builder, or should it provide a few mixes of parameter settings that are thought to be more safe or more tested or more useful than the rest?

FrankReh:
Add a Builder API as an option to start the tokio_uring runtime.

The preexisting tokio_uring::start function is maintained.

A new start mechanism is introduced that allows specifying the parts that the runtime start and the driver initialize use.

An example of the new usage is provided in the Builder::start method.

A goal of this Builder was to remain independant of the io-uring Builder progress and allow the caller to build their io-uring options and pass them to this crate's Builder. As new kernel versions support more io_uring setup parameters, the tokio io-uring crate will add more Builder options and they should not require lock-step changes in this crate.

Fixes #105

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -140,8 +140,94 @@ use std::future::Future;
 /// }
 /// ```
 pub fn start<F: Future>(future: F) -> F::Output {
-    let mut rt = runtime::Runtime::new().unwrap();
+    let mut rt = runtime::Runtime::new(&builder()).unwrap();
     rt.block_on(future)
+}
+
+/// Create and return an io_uring::Builder that can then be modified
+/// through its implementation methods.
+///
+/// This function is provided to avoid requiring the user of this crate from
+/// having to use the io_uring crate as well. Refer to Builder::start example
+/// for its intended usage.
+pub fn uring_builder() -> io_uring::Builder {
+    io_uring::IoUring::builder()
+}
+
+/// Builder API to allow starting the runtime and creating the io_uring driver with non-default
+/// parameters.
+// #[derive(Clone, Default)]
+pub struct Builder {
+    entries: u32,
+    urb: io_uring::Builder,
+}
+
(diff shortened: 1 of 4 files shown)


FrankReh:
I'm always open to name changes and comment changes. My comments could sometimes be too on the nose, while my local variable names could be too short for some.

FrankReh:
@Noah-Kennedy Github is confusing me. Do I need to do something else to get this reviewed by someone on the team?

src/lib.rs
```diff
@@ -140,10 +140,96 @@ use std::future::Future;
...
+    ///
+    ///             loop {
+    ///                 let (socket, _) = listener.accept().await?;
+    ///                 // process socket
+    ///             }
+    ///         }
+    ///     )
+    /// }
+    /// ```
+    pub fn start<F: Future>(&self, future: F) -> F::Output {
```
Noah-Kennedy, line 227: Ideally we would expose the runtime, but for now this works.

Noah-Kennedy (approved):
@FrankReh I had meant to get to this earlier, but ended up being quite busy. If something like this looks like it's falling through, please ping me like you did here.

Issue #112, opened 2022-09-14
Use of IORING_REGISTER_PBUF_RING, introduced in kernel 5.19


FrankReh:
An issue to track this crate's use of the IORING_OP_

First described here [io_uring: add support for ring mapped supplied buffers](https://lore.kernel.org/io-uring/20220429122803.41101-11-axboe@kernel.dk/#r)

FrankReh:
Only today I started to appreciate the value of the buf_ring feature added in 5.19. That was introduced in the kernel archives in April this year and perhaps I had been reading older man pages to begin with. I had already been wondering how we would want to compact the information when we are replenishing fixed buffers. Turns out the io_uring author had already worked on it earlier this year.

I would propose this crate prepare to jump right to this use of fixed buffers, using the pbuf_ring mechanism (not use the provide-buffers mechanism), even though it will only be available for kernels 5.19+. This crate doesn't need to support fixed buffers on older kernels IMHO as they would be much more work, and their overhead large enough they aren't clearly a win for users.

FrankReh:
The pbuf_ring mechanism seems to be the fourth form of buffer management provided by the kernel io_uring driver (as of 5.19). (Not even counting the mechanism where the caller provides their own buffer for each operation - that could be thought of a mechanism 0.)

1)
IORING_REGISTER_BUFFERS for use with two operators only IORING_OP_READ_FIXED and IORING_OP_WRITE_FIXED.

2)
 IORING_OP_PROVIDE_BUFFERS came in kernel 5.7 and allowed the creation of buffer groups that could then be referenced with various read and recv operations. But replenishing the provided buffers was not considered easy or efficient by April of 2022. 

3)
IORING_REGISTER_BUFFERS2 introduced in kernel 5.13. Seems useful only for the same two operations as listed above: IORING_OP_READ_FIXED and IORING_OP_WRITE_FIXED.

4)
IORING_REGISTER_PBUF_RING introduced in kernel 5.19. Provides for buffers per group id, like 2) but with a much more efficient mechanism of replenishing buffers. The link at the top of this issue provides the benchmarks showing how much more efficient this is compared with 2) for the system as a whole. And it is easier for the user-space caller as well. Rings of buffers to be used with the rings of submissions and completions.

FrankReh:
As another reference, here is the liburing UDP example using the io_uring_buf_ring mechanism. https://github.com/axboe/liburing/blob/master/examples/io_uring-udp.c

[...]

Issue #104, opened 2022-08-29
Multishot operations


Noah-Kennedy:
io_uring now has multiple multishot operations, including IORING_OP_ACCEPT and IORING_OP_POLL_ADD. We should add support for streaming back completions for such operations.

FrankReh:
Mind suggesting a unit test for the single shot poll_add? I would build off that for the function poll_add itself and then the *multi* version .

Noah-Kennedy:
@FrankReh I think we need to start with asking what the single-shot API should even look like. We can do this as a future, but many multi-shot operations are going to be more useful in the context of us doing something like implementing the (inherently readiness-based) AsyncRead and AsyncWrite APIs, in which case we may need to build something akin to the tokio Registration API at some point.

To be honest, I kinda wanna just try shoving this into a tokio fork. This is hard to do without first forking mio though. Hmm...

FrankReh:
@Noah-Kennedy The single shot and multi shot poll operation (POLL_ADD) could be handy for watching additional urings and tying their file descriptors to the current-thread's default uring. Sounds like glommio does something similar with their secondary uring. I had been wondering how tokio-uring could be expanded to support multiple urings. Perhaps such an ability does not have to built into tokio-uring, per se. A task could build its own uring, somehow, and loop on a poll stream for that uring's file descriptor. Any type of io_uring platform could be used for those extra urings (maybe) since it would be their completion queue status that was being monitored by the primary uring. And that would indicate some designs that tie io_uring functionality with futures should be done independently of the runtime - perhaps something that sits in between the bindings that io-uring provides, and the assumptions that tokio-uring makes.

Still catching up to your idea regarding tokio Registrations...

Noah-Kennedy:
@FrankReh I had been thinking that OP_POLL_ADD would be most effective for compatibility with readiness-based APIs. My thoughts on multi-uring runtimes would be to have a worker per core and to use the inter-ring messaging facilities.

FrankReh:
@Noah-Kennedy Thanks for the feedback.

[...]

PR #123, opened 2022-10-04
Support zero-copy send

ollie-etl:
Linux 6.0 introduced `IORING_OP_SEND_ZC`. This PR introduces the function `send_zc` on udp (todo: tcp) sockets which takes advantage of it.

~This is blocked on this [PR](https://github.com/tokio-rs/io-uring/pull/147) on io-uring~ *Merged*

As IORING_OP_SEND_ZC can return multiple CQE's per SQE, this also extends the `Lifecycle` handling in `impl Fut for Op<T>` to check for these occurences. If does this using a new method `update` in `Completable`, which I added to minimise the scope of the changes - it just falls back to the `complete` method for implementations which are just single entry.

This approach could probably be used as a the basis of streaming interfaces for some other submission types.

Since being split, is now blocked on #130


ollie-etl:
@FrankReh I'd appreciate your thought on the driver changes here, w.r.t support for  #112, which has the same 1 : many SQE/SQE behaviour.  

In the udp send_zc case, I believe the number of completions which can occur is limited to 64, which is the GSO segment limit. In the case of multishot recv, the limit is the number of buffers pre-registered and available to the ring. would that be deemed acceptable?

src/driver/op.rs
```diff
@@ -166,16 +218,30 @@ impl Lifecycle {
...
+                    true
+                }
+            },
+            Lifecycle::Completed(mut v) => {
+                // We've run a data race between a multi-completion event and
+                // the first call to poll().
+                v.push((result,flags));
+                *self = Lifecycle::Completed(v);
+                false
+            }
```
ollie-etl, line 243: I believe the growth of the smallvec is only bound by the number of CQE's which can occur without acknowledgement
this is one of:
  - <= 64 (for UDP send_zc, with 64 GSO segments)
  - ? for tcp
  - the number of pre-registered buffers, assuming something like multishot recv is also implemented here
  
I think the idea would be able to force the future to consume the results, but this is problematic for 2 reasons
  - we have no info on the op here
  - it would pose a problem should a stream be implemented atop this, as you are likely to want to emit results based on partial updates
  
It might also possible perform op poll during creating, registering the waker prior to submission, eliminating the Submitted state, and adding a waker state to Completed.

ollie-etl:
@Noah-Kennedy , do you have any thoughts on this?

FrankReh:
@ollie-etl I have a fork that implements a POC for streaming the results of multi operations. I'll try to get my forks of the two relevant crates committed tomorrow so you and Noah can have a look. Do you have an example that uses this new operation yet? I could try it on my fork.

ollie-etl:
@FrankReh Its now in use on our main dev branch, shortly to go into production. 

I could probably carve off a small benchmark or example and commit, but i won't have a chance until early next week. 

We see a ~0.9->3x performance gain over `write`. This will depend hugely on your hardware setup obviously, and the packets you shift. The best comes, as you would expect, when doing large buffers (64 kB) with hardware GSO enabled. Note that local loopback is not a good way to test the performance of this - it won't do zero copy (and will fall back to copying) - I think you actually need a NIC with scatter gather support.

Critically, this now exceeds the performance we managed to get with sendmmsg with zero copy + GSO. Thats not to say a better benchmark implementation couldn't of course.

I've literally no recent experience using TCP, so I can't comment there. Maybe it's even better, if larger segments can be offloaded? There is a reason I didn't add it to the tcp socket implementation here, I don't know the gotchas as well.

oliverbunting:
Still me, not on a work account. @FrankReh  thank you for your concidered reply 

> First question, why didn't you want to expose the CQE_F_NOTIF bit flag the way you did the more flag? I think the tokio-uring crate should use the CQE_F_NOTIF bit flag the way the liburing tests for send_zc do.

Oversight really. It's not strictly needed either. you either get CQE_F_NOTIF or CQR_F_MORE, so testing for one is sufficient.

> Did you mean issue #104, Multishot operations? I was confused for a bit how these changes are related to buffering.


I misunderstood the scope of the issue.


> I've cloned your branch and squashed the six commits. It was hard to see what was going on because the 'fix for a race' came in later. But I have to admit I don't understand the Lifecycle changes. A few comments in the poll area vs the complete area could go a long way.
> 
> For example, I don't understand how the complete Ignored(op) match is the only one that looks at the More flag, but probably that's because my own handling of More flags was being done differently - in a branch I hadn't shared with anyone yet.

Lifetime doesn't know what to do with a completion. So all it does when called by the completion handling logic is store them in a smallvec, and await for Poll to deal with them. 

smallvec was chosen to avoid allocating for the default case (single completion) and when you're actively polling, when you're less likely to have many unacknowledged completions.

The value of 4 was chosen arbitrarily to be small, so completion state should still fit on a cache line. 


> 
> Other thoughts regarding naming and comments, so they aren't very important as Noah will be the final arbiter, are why go with the name 'update/Update' rather than more and More. And why does the Completable trait get a default update that creates an Update::Finished. Probably obvious to someone writing it, but a comment about why it is the right default and what the user of the trait might want to do when providing their own update method would be very helpful.

I added update to avoiding breaking API changes on existing single completion types.

My rational is that update consumes the completion, updating state, and returns if it needs more or has finished. For the case for single  completion event types, this is just complete.

> 
> What is the exposure of tokio to the smallvec crate and is the default size of four that important? Most uses of tokio-uring operations will be things other than zero copy send so maybe the some other way of burdening only the operations that need the extra flexibility? I've created a mirror type to Op, named StrOp, for those operations that will result in streaming responses - but Noah has yet to sign off on that yet too, as I haven't pushed that up to my GitHub repo for him to see.

I could create a MultiCompleted enum varient. I don't believe smallvec will incur significant overhead, but I'm more than happy to put it in its own variant.

Vec would be a bad idea I believe. That would incur significant numbers of allocations, which would wipe out gains seen by zero copy. Micro seconds matter here, for me anyway.

> 
> And about the use of *fold* on the iterator: its used twice, once in the poll path, once in the cqe complete path, it looks like the logic is designed to iterate through a growing smallvec until the Finished state is reached. Is this because you didn't want to modify the Lifecycle enum states too much? I went a different route for handling 'more' with streams so no growing loops like that were necessary. I did track a vec of more results along the way, to deal with something of the same issue you are tackling  here.

I'm just consuming all completions currently recieved, in order, and checking if we still need more or are finished. I've no knowledge about how many mores I'll get before the op completes. A for loop would probably be better, but I'm a recovering Haskell developer.

> 
> Generally, I would say the send_zc and sendmsg_zc are very unusual operations for the io_uring driver at this point and I wouldn't mind a separate type of Lifecycle for such operations, defined in their own module. I've named my own new module strop, for streaming Op, and could leave the original Op basically unchanged. There is some logic duplication between op.rs and strop.rs, but as the nature of the streaming operations are so different from the single shot operations, I favored creating a new Lifecycle implementation. It does mean the entry in the slab has an extra level enum to split between the original Op and the new StrOp. Noah and I will have some work to do to get on the same page as  that decision comes with its own long term maintenance questions. But if that design is generally accepted, perhaps a NotiferOp or a ZcOp would allow special tailoring for special operations like this one.

I'm very keen to see how you avoided growing a collection of results in the presence of the race between the ring and Poll. That sounds better.

I also think send_zc could have just been a change to send. It is after all just a socket option on the normal syscall. I decided not to do that due to the scope of that change.

Noah-Kennedy:
I think we should split this into two PRs. One for the multishot support in the driver, and another for the ZC send ops.

oliverbunting:
> I think we should split this into two PRs. One for the multishot support in the driver, and another for the ZC send ops.

Ok, happy to do so.

W.r.t the driver bit, do you hold any opinions on the areas discussed above?

ollie-etl:
@FrankReh @Noah-Kennedy  I've updated this based on #130. I think its cleaner, and shows off the Op specializations .

FrankReh:
@ollie-etl I haven't forgotten to look at this. Just want to get a little closure on my buf_ring work first before loosing focus.

Noah-Kennedy:
I have been pretty busy all week, but I should be able to get to this over the weekend.

src/driver/op.rs
```diff
@@ -183,6 +191,76 @@ where
...
+                        driver.ops.remove(me.index);
+                        me.index = usize::MAX;
+                        Poll::Ready(me.data.take().unwrap().complete(cqe))
+                    }
+                    Lifecycle::CompletionList(indices) => {
+                        let mut data = me.data.take().unwrap();
+
+                        // Traverse the CqeResult list, calling update on the Op on all Cqe's flagged `more`
+                        // If the final Cqe is present, return it in an Option
+                        let last = indices.into_list(completions).find(|cqe| {
```
FrankReh, line 238: This find doesn't actually look at all the entries does it? It finishes as soon as the closure returns true, returning that cqe. So theoretically more than one entry could have no 'more' flag set. I don't quite know what to make of that. Just seems like things aren't completely tight. Just thinking out load here.

src/driver/op.rs
```diff
@@ -183,6 +191,76 @@ where
...
+                        // and the final CQE registered as Completed
+                        driver.ops.remove(me.index);
+                        me.index = usize::MAX;
+                        Poll::Ready(me.data.take().unwrap().complete(cqe))
+                    }
+                    Lifecycle::CompletionList(indices) => {
+                        let mut data = me.data.take().unwrap();
+
+                        // Traverse the CqeResult list, calling update on the Op on all Cqe's flagged `more`
+                        // If the final Cqe is present, return it in an Option
```
FrankReh, line 237: Rather than 'Traverse', could this say 'Consume' the completion list?

And the second line, something like

// If the final Cqe is present, cleanup and return Poll::Ready?

Or is that too much on the nose?

src/driver/op.rs
```diff
@@ -40,11 +40,19 @@ pub(crate) struct Op<T: 'static, CqeType = SingleCQE> {
...
+/// A Marker for Operations will process multiple completion events,
+/// which combined resolve to a single Future value
+pub(crate) struct MultiCQEFuture;
+
 pub(crate) trait Completable {
     type Output;
     fn complete(self, cqe: CqeResult) -> Self::Output;
 }
 
+pub(crate) trait Updateable: Completable {
```
FrankReh, line 52: I believe a comment is warranted for future designers. Something to the effect that the 'update' in this Trait, when used in the MultiCQEFuture poll, will be called when the future is polled, on all 'more' entries found in the completion list, draining the list in the one poll cycle.

Because another design choice could have been to have 'update' called when the driver read another cqe from the cqueue and not keep the slab linked list at all. ... Well, come to think of it, I'm not sure the driver's complete would have access to the 'update', maybe it doesn't know the type of the op.

Not to mention some other op implementations, like streams, would want each 'more' cqe sent back as the result of the poll, so the completion list wouldn't be drained immediately.

So maybe this is just me thinking out loud again.

src/driver/send_zc.rs
```diff
@@ -0,0 +1,82 @@
...
+    buf::IoBuf,
+    driver::{Op, SharedFd},
+    BufResult,
+};
+use std::{
+    io,
+    task::{Context, Poll},
+};
+
+pub(crate) struct SendZc<T> {
```
FrankReh, line 12: That's a lot of work to get one new type of operation in that looks so clean in and of itself. So it goes.

src/driver/socket.rs
```diff
@@ -53,6 +53,11 @@ impl Socket {
         op.await
     }
 
+    pub(crate) async fn send_zc<T: IoBuf>(&self, buf: T) -> crate::BufResult<usize, T> {
+        let op = Op::send_zc(&self.fd, buf).unwrap();
+        op.send().await
```
FrankReh, line 58: This could probably become 'op.await'.

src/net/udp.rs
```diff
@@ -175,6 +175,12 @@ impl UdpSocket {
         self.inner.send_to(buf, socket_addr).await
     }
 
+    /// Sends data on the socket. Will attempt to do so without intermediate copies.
+    /// On success, returns the number of bytes written.
```
FrankReh, line 179: I haven't read about the zero copy operation added to the uring interface. Is there any reason an app would still prefer to use send rather than send_zc? Maybe we could give some guidance here when one might use one over the other.

FrankReh (changes requested):

src/driver/op/slab_list.rs
```diff
@@ -120,6 +120,14 @@ impl<'a, T> Drop for SlabList<'a, T> {
     }
 }
 
+impl<'a, T> Iterator for SlabList<'a, T> {
+    type Item = T;
+
+    fn next(&mut self) -> Option<Self::Item> {
+        self.pop()
```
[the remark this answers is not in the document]
ollie-etl: yes

src/driver/op.rs
```diff
@@ -40,11 +40,19 @@ pub(crate) struct Op<T: 'static, CqeType = SingleCQE> {
...
+    fn update(&mut self, cqe: &CqeResult);
```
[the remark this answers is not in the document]
oliverbunting: That might be an oversight

src/driver/op.rs
oliverbunting: NN o, but then if we're receiving more cqes without more flagged, there is a bug in the kernel

src/driver/send_zc.rs
oliverbunting: Well, it was a lot easier when it just allocated a smallvec 🙄

ollie-etl:
@FrankReh I think the comments are addressed?

src/driver/op.rs
```diff
@@ -183,6 +194,78 @@ where
...
+                                let _ = me.data.insert(data);
+                                *lifecycle = Lifecycle::Waiting(cx.waker().clone());
+                                Poll::Pending
+                            }
+                            Poll::Ready(cqe) => {
+                                me.index = usize::MAX;
+                                driver.ops.remove(me.index);
+                                Poll::Ready(data.complete(cqe))
+                            }
+                        }
```
ollie-etl, line 261: This is pulled out of the for loop to allow access to driver.ops, which is otherwise prevented by the borrow checker by the borrow of lifecycle.

src/driver/op.rs
```diff
@@ -183,6 +194,78 @@ where
...
+                        match status {
+                            Poll::Pending => {
+                                // We need more CQE's. Restore the op state
+                                let _ = me.data.insert(data);
+                                *lifecycle = Lifecycle::Waiting(cx.waker().clone());
+                                Poll::Pending
+                            }
+                            Poll::Ready(cqe) => {
+                                me.index = usize::MAX;
+                                driver.ops.remove(me.index);
```
FrankReh, line 258: You probably meant the other order, so me.index is changed after this remove line.

src/net/udp.rs
FrankReh: So something would be worth saying, either about the tradeoffs as you described them in this review, or how one might reason about using one or the other.

src/driver/op.rs
FrankReh: How about

// `update` is called for the intermediate results when the future is polled. Before being polled, the intermediate results build up in a completions list.

FrankReh (changes requested):

src/driver/op.rs
ollie-etl: eek! good spot

ollie-etl:
@FrankReh thanks for the quick review. I think those are done now

ollie-etl:
Just adding a link to #54 to mark as relevant for when that merges.

FrankReh (approved):
Nice. Thanks for your patience. Nice addition.

Noah-Kennedy:
Let's hold off on merge for a little bit. We might need to add some more code around multishot ops to make shutdown work properly.

FrankReh:
@ollie-etl 
Now just wondering what the squashed commit message should say.

Care to offer a title and short paragraph here?

ollie-etl:
@Noah-Kennedy : this isn't like a multi-shot op in that sense. There is no uncertainty about the state like an `accept` call - if you've submitted a send_zc, then in fairly short order you'll get some cqe's

Noah-Kennedy:
@ollie-etl @FrankReh due to shutdown changes that had to be made, we can't merge this right now. We need a way to tell if a particular multishot completion is finished. I'll open up a PR tonight with this.

Noah-Kennedy (changes requested):
@ollie-etl @FrankReh due to shutdown changes that had to be made, we can't merge this right now. We need a way to tell if a particular multishot completion is finished. I'll open up a PR tonight with this.

FrankReh:
@ollie-etl FYI
Ran across this kernel submission. It says zero copy is supported for UDP and TCP, you knew that. And it says when it is not supported, it will no longer fall back to a copy but will return an error. Don't know if you knew that yet.

```
commit 942e01ab90151a16b79b5c0cb8e77530d1ee3dbb
Merge: d47136c28015 cc767e7c6913
Author: Linus Torvalds <torvalds@linux-foundation.org>
Date:   Sun Oct 23 09:55:50 2022 -0700

    Merge tag 'io_uring-6.1-2022-10-22' of git://git.kernel.dk/linux
    
    Pull io_uring follow-up from Jens Axboe:
     "Currently the zero-copy has automatic fallback to normal transmit, and
      it was decided that it'd be cleaner to return an error instead if the
      socket type doesn't support it.
    
      Zero-copy does work with UDP and TCP, it's more of a future proofing
      kind of thing (eg for samba)"
    
    * tag 'io_uring-6.1-2022-10-22' of git://git.kernel.dk/linux:
      io_uring/net: fail zc sendmsg when unsupported by socket
      io_uring/net: fail zc send when unsupported by socket
      net: flag sockets supporting msghdr originated zerocopy
```

ollie-etl:
@Noah-Kennedy I don't think there is any blocking reason for this now

Noah-Kennedy (approved):

PR #127, opened 2022-10-07
driver: get rid of RC

Noah-Kennedy:
We already store the driver in a thread local, so there is no need to Rc it.

### DIFF
--- a/src/driver/mod.rs
+++ b/src/driver/mod.rs
@@ -42,13 +42,12 @@ use slab::Slab;
 use std::cell::RefCell;
 use std::io;
 use std::os::unix::io::{AsRawFd, RawFd};
-use std::rc::Rc;
 
 pub(crate) struct Driver {
     inner: Handle,
 }
 
-type Handle = Rc<RefCell<Inner>>;
+type Handle = RefCell<Inner>;
 
 pub(crate) struct Inner {
     /// In-flight operations
@@ -62,16 +61,16 @@ pub(crate) struct Inner {
 // type wraps the slab and ensures that, on drop, the slab is empty.
 struct Ops(Slab<op::Lifecycle>);
 
-scoped_thread_local!(pub(crate) static CURRENT: Rc<RefCell<Inner>>);
+scoped_thread_local!(pub(crate) static CURRENT: Handle);
 
 impl Driver {
     pub(crate) fn new(b: &crate::Builder) -> io::Result<Driver> {
         let uring = b.urb.build(b.entries)?;
 
-        let inner = Rc::new(RefCell::new(Inner {
+        let inner = RefCell::new(Inner {
             ops: Ops::new(),
             uring,
-        }));
+        });
 
         Ok(Driver { inner })
     }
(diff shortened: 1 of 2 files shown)


ollie-etl:
I really like this

FrankReh:
Interesting. The change only affected two files. And the only thing lost is the ability to track the inner driver on a per Op basis and there is no need to support multiple such drivers on the same thread?

Noah-Kennedy:
I need to fix some tests here still. But yes, that is all that is lost.

FrankReh:
Dumb question but you've been gracious with your explanations so I'll give it a go.

Why is a RefCell needed if the driver is owned by the current thread and the driver is not Send and not Sync? Can't anything that needs it just use Current.with?

Noah-Kennedy:
Closing this for now, as it needs larger changes to get around soundness issues. Will reopen once those are done.

ollie-etl:
> Dumb question but you've been gracious with your explanations so I'll give it a go.
> 
> Why is a RefCell needed if the driver is owned by the current thread and the driver is not Send and not Sync? Can't anything that needs it just use Current.with?

@FrankReh The access could be concurrent, i.e, by multiple tasks. If one holds a reference across an await point, then you have a problem. I don't think anything does though - indeed, the fact it runs without crashing is witness to this

ollie-etl:
@Noah-Kennedy : Could you elaborate on the soundness issues?

Noah-Kennedy:
@ollie-etl so there were two things. Easily resolved was the fact that the `Rc<RefCell<...>>` was being used to ensure that the `Op`s were `!Send`. Not as easily resolved was the fact that I was unsure to what degree the soundness of things was reliant on these ops keeping the driver around.

FrankReh:
@ollie-etl 
> If one holds a reference across an await point, then you have a problem.

First, I was very confused when I thought this comment came from Noah. Now that I've gotten all my questions to him about it cleaned up, my question to you is, did you mean that as a pro or a con?

Are you saying it is good we have this problem because we don't want multiple await points doing things to the driver? I'm wondering what is being done to the driver, potentially, that *should* lock it across an await point?

@Noah-Kennedy 
It would be very nice if the inner driver didn't have to be held across await points. In fact, holding it across await points with borrow_mut creates a problem when something else fires and wants to use it and I've run into it twice in the last couple weeks, once inside a tick call and once inside a close of a SharedFd. I think there could be fields inside Inner that are Cell but holding the entire driver with a lock across an await seems limiting and I haven't figured out why it should be limited like that on purpose?

For safety, perhaps it is enough that the inner driver wouldn't go away if there are slab entries outstanding and slab entries are the lingua franca between future and when it comes to tracking something in the io-ring interface that is still being run. If the driver is going out of scope, slab entries can be used to cancel outstanding operations, either with an older async operation, or with the new sync cancel api in kernel 6.0. Maybe a RefCell can wrap the slab with the rule that is it not to be borrowed across outside function calls.

Personally, I would like to support multiple io-ring drivers in the same thread, so new features like buf_ring don't have to assume just keying off the CURRENT value, but that's a slightly different topic. I just wanted to mention in your 'get rid of RC' issue that the one inner driver might not be the only io-uring driver being used in a current-thread runtime, although it is fair that the current-thread comes with a default.

PR #128, opened 2022-10-07
Support multi completion events

ollie-etl:
Support of multi completion events, originally proposed in https://github.com/tokio-rs/tokio-uring/pull/123, and submitted separately for review as suggested by @Noah-Kennedy 

Multi completion events contain a `CQE_F_MORE` flag, which indicates more events are expected associated with the same submission entry. This leads to a 1:M correspondance, where all ops currently supported are 1:1

This PR modifies `struct Op<T>` to support both types. It does this by adding the `update`  method to the Completable trait. The semantics of this are:
  
> Take 1 or more completions, _update_ the state of the operation, and return if the operation is `Finished`, with the result, or if `More` are required

For single event Ops, the result will always be `Finished`. As this is the most likely behaviour, the default method is provided. 

To support this with the current Lifecycle handling, the `Lifecycle::Completed` variant now contains 1 or more completions, rather than 1. This is because there is a data race between reading the completion queue, and operations calling `poll()` to consume `Lifecycle::Completed`. To manage this, completions are stored in a growable conatiner. `smallvec` (a vec with small string optimization) has been chosen to avoid allocations in the vast majority of cases. The default size is chosen to be small (under a cache line), by no rigorous method. It may want to be bigger (8 or 16).

Other changes to Lifecycle include the Ignored variant. This now has to wait for _all_ outstanding completions for an Op before being dropped. It just checks the more flag to do this.

There are design decisions here which want discussing. 

The major one in my eyes is the decision to have an unbounded collection of completion events. The other option would be to stall the completion queue handling once some predetermined threshold of completion events for a single op is reached. A waker would be registered, and completion queue handling would be woken when the Lifecycle::Completed was consumed by the Op's Poll(). I deemed this unpalatable, as it stalls all Ops

### DIFF
--- a/src/driver/op.rs
+++ b/src/driver/op.rs
@@ -24,10 +24,20 @@ pub(crate) struct Op<T: Completable + 'static> {
 
 pub(crate) trait Completable {
     type Output;
-    fn update(self, result: io::Result<u32>, flags: u32) -> Update<Self::Output, Self>
+    /// Called with one or more completion events.
+    ///
+    /// The implementation should update the state of the Op, and return
+    /// either the completed result, or the Op if more completions are required.
+    ///
+    /// The default implementation is suitable for all single completion event Op's.
+    fn update(
+        self,
+        mut completions: SmallVec<[(io::Result<u32>, u32); 4]>,
+    ) -> Update<Self::Output, Self>
     where
         Self: Sized,
     {
+        let (result, flags) = completions.remove(0);
         Update::Finished(self.complete(result, flags))
     }
     fn complete(self, result: io::Result<u32>, flags: u32) -> Self::Output;
@@ -148,19 +158,10 @@ where
                 Poll::Pending
             }
             Lifecycle::Ignored(..) => unreachable!(),
-            Lifecycle::Completed(v) => {
+            Lifecycle::Completed(events) => {
                 let data = me.data.take().unwrap();
-                let updated = v
-                    .into_iter()
-                    .fold(Update::More(data), |data, (result, flags)| {
-                        if let Update::More(data) = data {
-                            data.update(result, flags)
-                        } else {
-                            data
-                        }
-                    });
                 // Check if the operation requires more events to complete
-                match updated {
+                match data.update(events) {
                     Update::Finished(result) => {
                         inner.ops.remove(me.index);
                         me.index = usize::MAX;
@@ -191,19 +192,10 @@ impl<T: Completable> Drop for Op<T> {
             Lifecycle::Submitted | Lifecycle::Waiting(_) => {
                 *lifecycle = Lifecycle::Ignored(Box::new(self.data.take()));
             }
-            Lifecycle::Completed(v) => {
+            Lifecycle::Completed(events) => {
                 let data = self.data.take().unwrap();
-                let updated = v
-                    .into_iter()
-                    .fold(Update::More(data), |data, (result, flags)| {
-                        if let Update::More(data) = data {
-                            data.update(result, flags)
-                        } else {
-                            data
-                        }
-                    });
                 // Check if the operation requires more events to complete
-                match updated {
+                match data.update(events) {
                     Update::Finished(_) => {
                         inner.ops.remove(self.index);
                     }
--- a/src/driver/op.rs
+++ b/src/driver/op.rs
@@ -24,13 +24,16 @@ pub(crate) struct Op<T: Completable + 'static> {
 
 pub(crate) trait Completable {
     type Output;
-    fn update(self, result: io::Result<u32>, flags: u32) -> Update<Self::Output, Self> where Self: Sized {
+    fn update(self, result: io::Result<u32>, flags: u32) -> Update<Self::Output, Self>
+    where
+        Self: Sized,
+    {
         Update::Finished(self.complete(result, flags))
     }
     fn complete(self, result: io::Result<u32>, flags: u32) -> Self::Output;
 }
 
-pub(crate) enum Update<F,M>{
+pub(crate) enum Update<F, M> {
     /// The operation has finished
     Finished(F),
 
@@ -53,7 +56,7 @@ pub(crate) enum Lifecycle {
     ///
     /// This allocates if more than the default number of completions (currently 4)
     /// Are received before poll is invoked
-    Completed( SmallVec<[(io::Result<u32>, u32); 4]>),
+    Completed(SmallVec<[(io::Result<u32>, u32); 4]>),
 }
 
 impl<T> Op<T>
@@ -147,21 +150,22 @@ where
             Lifecycle::Ignored(..) => unreachable!(),
             Lifecycle::Completed(v) => {
                 let data = me.data.take().unwrap();
-                let updated = v.into_iter()
-                    .fold(Update::More(data), |data, (result, flags)|
+                let updated = v
+                    .into_iter()
+                    .fold(Update::More(data), |data, (result, flags)| {
                         if let Update::More(data) = data {
                             data.update(result, flags)
                         } else {
                             data
                         }
-                    );
+                    });
                 // Check if the operation requires more events to complete
                 match updated {
                     Update::Finished(result) => {
                         inner.ops.remove(me.index);
                         me.index = usize::MAX;
                         Poll::Ready(result)
-                    },
+                    }
                     Update::More(op) => {
                         let _ = me.data.insert(op);
                         *lifecycle = Lifecycle::Waiting(cx.waker().clone());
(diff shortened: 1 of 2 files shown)


ollie-etl:
@FrankReh FYI

src/driver/op.rs
```diff
@@ -23,9 +24,20 @@ pub(crate) struct Op<T: 'static> {
 
 pub(crate) trait Completable {
     type Output;
+    fn update(self, result: io::Result<u32>, flags: u32) -> Update<Self::Output, Self> where Self: Sized {
```
Noah-Kennedy, line 27: ...and this is exactly the sort of reason I added this trait lol

Noah-Kennedy (approved):
Looks good!

Noah-Kennedy:
@ollie-etl can you run `cargo fmt` on your PR?

oliverbunting:
> @ollie-etl can you run `cargo fmt` on your PR?

Sure, I'll take a look tomorrow

FrankReh:
I'm still confused why there is a fold on an iterator in the poll and the drop. Couldn't that work be done in the driver's complete function? Probably a good reason, but then could there be a one line comment above one of the iterator folds about the reason so the next person that is confused like me can save some time?

oliverbunting:
> I'm still confused why there is a fold on an iterator in the poll and the drop. Couldn't that work be done in the driver's complete function? Probably a good reason, but then could there be a one line comment above one of the iterator folds about the reason so the next person that is confused like me can save some time?

Updat could be changed to take an array of completions I guess, and the fold logic moved there if needed

src/driver/op.rs
```diff
@@ -37,8 +52,11 @@ pub(crate) enum Lifecycle {
     /// must be passed to the driver and held until the operation completes.
     Ignored(Box<dyn std::any::Any>),
 
-    /// The operation has completed.
-    Completed(io::Result<u32>, u32),
+    /// The operation has 1 or more unserviced completion events, and poll has yet to be called
+    ///
+    /// This allocates if more than the default number of completions (currently 4)
+    /// Are received before poll is invoked
```
FrankReh, line 58: maybe 'are' rather than 'Are'

src/driver/op.rs
```diff
@@ -130,30 +148,69 @@ where
...
-            Lifecycle::Completed(result, flags) => {
-                inner.ops.remove(me.index);
-                me.index = usize::MAX;
-
-                Poll::Ready(me.data.take().unwrap().complete(result, flags))
+            Lifecycle::Completed(v) => {
+                let data = me.data.take().unwrap();
+                let updated = v
+                    .into_iter()
+                    .fold(Update::More(data), |data, (result, flags)| {
```
FrankReh, line 155: maybe the variable name 'data' in More(data) is confusing me. Is the More(op) below the same type?

src/driver/op.rs
```diff
@@ -23,9 +24,23 @@ pub(crate) struct Op<T: 'static> {
...
+        Self: Sized,
+    {
+        Update::Finished(self.complete(result, flags))
+    }
     fn complete(self, result: io::Result<u32>, flags: u32) -> Self::Output;
 }
 
+pub(crate) enum Update<F, M> {
+    /// The operation has finished
+    Finished(F),
```
FrankReh, line 38: Could this F be named Output or Result?

src/driver/op.rs
```diff
@@ -23,9 +24,23 @@ pub(crate) struct Op<T: 'static> {
...
+
+    /// The operation expects more completion events
+    More(M),
```
FrankReh, line 41: Could this M be renamed Op? Still trying to get my head around why there isn't a variant added to the Lifecycle to handle the new variant of More, but maybe I'm just stuck thinking about things the way I worked them in a different fork for the streaming operations.

FrankReh (commented):
Just more questions. If you and Noah want to ignore them, that's fine. I'll probably be able to catch up to understanding these changes once they are in place.

src/driver/op.rs
ollie-etl: These are just maximally generic, and match Finished and More (F/M).

src/driver/op.rs
ollie-etl: Gone now

ollie-etl:
I moved the fold logic from poll / drop into update. The thing i don't like about that is that you have to consume smallvec by value, not as a &[(Result, flag)], as Result doesn't implement clone, and won't play with complete

FrankReh:
> I moved the fold logic from poll / drop into update. The thing i don't like about that is that you have to consume smallvec by value, not as a &[(Result, flag)], as Result<_,io::Error> doesn't implement clone, and won't play with complete. Overall i think the changes is for the better though. Its less code

Noah had indicated somewhere else, I think, that the SmallVec would be a temporary solution until he got a linked list using slab entries into place. Then the pass by value wouldn't be a big deal at all. But see my recent comments, I'm looking for an understanding more than a change. The Op and its future and the LifeCycle variants seem enough to get things done, perhaps with another variant for the More case. I don't understand the need for a level added to Complete. And again, since I added a whole level with the "wonderful" name Split to the slab entry itself, I pushed them to GitHub this afternoon, I know I'm on very shaky ground here.

ollie-etl:
> The Op and its future and the LifeCycle variants seem enough to get things done, perhaps with another variant for the More case. I don't understand the need for a level added to Complete. And again, since I added a whole level with the "wonderful" name Split to the slab entry itself, I pushed them to GitHub this afternoon, I know I'm on very shaky ground here.

`Lifecycle` fundamentally doesn't know what to do with a completion - it can know the more flag is set, but nothing of what the data represents, which is Op specific. It can't assume all completions marked MORE until NOTIF are associated with one completion. That is the case with zerocopy, and all other ops currently in the crate, but if I were to implement a streaming op like multishot-recv, I wouldn't need to change the Lifecycle or Op definition here.

update got added in preference to changing the definition of complete - maybe I should have just done that. It felt more intrusive. @Noah-Kennedy ?

FrankReh:
> `Lifecycle` fundamentally doesn't know what to do with a completion

I don't mean to be a pest. Sometimes my inclination for how things are named doesn't match others, so I'll apologize up front.

My take of the io_uring kernel interface is that we *can* rely on all intermediate cq entries having the MORE flag set, and the last not having the MORE flag set - that's what they mean by MORE. So I would say, as the tokio-uring driver starts to handle operations where the MORE flag is set, the Lifecycle *should* change to accommodate it. I would find it much easier to talk about the op's lifecycle being in a Waiting state, or a More collection state, or a Complete state than I would having to say its in a Complete state but just More, not Final. I guess we'd get used to saying it is in a Complete/More state or a Complete/FInal state, but why the extra level of enum when the Op could handle it just as easily as another variant on Lifecycle, and why the extra level of nesting it requires in the code?

oliverbunting:
@FrankReh I see where you're coming from. That specifically assumes an op is not going to issue a result until NOTIF is received. But what if I want to implement an op which is a stream? I may want to produce results for intermediates. And lifecycle can't know this. All it can do pass the result to the op for it to decide

FrankReh:
> @FrankReh I see where you're coming from. That specifically assumes an op is not going to issue a result until NOTIF is received. But what if I want to implement an op which is a stream? I may want to produce results for intermediates. And lifecycle can't know this. All it can do pass the result to the op for it to decide

I had a different approach in mind (and implemented as a POC) for streams. Rather than use create a Future, a Stream is created. The Lifecycle variants for that were slightly different.

FrankReh:
Something tangential. I just wrote issue #129 and the solution may well be a new trait method, like *cleanup*, that takes the result and flags, so something like a fd could be dropped properly. But if not a trait method, a callback or stable entry to do the same.

ollie-etl:
@FrankReh I had a look at that branch of tours, but ciould you mark it up as a PR? There is a lot going on in there.

I hope to get around to multishot recv soon also. Time is always the problem.

FrankReh:
> I hope to get around to multishot recv soon also. Time is always the problem.

I hope we get the pbuf api figured out before a multishot recv is introduced, or maybe the two at the same time.

On my own proof of concept, I'm hoping to get feedback on my cancellable Index idea first. How cancellable can be achieved both explicitly, and implicitly when the Stream is dropped, will set how other features can be folded on top. Noah had started to introduce another tack that would make cancellable possible.

Noah-Kennedy:
@FrankReh I'd agree with you on the provided buffers here. Those really do need to be done first.

ollie-etl:
I think this has now been superseded by the implementation in #130. Closing on that basis.

PR #130, opened 2022-10-12
Support multi completion events: v2

ollie-etl:
@Noah-Kennedy @FrankReh . I took on board your comments with the other PR and wanted to offer an alternative implementation to #128. This addresses some of the shortcomings of #128.

- Multi CQE's are now held on a driver-scoped slab of completions. This goes some way to address the biggest shortcoming of the other approach, which was its propensity to allocate if multiple completion events were received for a single Op in quick succession. Obviously, allocation can still occur, but the total storage space is scoped to the runtime rather than per Op - so making it big initially benefits all Ops.

- Op is now parameterized over classes of Ops, discriminated with a marker type. This allows us to easily add classes of Operations. All current ones are represented by `SingleCQE`. ~I shall (haven't yet) rebase #123 to show its use.~ now merged onto this. Will update that with performance figures for my use case.

- @FrankReh I took on board your comments around Completed + enum variants vs adding another variant. 
   * Lifecycle::Completed is now guaranteed to not contain the MORE flag. 
   
- Lifecycle::CompletionList will never be observed for SingleCQE types.


src/driver/op.rs
```diff
@@ -37,28 +48,33 @@ pub(crate) enum Lifecycle {
     /// must be passed to the driver and held until the operation completes.
     Ignored(Box<dyn std::any::Any>),
 
-    /// The operation has completed.
+    /// The operation has completed with a single result
```
ollie-etl, line 51: The single Op case is unchanged. We don't pay any slab lookup overhead for single completions

src/driver/op.rs
```diff
@@ -103,7 +119,7 @@ where
     }
 }
 
-impl<T> Future for Op<T>
+impl<T> Future for Op<T, SingleCQE>
```
ollie-etl, line 122: Constrained this implementation to SingleCQE types, which includes all Ops in the current implementation

src/driver/op.rs
```diff
@@ -133,49 +153,82 @@ where
             Lifecycle::Completed(result, flags) => {
                 inner.ops.remove(me.index);
                 me.index = usize::MAX;
-
                 Poll::Ready(me.data.take().unwrap().complete(result, flags))
             }
+            Lifecycle::CompletionList(..) => {
```
ollie-etl, line 158: This can be safely ignore. By implmenting a SingleCQE class of operation, we're promising we'll never receive a CQE with more set, so this can never be constructed

FrankReh:
I like the linked list in a slab implementation. I'll note its a doubly linked list and I didn't see the benefit of the backward links but maybe that detail future proofs the list somehow.

Is there a branch that shows the CompletionList or the CompletionEntries being passed or consumed by a zero copy operation?

I had other comments but the Completion slab is the big addition and I like it. I wonder if @Noah-Kennedy  could chime in on it?

I could try to change my stream fork to use the same mechanism, getting rid of the Vec of results.

oliverbunting:
> I like the linked list in a slab implementation. I'll note its a doubly linked list and I didn't see the benefit of the backward links but maybe that detail future proofs the list somehow.

So both pop and push are 1 lookup. Admittedly, holding the end index is sufficient for this, so the list itself doesn't need to be doubly linked

oliverbunting:
> Is there a branch that shows the CompletionList or the CompletionEntries being passed or consumed by a zero copy operation?
> 
Yes, I've rebased the send_zc branch on to this

src/driver/op/completion.rs
```diff
@@ -0,0 +1,117 @@
+use slab::Slab;
```
Noah-Kennedy, line 1: I'd recommend adding some unit tests to this file.

src/driver/op/completion.rs
```diff
@@ -0,0 +1,117 @@
...
+use std::{
+    io,
```
Noah-Kennedy, line 3: Let's also add some top-level docs explaining what this is all for.

Noah-Kennedy (changes requested):
Sorry about the delay! I really like this approach. It needs a few finishing touches before it's good to go.

src/driver/op/completion.rs
```diff
@@ -0,0 +1,117 @@
...
+    start: usize,
+    end: usize,
+}
+
+/// Multi cycle operations may return an unbounded number of CQE's
+/// for a single cycle SQE.
+///
+/// These are held in an indexed linked list
+pub(crate) struct Completion {
+    val: (io::Result<u32>, u32),
```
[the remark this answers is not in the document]
ollie-etl: Done, on all use sitea, including Lifecycle.

ollie-etl:
@Noah-Kennedy - I think I've addressed your concerns - let me know if you've more.

src/driver/mod.rs
```diff
@@ -155,42 +160,40 @@ impl Drop for Driver {
...
     }
 }
 
 impl Drop for Ops {
     fn drop(&mut self) {
-        assert!(self.0.is_empty());
-    }
-}
-
-fn resultify(cqe: &cqueue::Entry) -> io::Result<u32> {
```
ollie-etl, line 188: This has been replaced with the `From<cqueue::Entry> for CqeResult` instance

ollie-etl:
@FrankReh thanks for the fairly rapid turnaround on #149. It makes this much cleaner

FrankReh:
@ollie-etl
> @FrankReh thanks for the fairly rapid turnaround on #149. It makes this much cleaner

Sorry I had to step out for a few hours. I'm on US East Coast time.

ollie-etl:
@FrankReh Not at all, I'd noticed we were in different time-zones. I'm also I'm genuinely grateful for the level of scrutiny these Pr's are receiving - these things are generally better got right or close to it first time around.

src/driver/op.rs
```diff
@@ -147,46 +174,87 @@ where
...
+                if io_uring::cqueue::more(cqe.flags) {
+                    let mut list = SlabListIndices::new().into_list(completions);
+                    list.push(cqe);
+                    *self = Lifecycle::CompletionList(list.into_indices());
+                } else {
+                    *self = Lifecycle::Completed(cqe);
+                }
+                if let Lifecycle::Waiting(waker) = x {
+                    waker.wake();
+                }
```
ollie-etl, line 235: There are 2 design decisions here, 

- completions with more flag set go straight onto the slab. It would be equally valid to put them into the Completed variant, and only upgrade to the slab if another is received before Poll.

- wake is called whenever we have more work. It would also be possible to defer calling wake until a completion without more is received. This behavior could be configured per Op type.

ollie-etl: Those are both optimisations for things we don't have yet, so I think definitely another PR

FrankReh (changes requested):
This is very nice. I think Noah has already said that too. I have a few small issues around comments.

src/driver/op/slab_list.rs
```diff
@@ -0,0 +1,170 @@
...
+/// A linked list backed by slab storage
+pub(crate) struct SlabList<'a, T> {
+    index: SlabListIndices,
+    slab: &'a mut Slab<SlabListEntry<T>>,
+}
+
+// Indices to the head and tail of a single list held within a SlabList
+pub(crate) struct SlabListIndices {
+    start: usize,
+    end: usize,
```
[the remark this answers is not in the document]
ollie-etl: yes. or even 2. u16 would be fine i suspect.

ollie-etl:
> Too bad the length of the slab linked list is so hard to track

It'd be very easy to add in `SlabListIndices`

src/driver/op/slab_list.rs
```diff
@@ -0,0 +1,170 @@
...
+    pub(crate) fn peek_end(&mut self) -> Option<&T> {
+        if self.index.end == usize::MAX {
+            None
+        } else {
+            Some(&self.slab[self.index.end].entry)
+        }
+    }
+
+    /// Pop from front of list
+    #[allow(dead_code)]
```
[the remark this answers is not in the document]
ollie-etl: in this pr, yes we've no consumer yet. I add one in send-zc

src/driver/op/slab_list.rs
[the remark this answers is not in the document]
ollie-etl: we might even get away with 1. I don't know all the usecases.

src/driver/op.rs
```diff
@@ -39,12 +58,15 @@ pub(crate) enum Lifecycle {
...
+
+    /// One or more completion results have been recieved
+    /// This holds the indices uniquely identifying the list within the slab
+    CompletionList(SlabListIndices),
 }
 
 /// A single CQE entry
 pub(crate) struct CqeResult {
     pub(crate) result: io::Result<u32>,
-    #[allow(dead_code)]
```
ollie-etl, line 47: now used

src/driver/op.rs
```diff
@@ -147,46 +174,92 @@ where
...
+            x @ Lifecycle::Submitted | x @ Lifecycle::Waiting(..) => {
+                if io_uring::cqueue::more(cqe.flags) {
+                    let mut list = SlabListIndices::new().into_list(completions);
+                    list.push(cqe);
+                    *self = Lifecycle::CompletionList(list.into_indices());
+                } else {
+                    *self = Lifecycle::Completed(cqe);
+                }
+                if let Lifecycle::Waiting(waker) = x {
+                    // wake is called whenever we have more work.
```
FrankReh, line 234: 'more work' is confusing with how 'more' is used to represent a More cqe in some places. Could this read something like "wake to pick up this latest cqe".

src/driver/op.rs
```diff
@@ -147,46 +174,92 @@ where
...
+                    // Also possible would be to defer calling until cqe with !more flag set
+                    waker.wake();
+                }
                 false
             }
-            Lifecycle::Waiting(waker) => {
-                *self = Lifecycle::Completed(cqe);
-                waker.wake();
+            lifecycle @ Lifecycle::Ignored(..) => {
+                // We must check if any more CQEs are expected before dropping
```
FrankReh, line 241: This comment could add that we keep the Op alive but there is no need to push the cqe onto a list because the Op has been dropped already.

src/driver/op.rs
```diff
@@ -147,46 +174,92 @@ where
...
-        match lifecycle {
+        match mem::replace(lifecycle, Lifecycle::Submitted) {
             Lifecycle::Submitted | Lifecycle::Waiting(_) => {
                 *lifecycle = Lifecycle::Ignored(Box::new(self.data.take()));
             }
             Lifecycle::Completed(..) => {
                 inner.ops.remove(self.index);
             }
+            Lifecycle::CompletionList(indices) => {
+                // Deallocate list entries, recording if the more CQE's are expected
```
FrankReh, line 202: // Deallocate list entries, recording if more CQE's are expected

src/driver/op.rs
```diff
@@ -147,46 +174,92 @@ where
...
     fn drop(&mut self) {
+        use std::mem;
+
         let mut inner = self.driver.borrow_mut();
-        let lifecycle = match inner.ops.get_mut(self.index) {
-            Some(lifecycle) => lifecycle,
+        let (lifecycle, completions) = match inner.ops.get_mut(self.index) {
+            Some(val) => val,
             None => return,
         };
```
FrankReh, line 193: You could do us all a favor by adding here something to the effect:

// From this point, 1) the lifecycle will be made Ignored because we don't care about further results but more results are still expected, 2) the Op's entry will be removed from the slab and there won't be any further lifecycle, or 3) panic due to an unexpected state being reached.

src/driver/op.rs
```diff
@@ -147,46 +174,92 @@ where
...
+            }
+
+            Lifecycle::Completed(..) => {
+                // Completions with more flag set go straight onto the slab,
+                // and are handled in Lifecycle::CompletionList.
+                // To construct Lifecycle::Completed, a CQE without MORE was received
+                // we shouldn't be receiving another.
+                unreachable!("invalid operation state")
+            }
+            Lifecycle::CompletionList(indices) => {
```
FrankReh, line 257: This case also causes non-More cqe to be pushed onto the list. We only expect one, but a second one doesn't get noticed here and doesn't cause an unreachable panic like the case above does. That difference bothers me enough that I think some comment is warranted here, about both kinds of cqe being pushed at least.

FrankReh (changes requested):
Hope you are fresh when you read these. I think they are all comment related.

ollie-etl:
@FrankReh All comments addressed

src/driver/op.rs
```diff
@@ -175,7 +175,7 @@ where
                 Poll::Ready(me.data.take().unwrap().complete(cqe))
             }
             Lifecycle::CompletionList(..) => {
-                unreachable!()
+                unreachable!("No more flag set for SinglCQE")
```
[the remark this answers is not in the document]
FrankReh: appears to be missing the e in SingleCQE

src/driver/op.rs
FrankReh: I'm sorry to be picky, could you reword this without the word 'more'?

src/driver/op.rs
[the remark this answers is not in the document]
ollie-etl: please refresh your browser

src/driver/op.rs
```diff
@@ -149,31 +172,57 @@ where
...
         CONTEXT.with(|runtime_context| {
             runtime_context.with_driver_mut(|driver| {
-                let lifecycle = match driver.ops.get_mut(self.index) {
-                    Some(lifecycle) => lifecycle,
-                    None => return,
+                // Get the Op Lifecycle state from the driver
+                let (lifecycle, completions) = match driver.ops.get_mut(self.index) {
+                    Some(val) => val,
+                    None => {
+                        // Op dropped after the driver
```
FrankReh, line 200: I like the addition of this comment. It highlights a potential new problem for me that I don't propose we address with this PR. Namely, we should allow a silent drop here if the driver truly has dropped before the Op, but I think we should panic if the driver exists but the slab entry for this Op's self.index is returning None, because that shows a serious invariant is being violated - if this can happen, then a cqe could come in for a slab index that had already been given to a new Op.

FrankReh (approved):

src/driver/op/slab_list.rs
```diff
@@ -119,3 +120,53 @@ impl<'a, T> Drop for SlabList<'a, T> {
...
+        {
+            let mut list =  SlabListIndices::new().into_list(&mut slab);
+            list.push(42);
+            assert!(!list.is_empty());
+            // This forgets the entries
+            let _ = list.into_indices();
+        }
+        assert!(!slab.is_empty());
+    }
+}
```
FrankReh: Is this missing newline at end of file still a problem? I was trying to see why GitHub didn't think all of Noah's reviews were satisfied.

FrankReh:
I'm going to have this page dismiss Noah's review. The page seems broken somehow. When his review is expanded, it shows everything having been resolved. Maybe it just means he was expected to review and accept.

I think we can risk it. If more changes are needed, we'll make them with a new PR.

PR #131, opened 2022-10-12
fs: implement fs::File::from_std

littledivy:
Implement `fs::File::from_std`. Similar to `tokio::fs::File::from_std` to convert 
an already opened fd to tokio-uring File.

### DIFF
--- a/src/fs/file.rs
+++ b/src/fs/file.rs
@@ -4,7 +4,7 @@ use crate::fs::OpenOptions;
 
 use std::fmt;
 use std::io;
-use std::os::unix::io::{AsRawFd, FromRawFd, RawFd};
+use std::os::unix::io::{AsRawFd, FromRawFd, IntoRawFd, RawFd};
 use std::path::Path;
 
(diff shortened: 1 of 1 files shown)


ollie-etl:
Have you tested this? I've tried something similar, and i get `-EINVAL`. I wonder if the file needs to be opened with uring?

Noah-Kennedy (approved):

src/fs/file.rs
```diff
@@ -120,6 +120,14 @@ impl File {
         File { fd }
     }
 
+    /// Converts a [`std::fs::File`][std] to a [`tokio_uring::fs::File`][file].
+    ///
+    /// [std]: std::fs::File
+    /// [file]: File
+    pub fn from_std(file: std::fs::File) -> File {
+        File::from_shared_fd(SharedFd::new(file.as_raw_fd()))
```
quininer, line 128: I think it should be `into_raw_fd`?

Noah-Kennedy: Ah missed that, you're right.

PR #134, opened 2022-10-13
Add NoOp support

ollie-etl:
Adds support for the NoOp (no-operation) operation. This does nothing except await a completion.

I'm trying to understand the performance aspects of various bits of my system, and understanding the runtime overhead is useful. This makes that a lot easier.

It doesn't clearly fall into `fs`, `net`, or `buf`, so I've placed it at the top level.

### DIFF
--- a/src/driver/mod.rs
+++ b/src/driver/mod.rs
@@ -7,6 +7,9 @@ mod connect;
 
 mod fsync;
 
+mod noop;
+pub(crate) use noop::NoOp;
+
 mod op;
 pub(crate) use op::Op;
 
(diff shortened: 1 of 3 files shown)


Noah-Kennedy:
@ollie-etl can you add a unit test for this?

ollie-etl:
Sure

Noah-Kennedy:
@ollie-etl the test fails

ollie-etl:
> @ollie-etl the test fails
Arg! I refactored before committing, obviously. Sorry about that. I'll fix it up

Noah-Kennedy (approved):

PR #136, opened 2022-10-15
Add writev to TcpStream

fasterthanlime:
Seemed like a glaring omission. Not sure how this fits in with in-progress work, let's see if this alone works.

### DIFF
--- a/src/driver/socket.rs
+++ b/src/driver/socket.rs
@@ -46,7 +46,7 @@ impl Socket {
 
     pub async fn writev<T: IoBuf>(&self, buf: Vec<T>) -> crate::BufResult<usize, Vec<T>> {
         let op = Op::writev_at(&self.fd, buf, 0).unwrap();
-        op.writev().await
+        op.await
     }
 
     pub(crate) async fn send_to<T: IoBuf>(
--- a/src/driver/socket.rs
+++ b/src/driver/socket.rs
@@ -44,6 +44,11 @@ impl Socket {
         op.write().await
     }
 
+    pub async fn writev<T: IoBuf>(&self, buf: Vec<T>) -> crate::BufResult<usize, Vec<T>> {
+        let op = Op::writev_at(&self.fd, buf, 0).unwrap();
+        op.writev().await
+    }
+
     pub(crate) async fn send_to<T: IoBuf>(
         &self,
         buf: T,
--- a/src/net/tcp/stream.rs
+++ b/src/net/tcp/stream.rs
@@ -149,6 +149,34 @@ impl TcpStream {
         (Ok(()), buf)
     }
 
+    /// Write data from buffers into this socket returning how many bytes were
+    /// written.
+    ///
+    /// This function will attempt to write the entire contents of `bufs`, but
+    /// the entire write may not succeed, or the write may also generate an
+    /// error. The bytes will be written starting at the specified offset.
+    ///
+    /// # Return
+    ///
+    /// The method returns the operation result and the same array of buffers
+    /// passed in as an argument. A return value of `0` typically means that the
+    /// underlying socket is no longer able to accept bytes and will likely not
+    /// be able to in the future as well, or that the buffer provided is empty.
+    ///
+    /// # Errors
+    ///
+    /// Each call to `write` may generate an I/O error indicating that the
+    /// operation could not be completed. If an error is returned then no bytes
+    /// in the buffer were written to this writer.
+    ///
+    /// It is **not** considered an error if the entire buffer could not be
+    /// written to this writer.
+    ///
+    /// [`Ok(n)`]: Ok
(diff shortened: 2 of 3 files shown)


src/driver/socket.rs
```diff
@@ -44,6 +44,11 @@ impl Socket {
         op.write().await
     }
 
+    pub async fn writev<T: IoBuf>(&self, buf: Vec<T>) -> crate::BufResult<usize, Vec<T>> {
```
ollie-etl, line 47: Could this be generalised to 

```suggestion
    pub async fn writev<T: IoBuf, B: Deref<Target = [T]>>(&self, buf: Buf) -> crate::BufResult<usize, Buf> {
```

fasterthanlime: I'm not really fond of the `Vec<T>` requirement either here or internally:

https://github.com/tokio-rs/tokio-uring/blob/913aa14c567cfae026b7d4f88a5155921cfb9575/src/driver/writev.rs#L22

And I am using `smallvec` in a bunch of places but I'm always wary of stuff moving after it's been sent to the uring submission queue. I guess this small change is probably harmless though? I wonder what @Noah-Kennedy thinks.

fasterthanlime: Ah, we'd also need to make the Future type generic over the buffer type.

Noah-Kennedy:
@fasterthanlime this would be nice to have in the crate.

fasterthanlime:
Note that in `tokio`, the equivalent method is named `write_vectored`: https://docs.rs/tokio/latest/tokio/io/trait.AsyncWriteExt.html#method.write_vectored

I also haven't written tests for it yet.

Noah-Kennedy:
I'm down to merge once tests are in. I'm fine with the Vec dependency here.

src/driver/socket.rs
FrankReh: Making the op futures generic on the buffer type seems to be the direction in general. The logic around op futures was simplified with the Completable trait recently and the Completable trait already is generic on the buffer type. .. If I understand your discussion here correctly ..

FrankReh: It seems fine to be following what was done for the File.writev_at at this time.

src/net/tcp/stream.rs
```diff
@@ -149,6 +149,34 @@ impl TcpStream {
...
+    /// The method returns the operation result and the same array of buffers
+    /// passed in as an argument. A return value of `0` typically means that the
+    /// underlying socket is no longer able to accept bytes and will likely not
+    /// be able to in the future as well, or that the buffer provided is empty.
+    ///
+    /// # Errors
+    ///
+    /// Each call to `write` may generate an I/O error indicating that the
+    /// operation could not be completed. If an error is returned then no bytes
+    /// in the buffer were written to this writer.
```
FrankReh, line 170: This comment about no bytes in the buffer seems suspect. The writev is of a list of buffers, the comment makes it sound like a single buffer. I don't see a kernel man page for writev making this kind of statement. But this comment was already made for the File writev_at function so no need to change it here. Just sayin, it looks odd.

FrankReh (changes requested):
These changes look good. Would you mind adding the same function for the UnixStream?

And if you wouldn't mind merging master in again. There's at least one line that should be 'op.await' now.

Barring anyone else's concerns, this can then be merged.

src/driver/socket.rs
```diff
@@ -44,6 +44,11 @@ impl Socket {
         op.await
     }
 
+    pub async fn writev<T: IoBuf>(&self, buf: Vec<T>) -> crate::BufResult<usize, Vec<T>> {
+        let op = Op::writev_at(&self.fd, buf, 0).unwrap();
+        op.writev().await
```
FrankReh, line 49: Change this to op.await and it will compile again having merged with master.

Noah-Kennedy:
CI is currently failing.

FrankReh:
> CI is currently failing.

Yea. That's because last night I hit the 'auto update' button and it updated the files it could but of course not the new file that was added. And I don't think I can go into the file from here and make the trivial fix. If we don't here from the OP soon, I can fork their branch and make this PR myself.

fasterthanlime:
> If we don't here from the OP soon, I can fork their branch and make this PR myself.

Go ahead and make your own PR! I'm not able to fix mine right now, and there's no point in blocking on me 😌

Noah-Kennedy (approved):

FrankReh (approved):

fasterthanlime:
@FrankReh thanks for getting this past the finish line!

FrankReh:
My pleasure. Still getting used to squash merges on GitHub and being able to push to someone else's branch for their PR but I found a how-to that made it easy. What with master changing so much now, this is going to come up often and generally the maintainers will know how to resolve differences much easier than the OP, even more so if the PR was written a while ago. Good luck with your http investigations!

PR #140, opened 2022-10-21
demo buffer group

FrankReh:
Warning. Don't actually merge this.

It has a hack in the Cargo.toml to use my io-uring crate because the standard version of the create isn't quite ready.

This shows how the generic BufGroup can be used for operations instead of a buffer.

So far only the read_bg function is defined. Not sure about the name. And it could be that the uring operations end up being created with a Builder and the final build step could actually be one of two forms of build, one that takes a buffer and one that takes a BufGroup.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -27,6 +27,14 @@ jobs:
     steps:
       - run: exit 0
 
+  bench:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+      - name: Install Rust
+        run: rustup update stable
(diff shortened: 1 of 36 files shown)


FrankReh:
I'm going to get back to this early next week. Having seen the work in the PR for fixed reads and writes, I may have a better way to introduce the concept of buffer groups. And I should have my unit tests for io-uring ready by then too so this can be built and tested against the normal io-uring crate.

So no reviews requested.

Noah-Kennedy:
@FrankReh might be a better idea to use an issue thread to bikeshed the API first before we get to the implementation.

PR #142, opened 2022-10-23
rt: refactor runtime to avoid Rc<RefCell<...>>

Noah-Kennedy:
This change refactors the runtime to not use reference counting directly in the ops themselves. Instead, ops access the driver via thread local variables.

This is sound because dropping the driver (which happens when it is removed from its thread-local state) blocks the thread until all ops complete, ensuring that we do not free the contents of the driver until after all operations have completed.

### DIFF
--- a/src/runtime/mod.rs
+++ b/src/runtime/mod.rs
@@ -62,7 +62,7 @@ impl Runtime {
         let rt = tokio::runtime::Builder::new_current_thread()
             .on_thread_park(|| {
                 CONTEXT.with(|x| {
-                    let _ = x.with_driver(|d| d.uring.submit());
+                    let _ = x.with_driver_mut(|d| d.uring.submit());
                 });
             })
             .enable_all()
@@ -96,7 +96,7 @@ impl Runtime {
                 loop {
                     // Wait for read-readiness
                     let mut guard = driver.readable().await.unwrap();
-                    CONTEXT.with(|cx| cx.with_driver(|driver| driver.tick()));
+                    CONTEXT.with(|cx| cx.with_driver_mut(|driver| driver.tick()));
                     guard.clear_ready();
                 }
             }
--- a/src/runtime/mod.rs
+++ b/src/runtime/mod.rs
@@ -1,15 +1,23 @@
-use crate::driver::{Driver, CURRENT};
-use std::cell::RefCell;
+use crate::driver::Driver;
 
 use std::future::Future;
 use std::io;
+use std::os::unix::io::{AsRawFd, RawFd};
 use tokio::io::unix::AsyncFd;
 use tokio::task::LocalSet;
+
+mod context;
+
+pub(crate) use context::RuntimeContext;
+
+thread_local! {
+    pub(crate) static CONTEXT: RuntimeContext = RuntimeContext::new();
+}
 
 /// The Runtime executor
 pub struct Runtime {
     /// io-uring driver
-    driver: AsyncFd<Driver>,
+    uring_fd: RawFd,
 
     /// LocalSet for !Send tasks
     local: LocalSet,
@@ -44,7 +52,7 @@ pub struct Runtime {
 ///     handle.await.unwrap();
 /// });
 /// ```
-pub fn spawn<T: std::future::Future + 'static>(task: T) -> tokio::task::JoinHandle<T::Output> {
+pub fn spawn<T: Future + 'static>(task: T) -> tokio::task::JoinHandle<T::Output> {
     tokio::task::spawn_local(task)
 }
 
@@ -53,8 +61,8 @@ impl Runtime {
     pub fn new(b: &crate::Builder) -> io::Result<Runtime> {
         let rt = tokio::runtime::Builder::new_current_thread()
             .on_thread_park(|| {
-                CURRENT.with(|x| {
-                    let _ = RefCell::borrow_mut(x).uring.submit();
+                CONTEXT.with(|x| {
+                    let _ = x.with_driver(|d| d.uring.submit());
                 });
             })
             .enable_all()
@@ -62,12 +70,17 @@ impl Runtime {
 
         let local = LocalSet::new();
 
-        let driver = {
-            let _guard = rt.enter();
-            AsyncFd::new(Driver::new(b)?)?
-        };
+        let driver = Driver::new(b)?;
 
-        Ok(Runtime { driver, local, rt })
+        let driver_fd = driver.as_raw_fd();
+
+        CONTEXT.with(|cx| cx.set_driver(driver));
+
+        Ok(Runtime {
+            uring_fd: driver_fd,
+            local,
+            rt,
+        })
     }
 
     /// Runs a future to completion on the current runtime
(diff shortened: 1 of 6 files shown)


Noah-Kennedy:
Huh, that doesn't happen on my machine.

Noah-Kennedy:
Hmm, actually it does, its just dependent on the order the tests run in (maybe?)

Noah-Kennedy:
Nevermind, CLion has a bug and apparantly hides test failures with double panics.

Noah-Kennedy:
Well that worked.

FrankReh:
Nice. You've removed the Rc from the highest level because the driver reference isn't being kept per running Op any longer. The running Op can access the TLS variable to get to the slab and the uring instance.

This doesn't completely rule out ever being able to support multiple urings in the same current thread runtime, but it would require some further refactoring, as it would have anyway. It's good to differentiate the runtime in this crate from the uring instance, it would seem to allow further flexibility later on.

Noah-Kennedy:
I'm going to be opening several additional refactoring PRs. I'd like to combine a lot of our TCP and Unix logic similar to what tokio does internally with PollEvented.

ollie-etl (commented):
I like this

ollie-etl:
Hmm, I tried merging #144 with this, and it seems this is a fairly significant performance regression

ollie-etl:
More notes on performance. Adding LTO flags to the release build `git cherry-pick 7d3d8df`, vs #144 with the same.

the fairly dreadful no_op/1 benchmark regresses significantly vs #144, but the 256 way concurrency is actually better.

```console
no_op/1                 time:   [182.93 ms 183.70 ms 184.50 ms]
                        thrpt:  [541.99 Kelem/s 544.37 Kelem/s 546.67 Kelem/s]
                 change:
                        time:   [+40.007% +40.654% +41.302%] (p = 0.00 < 0.05)
                        thrpt:  [-29.229% -28.904% -28.575%]
                        Performance has regressed.

no_op/256               time:   [19.945 ms 20.011 ms 20.087 ms]
                        thrpt:  [4.9784 Melem/s 4.9973 Melem/s 5.0139 Melem/s]
                 change:
                        time:   [-3.8566% -3.4359% -2.9896%] (p = 0.00 < 0.05)
                        thrpt:  [+3.0817% +3.5582% +4.0114%]
                        Performance has improved.
Found 10 outliers among 100 measurements (10.00%)
  6 (6.00%) high mild
  4 (4.00%) high severe

no_op_x1
  Instructions:           566777937 (+53.60089%)
  L1 Accesses:            828856458 (+51.37721%)
  L2 Accesses:                  312 (-99.96116%)
  RAM Accesses:                1202 (+8.679928%)
  Estimated Cycles:       828900088 (+50.27212%)

no_op_x256
  Instructions:            89362937 (-1.095379%)
  L1 Accesses:            140236031 (-2.657184%)
  L2 Accesses:               732513 (+14.23632%)
  RAM Accesses:                1760 (+5.960265%)
  Estimated Cycles:       143960196 (-2.286151%)
```

Noah-Kennedy:
This is likely noise. We might want to try using a longer benchmark period for this.

ollie-etl:
@Noah-Kennedy the benchmarks are performing 100k `Noop`, and iterating 100-300 times. Each criterion benchmark is therefore 10M+ operations. I can reproduce very consistently.

Noah-Kennedy:
Ah, interesting.

Noah-Kennedy:
I'll profile this tonight then to see what is happening.

ollie-etl:
@Noah-Kennedy : You may find `simple-benchmark-2` from git@github.com:etlsystems/tokio-uring` helpful. It extends the `simple-benchmark` I've submitted as a PR with a few things:

   - much cleaner benchmark code, enabled by exposing the runtime
   - [pprof](https://github.com/tikv/pprof-rs) flamegraph support.
   - Adds lto flags and targets modern x86s, for representative benchmarks
   
   Given your profile indicates you work for Cloudflare, I suspect you're fairly clued up on profiling anyway. My experience is if I haven't got something terribly wrong, a criterion result of more than 2% is real. Indeed, trying to find many 2% improvements is basically were I spend my days.
   
I'm having a look at the benchmarks on master currently

ollie-etl:
And looking at the flamegraphs, you can just disregard any low number of concurrency. thread_park totally dominates

FrankReh:
> the fairly dreadful no_op/1 benchmark regresses significantly vs #144, but the 256 way concurrency is actually better.

How to interpret the no_op/1 name and the no_op/256 names? Is the /<number> indicating the number of concurrent tasks spawned? What does it mean that there are per second numbers? Is a spawned task replaced by another when the first one finishes? Or are there a certain number run per second?

I'm wondering if no_op/1 means just one operation is performed at a time? So there is little stress on the system and the test has to wait for the round trip from userland to kernel and back to userland for a single test to be counted before the next is started.

And how many cores is this running on? Sorry if that is mentioned somewhere already. And is this a native machine or a vm on a hypervisor?

Noah-Kennedy:
Do we want to gate this PR on improving the no-op perf? I'd vote for not doing so, but I'd like to get your thoughts @ollie-etl @FrankReh.

Noah-Kennedy:
@FrankReh I think that we should probably merge this asap since it touches things we want to change in other PRs. Merging it post-review sounds like a good idea here.

I'll resolve the conflicts.

Noah-Kennedy:
@FrankReh merge conflicts are resolved

oliverbunting:
From my perspective, please merge

Noah-Kennedy:
Cool, as soon as @FrankReh signs off, I'll merge.

src/runtime/mod.rs
```diff
@@ -53,46 +61,61 @@ impl Runtime {
...
-            tokio::pin!(future);
+        tokio::pin!(future);
 
-            self.rt
-                .block_on(self.local.run_until(crate::future::poll_fn(|cx| {
-                    assert!(drive.as_mut().poll(cx).is_pending());
-                    future.as_mut().poll(cx)
-                })))
-        })
+        self.local.spawn_local(drive);
```
FrankReh, line 107: Surprise! This is the biggest change in this PR in my book. I always had a hard time wrapping my head around the earlier block_on workings, what with its assert that gets the drive poll run once before the actual future is polled. And here it is made independent and simply a local task, that I guess will be polled by tokio any time the syncfd becomes ready. Very interesting. Perhaps a significant change. Perhaps very nice.

src/util.rs
```diff
@@ -0,0 +1,4 @@
+use std::marker::PhantomData;
+
+/// Utility ZST for ensuring that opcodes are `!Send` and `!Sync`.
```
FrankReh, line 3: What does ZST mean?

FrankReh (changes requested):
Cool. Small questions.

FrankReh:
Be back in an hour.

src/runtime/mod.rs
```diff
@@ -1,15 +1,23 @@
...
+
+thread_local! {
+    pub(crate) static CONTEXT: RuntimeContext = RuntimeContext::new();
+}
+
 /// The Runtime executor
 pub struct Runtime {
     /// io-uring driver
-    driver: AsyncFd<Driver>,
+    uring_fd: RawFd,
```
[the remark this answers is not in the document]
Noah-Kennedy: It's setup like this now to allow us to build the Runtime struct outside of a runtime context. An `AsyncFd` cannot be constructed outside of a tokio runtime context.

src/util.rs
Noah-Kennedy: Zero-Sized Type

src/runtime/mod.rs
Noah-Kennedy: This should hopefully clean things up a fair bit.

FrankReh (approved):
Looks very good to me!

FrankReh:
Interesting. A few rounds of testing my buf_ring branch with these changes. Nothing is broken but it could be the driver tick isn't being called as often as it had been before. That's interesting. We want to visit a separate mechanism for submission and completion checking anyway.

Noah-Kennedy:
Oh we absolutely need to. Like I've said before, we should do this the same way that tokio invokes epoll_wait, which is to say by leveraging the maintenance hook for periodic polling, and by then making use of parking (as we do now).

FrankReh:
> I suspect that the main impacting change is moving the driver ticking to a separate task.

Me too. The tokio scheduler could make us crazy if we worry too much about which task gets to go most often and how long the ready queue is getting in our stress/flood tests.

Noah-Kennedy:
It would probably be best if we tried to avoid using the task as anything other than a way of getting tokio to unpark on uring completions arriving. I suspect that driving completions on unpark would potentially help improve perf a fair bit, as well as periodically flushing the queue every X ticks by getting something into tokio to hook into the maintenance routine.

PR #148, opened 2022-10-25
Expose runtime structure

### LINKED ISSUE #91: Allow access for Runtime struct?
Hi

I'm trying to benchmark tokio-uring and different implementations. As criterion [requires Runtime struct](https://bheisler.github.io/criterion.rs/book/user_guide/benchmarking_async.html) to run async code, I'd like to ask for changing Runtime struct visibility.

Is it valid change or any other viable solution to achieve it?

Thanks

ollie-etl:
A recreation of #88, which doesn't seem to be updated. See discussion in that PR. Closes #91.

The motivationis as expressed in #91 and #88, to allow access to the runtime. An example of where this is used is criterion. If this is merged and a release issued, I will put a PR in to criterion with `AsyncExecutor` trait impl as a criterion feature. Until then, the following works

```rust
struct AsyncRuntime(tokio_uring::Runtime);

impl criterion::async_executor::AsyncExecutor for &AsyncRuntime {
    fn block_on<T>(&self, future: impl futures::Future<Output = T>) -> T {
        self.0.block_on(future)
    }
}

// ...

fn bench(c: &mut Criterion) {
    // ...
    let mut ring_opts = tokio_uring::uring_builder();  
    let mut builder = tokio_uring::builder();
    builder.uring_builder(&ring_opts);
    let runtime = AsyncRuntime(tokio_uring::Runtime::new(&builder).unwrap());
    let runtime = &runtime;

   c.bench_with_input(BenchmarkId::new("input_example", size), &size, |b, &s| {
        // Insert a call to `to_async` to convert the bencher to async mode.
        // The timing loops are the same as with the normal bencher.
        b.to_async(runtime).iter(|| do_something(s));
    });
}
```

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -79,6 +79,7 @@ pub mod fs;
 pub mod net;
 
 pub use runtime::spawn;
+pub use runtime::Runtime;
 
 use std::future::Future;
 
@@ -140,7 +141,7 @@ use std::future::Future;
 /// }
 /// ```
 pub fn start<F: Future>(future: F) -> F::Output {
-    let mut rt = runtime::Runtime::new(&builder()).unwrap();
+    let rt = runtime::Runtime::new(&builder()).unwrap();
     rt.block_on(future)
 }
 
@@ -225,7 +226,7 @@ impl Builder {
     /// }
     /// ```
     pub fn start<F: Future>(&self, future: F) -> F::Output {
-        let mut rt = runtime::Runtime::new(self).unwrap();
+        let rt = runtime::Runtime::new(self).unwrap();
         rt.block_on(future)
     }
 }
--- a/src/runtime.rs
+++ b/src/runtime.rs
@@ -6,7 +6,8 @@ use std::io;
 use tokio::io::unix::AsyncFd;
 use tokio::task::LocalSet;
 
-pub(crate) struct Runtime {
+/// The Runtime executor
+pub struct Runtime {
     /// io-uring driver
     driver: AsyncFd<Driver>,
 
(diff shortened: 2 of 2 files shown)


ollie-etl:
@FrankReh Is that a suitable usage statement?

FrankReh:
@ollie-etl Thanks for the quick uptake. I have more questions about the commit message than I do the files changes. The changed files look good to me.

Is the usage showing an example bench that would act like a main for criterion benchmarking or does all that setup need to be duplicated for each benchmark or each file that would have benchmarking in it?

FrankReh:
@Noah-Kennedy If these changes look good to you too, will you merge? My question about the running example can wait.

ollie-etl:
@FrankReh With criterion, you can share setup per group, https://bheisler.github.io/criterion.rs/book/user_guide/benchmarking_with_inputs.html#benchmarking-with-a-range-of-values. 

My example is only a single function.

The newtyped trait implementation is something that I'd hope to get up-streamed to criterion. That will require a new published version

FrankReh:
Just going to wait for @Noah-Kennedy to accept this one. It looks good to me. I wish there were an example binary using criterion but we have lots of other functionality that doesn't come with an example binary either so there's no need to start that here.

FrankReh (approved):
These changes look reasonable and are what folks have been asking for to be able to use criterion.

Noah-Kennedy (approved):

Noah-Kennedy:
It looks like the benchmarks spend most of their time in FuturesUnordered. I'm going to try and cleanup our results a bit and see what can be optimized.

flamegraph

Noah-Kennedy:
We should probably move this over to discord or to a dedicated issue thread.

Noah-Kennedy:
Discord seems better.

Noah-Kennedy:
Ah, meant to post this svg in another thread.

Issue #147, opened 2022-10-24
rt: Hang on too small completion queue


ollie-etl:
Similar to #145, and again exercised by #144. If there size of the completion queue is smaller than the number of concurrent writers to the submission queue, the runtime just hangs.

Probably not commiting?

FrankReh:
Yes, the driver may be making too strict assumptions about when to submit and when to check the cqueue.

The driver isn't flushing the squeue and handling the cqueue as often as it could and it can use the overflow bit to make it take an extra loop before getting back to the caller.

And also, like another issue and commit found and handled, it is not enough to just submit when the thread is going idle, there will be times we want submit and tick should run without being idle first. Whether we wait for the squeue to be full or not is debatable. The squeue drains completely with any call to submit so its probably okay to wait for the squeue to become full but the app can decide what the size of the squeue is and if it is made to large, it can starve the kernel from work it could be processing on. Some day I want to ask the uring author what they think the right sizing is, but maybe this crate's benchmark abilities will already have given us some good ideas.

I believe my thinking about this earlier, and my pseudo fix, was any time the squeue is found full, we should run a tick first but we don't want to recurse and call a tick again if a tick completion handler had tried to make its own submission entry and was thwarted by finding the squeue full too. We also don't want to force the kernel to go into overflow mode if we can help it because that creates a slower path and burns cycles that will sometimes be avoidable by checking the cqueue periodically.

And there is also the non-sys call mode which the driver doesn't have an option for yet which would let the two sides run in parallel even more efficiently.

Great ways are being found to stress the system reliably. Will make it easier to push performance and detect when regressions are happening.

FrankReh:
@Noah-Kennedy Are you saying there is already a hook in tokio we can use, or you're going to try and get one added.

If that is new, perhaps before that lands, we can implement our own maintenance countdown in the op creation to trigger the same uring maintenance.

Noah-Kennedy:
I'm going to try to get one added. In the meantime, I'm wondering if it is best to address this in the op creation or to do a "duck tape and zip ties" fix by having an always ready task that no-ops and yields back most of the time like this (pardon my pseudocode):

```rust
async fn maintenance_hack() {
    loop {
        for _ in 0..16 {
            // yield back to the runtime 16 times
            tokio::task::yield_now().await;
        }
        
        // flush the queue and dispatch completions.
        // I'm acting like the thread locals are statics here for simplicity's sake.
        DRIVER.tick();
    }
}
```

[...]

PR #152, opened 2022-10-27
driver: improve cqueue dispatch

Noah-Kennedy:
This PR makes a few changes:
- It changes our ticking behavior to loop on a full cqueue to ensure that we actually fully clear out the queue in the event of an overflow.
- It adds an optimization to our parking routine which allows us to potentially wake tasks instead of parking on epoll.
- It makes our parking routines attempt to tick futures if we hit EBUSY

### DIFF
--- a/src/driver/mod.rs
+++ b/src/driver/mod.rs
@@ -77,20 +77,30 @@ impl Driver {
     }
 
     pub(crate) fn tick(&mut self) {
-        let mut cq = self.uring.completion();
-        cq.sync();
+        loop {
+            let mut cq = self.uring.completion();
 
-        for cqe in cq {
-            if cqe.user_data() == u64::MAX {
-                // Result of the cancellation action. There isn't anything we
-                // need to do here. We must wait for the CQE for the operation
-                // that was canceled.
-                continue;
+            cq.sync();
+
+            // if the cqueue is full, we are experiencing overflow and need to keep looping
+            let should_continue = cq.is_full();
+
+            for cqe in cq {
+                if cqe.user_data() == u64::MAX {
+                    // Result of the cancellation action. There isn't anything we
+                    // need to do here. We must wait for the CQE for the operation
+                    // that was canceled.
+                    continue;
+                }
+
+                let index = cqe.user_data() as _;
+
+                self.ops.complete(index, cqe.into());
             }
 
-            let index = cqe.user_data() as _;
-
-            self.ops.complete(index, cqe.into());
+            if !should_continue {
+                break;
+            }
         }
     }
 
--- a/src/runtime/mod.rs
+++ b/src/runtime/mod.rs
@@ -62,7 +62,22 @@ impl Runtime {
         let rt = tokio::runtime::Builder::new_current_thread()
             .on_thread_park(|| {
                 CONTEXT.with(|x| {
-                    let _ = x.with_driver_mut(|d| d.uring.submit());
+                    let _ = x.with_driver_mut(|d| {
+                        // optimization: we can potentially avoid an epoll_wait call if we try to
+                        // dispatch completions before we park on epoll
+                        d.tick();
+                        d.submit()
+                    });
+                });
+            })
+            .on_thread_unpark(|| {
+                CONTEXT.with(|x| {
+                    let _ = x.with_driver_mut(|d| {
+                        // Dispatch completions to wake tasks based on any completed ops.
+                        // this is an optimization to try and avoid the whole "give the io driver
+                        // a dedicated task" thing
+                        d.tick();
+                    });
                 });
             })
             .enable_all()
@@ -106,11 +121,10 @@ impl Runtime {
 
         self.local.spawn_local(drive);
 
-        self.rt
-            .block_on(self.local.run_until(crate::future::poll_fn(|cx| {
-                // assert!(drive.as_mut().poll(cx).is_pending());
-                future.as_mut().poll(cx)
-            })))
+        self.rt.block_on(
+            self.local
+                .run_until(crate::future::poll_fn(|cx| future.as_mut().poll(cx))),
+        )
     }
 }
 


Noah-Kennedy:
We should not merge this until we have a repro of #147.

FrankReh:
Okay, you have my attention. I have a reproducer. I'll try it out. I can also share the reproducer with you if you want to try it out yourself, along with all those new flames.

FrankReh:
Here's my reproducer if you feel like trying it yourself.

```rust
examples $ cat test_hang_on_small_completion_queue.rs 
use std::io;
use clap::Parser;
use tokio::sync::mpsc;

#[derive(Debug, Parser)]
#[command(about = "Serve TCP connections, testing the buf_ring feature. Client can be 'nc <ip-addr> <port>'.", long_about = None)]
struct Args {
    /// Spawn count, how many no_op tasks to spawn consecutively
    #[arg(long, default_value_t = 16)]
    spawn_cnt: usize,

    /// Squeue entries
    #[arg(long, default_value_t = 4)]
    squeue_entries: u32,

    /// Cqueue entries
    #[arg(long, default_value_t = 0)]
    cqueue_entries: u32,
}


fn main() -> io::Result<()> {
    let args = Args::parse();

    let spawn_cnt = args.spawn_cnt;
    let squeue_entries = args.squeue_entries;
    let cqueue_entries = std::cmp::max(args.cqueue_entries, 2* squeue_entries);

    tokio_uring::builder()
        .entries(squeue_entries)
        .uring_builder(tokio_uring::uring_builder().setup_cqsize(cqueue_entries))
        .start( async move {

            let (tx, mut rx) = mpsc::channel(2*spawn_cnt);

            for i in 0..spawn_cnt {
                let tx = tx.clone();
                    tokio_uring::no_op().await.unwrap();
                    tx.send(i).await.unwrap();
                    Ok::<(), io::Error>(())
                });
            }

            while let Some(i) = rx.recv().await {
                println!("got = {}", i);
            }
        });

    Ok(())
}
```

FrankReh:
And ran it with ' --spawn-cnt 100 --squeue-entries 2' the last when it didn't work. It didn't print out all 100 lines.

FrankReh:
It could be that when I was working on this, last month I think, I had to change my version of io-uring too, and without proof of whether it was needed or not, I didn't  push to get it upstreamed yet.  The crux is the uring interface won't push its overflow to the cqueue unless IORING_ENTER_GETEVENTS is set in the call to enter. I don't think the api let's you do that from tokio-uring manually.

Here's the basic diff. By having io-uring also check for the overflow bit, it knows when it should set the bit, otherwise since we don't *want* more than 0, that bit doesn't get set.
```rust
Before
    pub fn submit_and_wait(&self, want: usize) -> io::Result<usize> {
        let len = self.sq_len();
        let mut flags = 0;

        if want > 0 || self.params.is_setup_iopoll() {
            flags |= sys::IORING_ENTER_GETEVENTS;
        }
After
    pub fn submit_and_wait(&self, want: usize) -> io::Result<usize> {
        let len = self.sq_len();
        let mut flags = 0;

        // TODO is this the right way to handle the overflow, from this crate's perspective?
        // Will need to ask the maintainer.
        // Was
        //    if want > 0 || self.params.is_setup_iopoll() {..}
        if want > 0 || self.params.is_setup_iopoll() || self.sq_cq_overflow() {
            flags |= sys::IORING_ENTER_GETEVENTS;
        }

And sq_cq_overflow is defined like this:
    fn sq_cq_overflow(&self) -> bool {
        unsafe {
            atomic::fence(atomic::Ordering::SeqCst); // Refer to comment in squeue.rs for same code
            (*self.sq_flags).load(atomic::Ordering::Acquire) & sys::IORING_SQ_CQ_OVERFLOW != 0
        }
    }
```

Noah-Kennedy:
@FrankReh I've slightly modified it and added it as a test.

FrankReh:
@Noah-Kennedy That's nice. We can tweak the numbers. Something a little smaller or perhaps larger, if it is truly fast enough once it works. If we can make it gracefully timeout and report an error, that would be more user friendly. Maybe the std test timeout is short enough, and it keeps the logic simple.

ollie-etl:
> We should not merge this until we have a repro of #147.

Simply increase the concurreny factor in the benchmark past the size of the cqe.

tests/driver.rs
```diff
@@ -90,6 +91,28 @@ fn too_many_submissions() {
     });
 }
 
+#[test]
+fn completion_overflow() {
+    let spawn_cnt = 100;
+    let squeue_entries = 2;
+    let cqueue_entries = 2 * squeue_entries;
```
FrankReh, line 98: How about adding a second timeout so when this does hang, it fails relatively quickly. I found this worked well

```rust
    std::thread::spawn(|| {
        thread::sleep(time::Duration::from_millis(1000));
        eprintln!("Test failed, timeout reached.");
        process::exit(1);
    });
```

FrankReh (changes requested):
Just reviewing the unit test.

I haven't looked at the rest of the changes to tick as I thought getting the hang fixed could be done without any of these changes first.

tests/driver.rs
Noah-Kennedy: I don't think that using `std::thread` and `process::exit` is a good way of doing this. If we need to do this with a timeout, I'd rather we wrap an async block in `tokio::time::timeout`.

Generally for things like this, I prefer to move the main logic off to a spawned task and to have the entry task run `tokio::task::yield_now().await` a bunch of times in a row before checking if we are finished or not and exiting. This allows us to be bounded by how many cycles through the scheduler we take rather than by time elapsed, and is far more deterministic in CI.

FrankReh: Am looking for an example of this in a test. Would like to get a version of this unit test committed this morning with a timeout in it now that io-uring 0.5.8 is available - and I'm preparing a commit to bump the io-uring dependency in tokio-uring's Cargo.toml.

FrankReh: I couldn't get any option like that to work. It's always possible that when it's hung, it is really blocking the entire thread.

FrankReh:
Can this wait until the older PRs are in now? I don't believe this PR is a blocker for other PRs or issues. I would like to participate in looking at benchmarks once the benchmarking PR is in and once the older PRs are handled.

Noah-Kennedy:
This can absolutely wait

ollie-etl:
@Noah-Kennedy See #160

PR #153, opened 2022-10-28
Depend on io-uring version ^0.5.8

FrankReh:
io-uring v0.5.8 handles the overflow of the cqueue so this bump on the dependancy fixes hangs caused when high frequency submissions fill the cqueue.

While we're at it, adds a unit test for such a hanging cqueue.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -21,8 +21,8 @@ tokio = { version = "1.2", features = ["net", "rt"] }
 scoped-tls = "1.0.0"
 slab = "0.4.2"
 libc = "0.2.80"
-io-uring = { version = "0.5.0", features = [ "unstable" ] }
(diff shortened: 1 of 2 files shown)


FrankReh:
@Noah-Kennedy I know the unit test isn't structured to your liking. I'm happy to change it if you can give me more help with it. The  things I tried this morning didn't work, I suspect because the hang is really a thread block, otherwise I can't explain why my other versions just hung there indefinitely.

FrankReh:
@Noah-Kennedy At least I can say it's nice to see this ci test passing, even with this unit test in. It wouldn't have passed with io-uring version =0.5.7.

Noah-Kennedy:
Let's just merge this now and cleanup later.

Noah-Kennedy (approved):

PR #155, opened 2022-10-29
rt: drop runtime before driver during shutdown

Noah-Kennedy:
This fixes the issue where we can potentially drop the driver while there are active tasks on the runtime.

### DIFF
--- a/src/runtime/mod.rs
+++ b/src/runtime/mod.rs
@@ -24,6 +24,15 @@ pub struct Runtime {
 
     /// Tokio runtime, always current-thread
     rt: tokio::runtime::Runtime,
+
+    /// This is here for drop order reasons.
+    ///
+    /// We can't unset the driver in the runtime drop method, because the inner runtime needs to
+    /// be dropped first so that there are no tasks running.
+    ///
+    /// The rust drop order rules guarantee that the inner runtime will be dropped first, and
+    /// this last.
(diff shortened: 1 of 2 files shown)


FrankReh:
This looks nice. Did you find it easy to reproduce the crash without this? Which stack trace did it look like, the first or the second? Or sometimes either one?

src/runtime/mod.rs
```diff
@@ -24,6 +24,15 @@ pub struct Runtime {
 
     /// Tokio runtime, always current-thread
     rt: tokio::runtime::Runtime,
+
+    /// This is here for drop order reasons.
+    ///
+    /// We can't unset the driver in the runtime drop method, because the inner runtime needs to
```
FrankReh, line 30: What is the 'inner' runtime referred to here?

FrankReh (changes requested):

src/runtime/mod.rs
Noah-Kennedy: The tokio runtime we store.

FrankReh (approved):

FrankReh:
I didn't mean to delete the noah/fix-shutdown branch. I don't know if my issued a tap I didn't mean to or if it happened automatically. Oh well. You've still got it if you want to push it up again for some reason.

Noah-Kennedy:
No worries about deleting the branch, you saved me a click.

PR #156, opened 2022-10-30
poll_fn and Unpin: fix pinning

FrankReh:
Keeping this crate's poll_fn in sync with that of rust's library/core/src/future/poll_fn.rs.

Their merge provides all the details.

    https://github.com/rust-lang/rust/pull/102737

None of this crate's uses of poll_fn are affected. It protects against misuse in the future which could have led to UB as the above mentioned PR points out with an example of a double free becoming possible.


FrankReh:
@Noah-Kennedy Please check that my changes look the ones introduced with that PR referenced because I did them by hand. I didn't pull the git commit and apply it to this file.

Noah-Kennedy:
Thinking about this now, do we even need our own `poll_fn`? This was added when `poll_fn` wasn't stable in the stdlib, and the tweaked version of this function will be stable in several days.

FrankReh:
Does that mean tokio-uring MSRV becomes 1.64.0?

Noah-Kennedy:
We don't really have an MSRV policy for tokio-uring.

FrankReh:
So far, this crate has been piggy backing off tokio. And tokio has an MSRV that goes back six months. We could merge this with a note saying it should be removed when 1.64.0 becomes the tokio MSRV. So we continue to track tokio.

Noah-Kennedy:
I'm not entirely sure which older Rust versions we work on currently. We may not even work on tokio's oldest supported version, and I don't see any reason that we should maintain an MSRV at this point.

ollie-etl:
My view: this crate is marked as both experimental and likely to changed. We should not even be attempting to keep a MSRV in tokio or rustc.

FrankReh:
I would just like there to be a better reason someone had to  use the 1.64.0 compiler. For this, it seems easy to not change what we were doing.

ollie-etl:
It could always be punted to an issue

FrankReh:
@Noah-Kennedy what should we do with this? I think it safer to just merge the changes and remove the code at a later time. But the other way is okay too.

Noah-Kennedy:
Let's close this and just get rid of our version of this method.

FrankReh:
Obsoleted given merge of #194

PR #157, opened 2022-10-30
rt: refactor drop logic

Noah-Kennedy:
This shouldn't change any functionality, but it does use ManuallyDrop to cleanup the drop logic a fair bit. This is much nicer than relying on drop order.

### DIFF
--- a/src/runtime/mod.rs
+++ b/src/runtime/mod.rs
@@ -1,6 +1,5 @@
 use crate::driver::Driver;
 
-use libc::servent;
 use std::future::Future;
 use std::io;
 use std::mem::ManuallyDrop;
--- a/src/runtime/mod.rs
+++ b/src/runtime/mod.rs
@@ -1,7 +1,9 @@
 use crate::driver::Driver;
 
+use libc::servent;
 use std::future::Future;
 use std::io;
+use std::mem::ManuallyDrop;
 use std::os::unix::io::{AsRawFd, RawFd};
 use tokio::io::unix::AsyncFd;
 use tokio::task::LocalSet;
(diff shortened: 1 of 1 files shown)


src/runtime/mod.rs
```diff
@@ -124,10 +118,16 @@ impl Runtime {
...
+impl Drop for Runtime {
     fn drop(&mut self) {
+        // drop tasks
+        unsafe {
+            ManuallyDrop::drop(&mut self.local);
+            ManuallyDrop::drop(&mut self.rt);
+        }
+
+        // once tasks are dropped, we can unset the driver
+        // this will block until all completions are received
```
FrankReh, line 130: Interesting. This shows the unset happens after local tasks and rt are dropped. Well, if that works, it seems very clean. The order is surprising given the order of creation in new, but I guess that's the point.

FrankReh (approved):
I like it. Less code but more  importantly, makes reasoning about the order of things very straight forward.

src/runtime/mod.rs
```diff
@@ -1,7 +1,9 @@
 use crate::driver::Driver;
 
+use libc::servent;
```
FrankReh, line 3: This line is breaking CI it appears.

FrankReh (changes requested):

src/runtime/mod.rs
Noah-Kennedy: Yeah, this got pulled in by mistake. I'll pull it out once I finish another PR.

Noah-Kennedy: Fixed.

Noah-Kennedy:
@FrankReh its ready for re-review.

FrankReh:
On my cell. Be home in an hour. 

-Frank

FrankReh:
If ci passes now. Squash commit yourself and don’t wait for me. Code looked good before. 

-Frank

Noah-Kennedy:
I need to wait for your approval because you requested changes.

FrankReh (approved):

FrankReh:
Somehow the system deletes the branch automatically just because I squashed committed. It wasn't me. I promise. ;)

PR #162, opened 2022-11-02
rt: fix error when calling block_on twice

Noah-Kennedy:
This fixes an issue where we would try and spawn a separate background task on the `LocalSet` for watching the `AsyncFd` for the uring instance every time we called `block_on`. Now, we only spawn at runtime creation.

### DIFF
--- a/src/runtime/mod.rs
+++ b/src/runtime/mod.rs
@@ -3,7 +3,7 @@ use crate::driver::Driver;
 use std::future::Future;
 use std::io;
 use std::mem::ManuallyDrop;
-use std::os::unix::io::{AsRawFd, RawFd};
+use std::os::unix::io::AsRawFd;
 use tokio::io::unix::AsyncFd;
 use tokio::task::LocalSet;
 
@@ -17,9 +17,6 @@ thread_local! {
 
 /// The Runtime executor
 pub struct Runtime {
-    /// io-uring driver
-    uring_fd: RawFd,
-
     /// LocalSet for !Send tasks
     local: ManuallyDrop<LocalSet>,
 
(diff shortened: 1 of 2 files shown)


Noah-Kennedy:
CC @ollie-etl @FrankReh

src/runtime/mod.rs
```diff
@@ -105,9 +90,17 @@ impl Runtime {
             }
         };
 
-        tokio::pin!(future);
+        local.spawn_local(drive);
```
FrankReh, line 93: Here's the big change. Spawning the driver block at creation, rather than per block_on call.

Interesting. Here we let driver.tick get called whenever the fd becomes ready, even if there is no block_on happening.

It doesn't hurt anything that the complete calls the tick can invoke may try to call an op future's waker?

Oh ... I bet the LocalSet is only live, during the rt.block_on call. And when the block_on is done, the LocalSet goes quiescent again until/if another rt.block_on call is made.

Very interesting indeed.

It's nice how futures are designed to just be dropped, with no callbacks needing to be executed. I guess a future can always have its own drop method defined for cleanup if necessary.

FrankReh (approved):
This is one of those changes, if it works at least as well as things did before, it looks good. Without a formal proof of correctness, who's to say whether some other corner case is still lurking. I like it.

src/runtime/mod.rs
Noah-Kennedy: Yup, the `LocalSet` only runs when run during the block_on call. It runs on the current thread, not in a background one.

src/runtime/mod.rs
```diff
@@ -105,9 +90,17 @@ impl Runtime {
...
-        self.local.spawn_local(drive);
+        Ok(Runtime { local, rt })
+    }
+
+    /// Runs a future to completion on the current runtime
+    pub fn block_on<F>(&self, future: F) -> F::Output
+    where
+        F: Future,
+    {
+        tokio::pin!(future);
```
FrankReh, line 103: Do you mean to leave the commented code at line 107?

src/runtime/mod.rs
```diff
@@ -105,9 +90,17 @@ impl Runtime {
...
 
         self.rt
             .block_on(self.local.run_until(crate::future::poll_fn(|cx| {
```
FrankReh, line 106: Oh, self.local.run_until(poll_fun(... future.as_mut().poll ...))

That's a good synchronous stack to get one's async head around.

FrankReh:
Just one question about the commented line 107 really.

src/runtime/mod.rs
Noah-Kennedy: I forgot that was there lol. We can delete this later in a cleanup PR.

ollie-etl:
@Noah-Kennedy Nice

Issue #161, opened 2022-10-31
Generialise Buffer Types


ollie-etl:
Using io_uring requires ownership of the backing buffers for the duration of a Operation lifetime. This is currently enforced using the traits `IoBuf` and `IoBufMut`. This issue is suggesting dropping IoBuf and IoBufMut in favour of `Pin`. In conjunction with a slightly modified Slice struct, I believe `Pin` provides all the invariant we currently have, whilst allowing us to generalize over any input type (without newtyping everything)

I'd like to illustrate my proposal by considering the Op<Write>

```rust
// current
impl<T: IoBuf> Op<Write<T>> {
    pub(crate) fn write_at(fd: &SharedFd, buf:T, offset: u64) -> io::Result<Op<Write<T>>> {
        // ...
    }
}

// Pr #53
impl<T: IoBuf> Op<Write<T>> {
    pub(crate) fn write_at(fd: &SharedFd, buf: Slice<T>, offset: u64) -> io::Result<Op<Write<T>>> {
        // ...
    }
}

// Proposed
impl<T: Deref<Target = [u8]>> Op<Write<T>> {
    pub(crate) fn write_at(fd: &SharedFd, buf: Pin<Slice<T>>, offset: u64) -> io::Result<Op<Write<T>>> {
    }
}
```

 A PR will be along shortly to give a concrete reference point, but I wanted to put this out there for discussion

FrankReh:
I thought it would be a &mut to a pinned slice. So here the Op takes the ownership, as before, and has to return it when the future is complete?

ollie-etl:
That was my thoughts, yes. We know (by construction) that as long as we hold the pin, the buffer is stable, but we don't have to handle the lifetime in the Op

ollie-etl:
Although my thinking is likely to chnge as I will undoubted hit compile errors en-route

FrankReh:
> but we don't have to handle the lifetime in the Op

I don't follow. Just being dense I think. So we can't hold a &mut Pin<Slice<T>> for the lifetime of the Op? The type doesn't have to be 'static? Aren't we saying the buffer's memory is 'static anyway, but otherwise taking ownership of it as long as the future is active?

ollie-etl:
@FrankReh I called it a night, because its late here, but having thought about it for 5 minutes - you are of course completely correct `&Pin<Slice<T>>` is the correct thing. I'll take a look when fresh

FrankReh:
[...]

PR #163, opened 2022-11-03
Fix chained .slice to not produce nested Slice wrappers

mzabaluev:
A rework of #53 without introducing a utility trait, based in part on suggestions in https://github.com/tokio-rs/tokio-uring/pull/53#discussion_r1011622932

`Slice` no longer implements `IoBuf`, but gains an inherent `slice` method that returns a `Self` subslice.
There is `impl<T: IoBuf> From<T> for Slice<T>`, and the generic buffers in the API methods are bound as `impl Into<Slice<T>>`.
Like in #53, the buffers are returned back as `Slice<T>`.


mzabaluev:
> I might pull both branches and run git vimdiff against the files in the two tips unless there is an easier way I can be pointed to.

@FrankReh This branch only adds 99aef458fbbb9d6d16f5a2392408641818b3bbf5 and a small doc revert on top of #53.

FrankReh:
@mzabaluev Thanks for the confirmation. I've started taking a look.

FrankReh:
Random musings:

rust 1.65 is released today. Does the GAT support make any of this crate's IoBuf/Slice support even cleaner?

Should some of the crate's unit tests work with the bytes type when the bytes feature is enabled?

ollie-etl:
I still think we could go further, although maybe these items are for a different PR

### Remove IoBuf
IoBuf and IoBufMut are used for several purposes:
1. To ensure that buffers are stable for the duration of an Op
2. To access the pointer to the start of the buffer, 
3. To get and mark the initialised bytes

w.r.t `1.`, the same thing can be achieve by using the constraint `T: Unpin + 'static` where we currently use `T: IoBuf`

w.r.t `2.` we can achieve the same by using `T: AsRef<[u8]>` where we currently use `T: IoBuf`, and `T: AsMut<[u8]>` where we currently use `T: IoBufMut`, 

w.r.t `3.` This can be managed with Slice directly

FrankReh:
@ollie-etl Interesting. Yes, let me finish my review. @Noah-Kennedy and @mzabaluev can then decide what to do from there.

Noah-Kennedy:
@ollie-etl @FrankReh I went over this on #161, but IoBuf is needed, at the very least as a marker type.

Noah-Kennedy:
Unpin just means "I can move this", it does not guarantee a stable pointer from a deref. We need our own trait to signal this.

src/net/tcp/listener.rs
```diff
@@ -31,9 +31,9 @@ use std::{io, net::SocketAddr};
 ///
 ///     tx.write(b"test" as &'static [u8]).await.0.unwrap();
 ///
-///     let (_, buf) = rx.read(vec![0; 4]).await;
+///     let (_, slice) = rx.read(vec![0; 4]).await;
 ///
-///     assert_eq!(buf, b"test");
+///     assert_eq!(slice.to_vec(), b"test");
```
FrankReh, line 36: That's cool. Almost magic that the Slice has to_vec without this crate defining it. That's what ops::Deref gives us?

src/net/tcp/stream.rs
```diff
@@ -113,14 +119,13 @@ impl TcpStream {
...
-    ///                 let (result, nbuf) = stream.read(buf).await;
-    ///                 buf = nbuf;
+    ///                 let (result, slice) = stream.read(buf).await;
     ///                 let read = result.unwrap();
     ///                 if read == 0 {
     ///                     break;
     ///                 }
     ///
-    ///                 let (res, slice) = stream.write_all(buf.slice(..read)).await;
+    ///                 let (res, slice) = stream.write_all(slice.slice(..read)).await;
```
FrankReh, line 128: Wonder how many times people will forget they need to qualify the slice they were given by the value returned. 

And will we come up with a macro, kind of like ready!, for reads that return a result and the slice where the error is returned if the result was an error, and otherwise unwraps the read integer?

Not in this PR of course.

src/net/tcp/stream.rs
```diff
@@ -131,23 +136,26 @@ impl TcpStream {
...
                         )),
-                        slice.into_inner(),
+                        slice.into_inner().slice(in_begin..in_end),
                     )
                 }
-                (Ok(m), slice) => {
-                    n += m;
-                    buf = slice.into_inner();
+                (Ok(n), slice) => {
+                    buf = slice.slice(n..);
```
FrankReh, line 158: I'm beginning to read slice.slice as not a dup of anything any longer. Took a few hours but now I read the first as a noun and the second as the verb acting on the noun. I'm guessing everyone else here had already been doing that.

tests/buf.rs
```diff
@@ -110,31 +111,83 @@ macro_rules! test_slice {
                     assert_eq!(&slice[5..10], &DATA[5..10]);
                     assert_eq!(&slice[..5], &DATA[..5]);
                 }
+
+                #[test]
```
FrankReh, line 115: Just saw this. Ignore my other comment about testing slice of slice.

FrankReh (changes requested):
Nice. Sorry that was a lot of of signatures to change.

src/buf/slice.rs
```diff
@@ -124,6 +124,55 @@ impl<T> Slice<T> {
     }
 }
 
+impl<T: IoBuf> Slice<T> {
```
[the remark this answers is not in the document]
mzabaluev: I think the general practice is to only request type parameter bounds on APIs that need them.

src/net/tcp/listener.rs
mzabaluev: Yes that's available through `Deref`. But it gives a bad usage example, there is no need to copy the data. Fixed this to access the byte slice through `Deref`.

src/net/udp.rs
```diff
@@ -169,27 +169,36 @@ impl UdpSocket {
     /// number of bytes written.
     pub async fn send_to<T: IoBuf>(
         &self,
-        buf: T,
+        buf: impl Into<Slice<T>>,
```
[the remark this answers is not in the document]
mzabaluev: The mutability contract is enforced through the `IoBuf`/`IoBufMut` bounds. For example, it's not possible to give a `&'static [u8]` to `read` because it does not implement `IoBufMut`. But I don't think the compiler is able to take this into account for dynamic buffers unless some deep inlining is possible; per the method signature, the input and return buffers may as well be different values altogether.

src/net/tcp/stream.rs
mzabaluev: I've been thinking of giving the returned slice an additional purpose. Perhaps the slice returned by `read` could be truncated to the bytes that were read.

src/fs/file.rs
```diff
@@ -173,9 +173,13 @@ impl File {
     ///     })
     /// }
     /// ```
-    pub async fn read_at<T: IoBufMut>(&self, buf: T, pos: u64) -> crate::BufResult<usize, T> {
+    pub async fn read_at<T: IoBufMut>(
+        &self,
+        buf: impl Into<Slice<T>>,
+        pos: u64,
+    ) -> crate::BufResult<usize, Slice<T>> {
```
mzabaluev, line 180: To follow https://github.com/tokio-rs/tokio-uring/pull/163#discussion_r1013275409, perhaps the return type could be `BufResult<(), Slice<T>>`, and the returned slice's end is adjusted to reference the bytes that have been read. But that's a more extensive API change.

src/fs/file.rs
```diff
@@ -336,8 +340,12 @@ impl File {
     /// ```
     ///
     /// [`Ok(n)`]: Ok
-    pub async fn write_at<T: IoBuf>(&self, buf: T, pos: u64) -> crate::BufResult<usize, T> {
-        let op = Op::write_at(&self.fd, buf, pos).unwrap();
+    pub async fn write_at<T: IoBuf>(
+        &self,
+        buf: impl Into<Slice<T>>,
+        pos: u64,
+    ) -> crate::BufResult<usize, Slice<T>> {
```
mzabaluev, line 347: Similarly, the returned slice here could refer to the unwritten portion of the input buffer/slice.

examples/udp_socket.rs
```diff
@@ -15,8 +15,9 @@ fn main() {
 
         let buf = vec![0u8; 128];
 
-        let (result, mut buf) = socket.recv_from(buf).await;
+        let (result, buf) = socket.recv_from(buf).await;
         let (read, socket_addr) = result.unwrap();
+        let mut buf = buf.into_inner();
```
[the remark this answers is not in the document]
mzabaluev: I think `buf.truncate(read)` is sufficient here, too. But perhaps `&buf[..read]` is the least work-intensive way to get the received bytes.

src/fs/file.rs
FrankReh: Yes. If it hasn't been done, yet, now is probably not the time. This ownership API is probably new for anyone coming to this runtime and is going to take some getting used to.

src/fs/file.rs
FrankReh: Yes. Probably not for this PR unless @Noah-Kennedy chimes in and says this is the right time to make that leap. I suspect it is something he and @carllerche are going to want to think about. The whole idea of a buffer representing filled data or initialized data is something that tokio too is dealing with and my last reading was they weren't happy with their APIs either.

Noah-Kennedy:
I should be able to take a look at this tonight.

src/fs/file.rs
ollie-etl: Another way of thinking about this (not in this PR), is when we do a read, should we read into the beginning of the buffer, or the first uninitialised byte?

Because if its into the first uninitialised byte, the returned usize still serves a purpose.

src/net/tcp/stream.rs
ollie-etl: @mzabaluev I absolutely agress with this - following a read, the buffer you get back should only contain valid data.

ollie-etl: I think that is current behavior?

src/fs/file.rs
FrankReh: I think the use of Slice as an input to the read op makes this easy since the definition can be to always read into the beginning of the slice. Let the caller define the slice to start at the uninitialized spot if they like. And I think that's what is currently done.

src/net/tcp/stream.rs
mzabaluev: @ollie-etl No, both in the current API and in this PR the returned buffer/slice has the same range as the input. The length of the bytes updated by read is returned in `Ok(n)`, but the buffer view may also expose pre-existing data in the rest of its initialized part.

src/fs/file.rs
mzabaluev: The reading starts into the beginning of the buffer. This is consistent with the behaviour of `tokio::io` and `std::io`, where similar buffer adapters are used. Buffers are meant to be reused and overwritten, but a buffer should not need to be completely initialized (the innate problem of `Read::read`) in order to be used safely.

src/net/tcp/stream.rs
FrankReh: Resolving this. I think we agree we're getting as close a behavior as possible to before the PR, the Slice<T> being returned not withstanding.

src/fs/file.rs
FrankReh: Agree.

src/net/tcp/stream.rs
```diff
@@ -131,23 +145,30 @@ impl TcpStream {
...
-            let res = self.write(buf.slice(n..)).await;
+    pub async fn write_all<T: IoBuf>(
+        &self,
+        buf: impl Into<Slice<T>>,
+    ) -> crate::BufResult<(), Slice<T>> {
+        self.write_all_slice(buf.into()).await
+    }
+
+    // The implementation of write_all, factored out to reduce code bloat.
+    async fn write_all_slice<T: IoBuf>(&self, mut buf: Slice<T>) -> crate::BufResult<(), Slice<T>> {
```
FrankReh, line 156: Nice catch. @Noah-Kennedy has said methods like write_all will be moved to a trait some time in the future.

src/net/unix/stream.rs
```diff
@@ -96,24 +111,31 @@ impl UnixStream {
...
+        self.write_all_slice(buf.into()).await
+    }
+
+    // The implementation of write_all, factored out to reduce code bloat.
+    async fn write_all_slice<T: IoBuf>(&self, mut buf: Slice<T>) -> crate::BufResult<(), Slice<T>> {
         // This function is copied from the TcpStream impl.
-        let mut n = 0;
-        while n < buf.bytes_init() {
-            let res = self.write(buf.slice(n..)).await;
+        let (in_begin, in_end) = (buf.begin(), buf.end());
```
FrankReh, line 124: Nice catch, getting tcp and unix in sync again. Another reason it will be nice to have one trait that covers both write_all in the future.

FrankReh (approved):
Very nice. Waiting for @Noah-Kennedy to review.

FrankReh:
@mzabaluev Another PR was merged today that added two functions, write_all_at and read_exact_at. Do you want to update your branch and handle those two, they come with a unit test each I think. Or would you like me to?

mzabaluev:
@FrankReh I will update the branch.

src/fs/file.rs
```diff
@@ -491,31 +515,32 @@ impl File {
...
+                        slice.into_inner().slice(in_begin..in_end),
                     )
                 }
                 Ok(n) => {
                     bytes_written += n;
+                    buf = slice.slice(n..);
                 }
-                Err(ref e) if e.kind() == io::ErrorKind::Interrupted => {}
-                Err(e) => return (Err(e), buf),
+                Err(ref e) if e.kind() == io::ErrorKind::Interrupted => {
```
mzabaluev, line 536: I wonder if `EINTR` should occur with the way this runtime uses io-uring. The `write_all` implementations for socket streams says it does not happen.

mzabaluev: @FrankReh please comment whether this match arm is needed at all, or is it effectively dead code.

FrankReh: Yes, it is not needed. Could be called dead code. True for both functions that were recently added to this file.

mzabaluev: OK, I thought there might be a separate PR to fix this because this issue now exists in master.

src/fs/file.rs
```diff
@@ -331,7 +331,7 @@ impl File {
     ///         let (res, buffer) = f.read_exact_at(buffer, 0).await;
     ///         res?;
     ///
-    ///         println!("The bytes: {:?}", buffer);
+    ///         println!("The bytes: {:?}", buffer.into_inner());
```
FrankReh, line 335: This line 334 println needed changing? If the Slice can't be printed, is that because it doesn't define a debug fmt that pretty prints just the text? What does Vec do, when it's not promised to be holding a sequence of valid UTF-8? I wonder.

It is subtle why this works without qualifying the Vec with the result bytes, at least to me, that into_inner will return the vec but now with a length that is no longer zero. I wonder what will become idiomatic for these returned variables, buffer or slice. I prefer slice but that's just me, I like the name reflecting the type just a little bit when it's convenient, but I know there's just as many people who don't think the name should reflect the type. JTOL.

src/fs/file.rs
```diff
@@ -303,19 +303,20 @@ impl File {
...
+    /// covering the same range of the buffer as was passed to the method.
     ///
     /// If the method returns [`Ok(())`], then the read was successful.
     ///
     /// # Errors
     ///
     /// If this function encounters an "end of file" before completely filling
     /// the buffer, it returns an error of the kind [`ErrorKind::UnexpectedEof`].
-    /// The buffer is returned on error.
+    /// The buffer is returned on error as described above.
```
FrankReh, line 316: Maybe add, "but the data that could be read will have been written to the buffer.

So it's clear that when there's an error, the buffer may have been written to somewhat. These write_all type functions are just helpers and can't be expected to be magic. If the caller cares, they should make their own copy beforehand.

FrankReh (changes requested):
Sorry, unless I missed the explanation somewhere else in the comment, there is probably one comment that could be made clearer.

Your comments are great and you're being super thorough.

src/fs/file.rs
mzabaluev: `read_exact_at` only succeeds when it fills the entire buffer, so that `Vec` is going to contain 10 bytes. The `Debug` impl prints them out in the form of array of decimal numbers.

src/driver/read.rs
```diff
@@ -11,12 +11,12 @@ pub(crate) struct Read<T> {
...
 
-    /// Reference to the in-flight buffer.
-    pub(crate) buf: T,
+    /// Owns the in-flight buffer.
+    buf: Slice<T>,
 }
 
 impl<T: IoBufMut> Op<Read<T>> {
-    pub(crate) fn read_at(fd: &SharedFd, buf: T, offset: u64) -> io::Result<Op<Read<T>>> {
+    pub(crate) fn read_at(fd: &SharedFd, buf: Slice<T>, offset: u64) -> io::Result<Op<Read<T>>> {
```
Noah-Kennedy, line 19: I'm not sure how I feel about changing the APIs like this. Why is this necessary?

mzabaluev: I also feel unhappy about `Slice` popping up everywhere even when not needed, and perhaps it's not justified by fixing `.slice(..).slice(..)`.

I've got another idea which I'm going to try first.

mzabaluev:
Superseded by #172.

PR #164, opened 2022-11-04
driver: MultiCqe support for Driver Drop

ollie-etl:
Handles `LifeCycle::CompletionList` in driver Drop logic

During the first pass over all lifecycles, identifies if CompletionLists are finished or not. 

  - If they are, it changes the Lifecycle to Completed with null data.
  - If they are not, adds them to the cancel list, and converts to Ignored

This means that after this pass, we will never observe `CompletionList` in the lifecycles. This means the existing termination logic is ok unchanged. The ignored entry will stay until removed from the Lifecycle slab by the call to complete.

Fixes a gap left by #158 and unblocks #123

### DIFF
--- a/src/driver/mod.rs
+++ b/src/driver/mod.rs
@@ -145,10 +145,39 @@ impl Drop for Driver {
 
         // pre-determine what to cancel
         let mut cancellable_ops = Vec::new();
-        for (id, cycle) in self.ops.lifecycle.iter() {
-            // don't cancel completed items
-            if !matches!(cycle, Lifecycle::Completed(_)) {
-                cancellable_ops.push(id);
+
+        let Ops {
+            lifecycle,
+            completions,
+        } = &mut self.ops;
+
+        for (id, cycle) in lifecycle.iter_mut() {
+            match cycle {
+                Lifecycle::Completed(_) => {
+                    // don't cancel completed items
+                }
+
+                Lifecycle::CompletionList(indices) => {
+                    let mut list = indices.clone().into_list(completions);
+                    if io_uring::cqueue::more(list.peek_end().unwrap().flags) {
+                        // If the completion list is not finished,
+                        // mark for cancellation, and mark the lifecycle as ignored
+                        cancellable_ops.push(id);
+                        *cycle = Lifecycle::Ignored(Box::new(()));
+                        // Dropping the list here deallocates the unconsumed cqe slab entries
+                    } else {
+                        // This op is complete. Replace with a null Completed entry
+                        *cycle = Lifecycle::Completed(op::CqeResult {
+                            result: Ok(0),
+                            flags: 0,
+                        });
+                    }
+                }
+
+                _ => {
+                    // All other states need cancelling
+                    cancellable_ops.push(id)
+                }
             }
         }
 
@@ -166,8 +195,6 @@ impl Drop for Driver {
             }
         }
 
-        // TODO: add a way to know if a multishot op is done sending completions
-        // SAFETY: this is currently unsound for multishot ops
         while !self
             .ops
             .lifecycle
--- a/src/driver/mod.rs
+++ b/src/driver/mod.rs
@@ -202,7 +202,7 @@ impl Drop for Driver {
             };
 
             // Remove completed lifecycles from Slab.
-            // This prevents worst case quadtratic processing
+            // This prevents worst case quadratic processing
             if let Some(id) = remove {
                 self.ops.lifecycle.remove(id);
             } else {
--- a/src/driver/mod.rs
+++ b/src/driver/mod.rs
@@ -188,11 +188,11 @@ impl Drop for Driver {
             }
         }
 
-        // Wait until all Lifetimes have been removed from the slab. 
+        // Wait until all Lifetimes have been removed from the slab.
         //
         // Ignored entries will be removed from the Lifecycle slab
         // by the complete logic called by `tick()`
-        // 
+        //
         // Completed Entries are removed here directly
         let mut id = 0;
         loop {
--- a/src/driver/mod.rs
+++ b/src/driver/mod.rs
@@ -202,7 +202,7 @@ impl Drop for Driver {
                 let _ = self.wait();
                 self.tick();
             } else {
-                id +=1;
+                id += 1;
             };
         }
     }
--- a/src/driver/mod.rs
+++ b/src/driver/mod.rs
@@ -143,8 +143,8 @@ impl Drop for Driver {
             self.submit().expect("Internal error when dropping driver");
         }
 
-        // pre-determine what to cancel
-        // All Cancallable Ops are marked ignored
+        // Pre-determine what to cancel
+        // After this pass, all LifeCycles will be marked either as Completed or Ignored, as appropriate
         for (_, cycle) in self.ops.lifecycle.iter_mut() {
             match std::mem::replace(cycle, Lifecycle::Ignored(Box::new(()))) {
                 lc @ Lifecycle::Completed(_) => {
@@ -165,6 +165,7 @@ impl Drop for Driver {
 
                 _ => {
                     // All other states need cancelling.
+                    // The mem::replace means these are now marked Ignored.
                 }
             }
         }
@@ -187,8 +188,12 @@ impl Drop for Driver {
             }
         }
 
-        // Wait until all in flight cancellations have stopped
-        // Manually iterate the slab to process it in a single pass
+        // Wait until all Lifetimes have been removed from the slab. 
+        //
+        // Ignored entries will be removed from the Lifecycle slab
+        // by the complete logic called by `tick()`
+        // 
+        // Completed Entries are removed here directly
         let mut id = 0;
         loop {
             if self.ops.lifecycle.is_empty() {
@@ -205,6 +210,7 @@ impl Drop for Driver {
                 }
 
                 Some(_) => {
+                    // Remove Completed entries
                     let _ = self.ops.lifecycle.remove(id);
                     id += 1;
                 }
--- a/src/driver/mod.rs
+++ b/src/driver/mod.rs
@@ -196,14 +196,23 @@ impl Drop for Driver {
             }
             // Cycles are either all ignored or complete
             // If there is at least one Ignored still to process, call wait
-            if let Some(Lifecycle::Ignored(..)) = self.ops.lifecycle.get(id) {
-                // If waiting fails, ignore the error. The wait will be attempted
-                // again on the next loop.
-                let _ = self.wait();
-                self.tick();
-            } else {
-                id += 1;
-            };
+            match self.ops.lifecycle.get(id) {
+                Some(Lifecycle::Ignored(..)) => {
+                    // If waiting fails, ignore the error. The wait will be attempted
+                    // again on the next loop.
+                    let _ = self.wait();
+                    self.tick();
+                }
+
+                Some(_) => {
+                    let _ = self.ops.lifecycle.remove(id);
+                    id += 1;
+                }
+
+                None => {
+                    id += 1;
+                }
+            }
         }
     }
 }
--- a/src/driver/mod.rs
+++ b/src/driver/mod.rs
@@ -201,7 +201,7 @@ impl Drop for Driver {
                 None
             };
 
-            // Remove completed lifecycles from Slab.
+            // Remove completed lifecycles from Slab
             // This prevents worst case quadratic processing
             if let Some(id) = remove {
                 self.ops.lifecycle.remove(id);
--- a/src/driver/mod.rs
+++ b/src/driver/mod.rs
@@ -188,29 +188,22 @@ impl Drop for Driver {
         }
 
         // Wait until all in flight cancellations have stopped
+        // Manually iterate the slab to process it in a single pass
+        let mut id = 0;
         loop {
             if self.ops.lifecycle.is_empty() {
                 break;
             }
             // Cycles are either all ignored or complete
-            let remove = if let (id, Lifecycle::Completed(..)) =
-                &self.ops.lifecycle.iter().next().unwrap()
-            {
-                Some(*id)
-            } else {
-                None
-            };
-
-            // Remove completed lifecycles from Slab
-            // This prevents worst case quadratic processing
-            if let Some(id) = remove {
-                self.ops.lifecycle.remove(id);
-            } else {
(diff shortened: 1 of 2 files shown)


ollie-etl:
Just realised there is nothing stopping a op polling after the runtime drops

src/driver/mod.rs
```diff
@@ -145,10 +145,39 @@ impl Drop for Driver {
...
+                Lifecycle::CompletionList(indices) => {
+                    let mut list = indices.clone().into_list(completions);
+                    if io_uring::cqueue::more(list.peek_end().unwrap().flags) {
+                        // If the completion list is not finished,
+                        // mark for cancellation, and mark the lifecycle as ignored
+                        cancellable_ops.push(id);
+                        *cycle = Lifecycle::Ignored(Box::new(()));
+                        // Dropping the list here deallocates the unconsumed cqe slab entries
+                    } else {
+                        // This op is complete. Replace with a null Completed entry
```
FrankReh, line 169: Interesting the last peeked entry, found not to be More (so it represents finished), could be put into the Completed, result and flags fields, but I also don't see the point. But saving Ok(0) seems odd. I don't have a better alternative though.

ollie-etl:
I've another commit for this, in progress

src/driver/mod.rs
```diff
@@ -145,10 +145,39 @@ impl Drop for Driver {
...
-            // don't cancel completed items
-            if !matches!(cycle, Lifecycle::Completed(_)) {
-                cancellable_ops.push(id);
+
+        let Ops {
+            lifecycle,
+            completions,
+        } = &mut self.ops;
+
+        for (id, cycle) in lifecycle.iter_mut() {
```
FrankReh, line 154: Just like the code it is replacing, this will iterate through the full length of the Vec that has been allocated for the slab, even if the slab could report a 0 length to us. Also I think the slab, when it is dropped, will be drop the vec normally, so that will also be iterating through the entire Vec, looking for Some(T) entries that need to have their drop called. Guess we shouldn't try to change the slab drop, so the Vec will be brought into cache one way or the other anyway.

FrankReh (approved):
I like it. Wish it were testable because it is getting just complicated enough now that we might be overlooking something. The rust compiler and our ability to reason through the logic makes me  pretty sure this does what is intended.

ollie-etl:
@FrankReh : I've made an update optimizing the drop logic

- Now We don't have to allocate vec, we just mark the Op as ignored, and submit completions for those that have been ignored. This is ok, as no poll will occur during Drop, and the slab will be gone once its finished drop.

- During the await loop, remove the completions from the slab, so we don't keep iterating over them. There is some odd structure there to avoid double borrows

ollie-etl:
I do wonder about both this and the original though. How do we know we've awaited all the AsyncCancel's we've pushed? I don't think either the oringal or this handle that properly

FrankReh:
> How do we know we've awaited all the AsyncCancel's we've pushed?

It doesn't need to wait for the AsyncCancel's to be done. It uses a loop at the bottom to wait for the original's to move to complete I believe.  I'll double check that now.

FrankReh:
Odd, this page. My approval shouldn't be counted if another change comes in.

ollie-etl:
We'll there was a bug in the original - things wouldn't move to complete. Submitted may have moved to waiting, and would never be polled. That is fixed here

ollie-etl:
@FrankReh I'm happy this and the original awaited all the ops in the slab. I just don't know it its safe to drop the rings if there are in-fligh AsyncCancels, which we don't account for

src/driver/mod.rs
```diff
@@ -144,40 +144,73 @@ impl Drop for Driver {
...
+            let remove = if let (id, Lifecycle::Completed(..)) =
+                &self.ops.lifecycle.iter().next().unwrap()
+            {
+                Some(*id)
+            } else {
+                None
+            };
+
+            // Remove completed lifecycles from Slab.
+            // This prevents worst case quadtratic processing
```
FrankReh, line 205: I like this logic. Just a typo in what should be quadratic.

FrankReh: Super  picky, while you are at it, maybe no period at end of first sentence? One with a period and one without looks odd.

ollie-etl:
@frank, I'm going to have 1 more edit. I think we can do the wait logic in a single pass of the slab

src/driver/mod.rs
FrankReh: The slab iter processing actually still runs through the whole Vec. And this new logic looks like it could pick up the same id over and over, if our attempt to cancel it doesn't work quickly. I had a comment in the other PR about considering a way of baling out, maybe with an error, or a panic, if progress at canceling had stalled.

Was the goal to avoid a new vec building up with ids to cancel? That had the benefit of just calling cancel once for each but it suffered from having to build up a list first.

I hadn't thought it through fully, but I did wonder why the cancel wasn't issued right there, but maybe having to call tick would mean the slab was changing under us and the borrow checker might even have thought that a bad idea.

FrankReh (changes requested):
Not sure I like this version as much. Do I misunderstand what's going on in that final loop?

ollie-etl:
@FrankReh Cancel is called exactly once for each outstanding op. The logic is pretty much the same as the orignal, only without a vec

- Iterate over the slab, and change each lifecycle to either completed or ignored
- iterate again (which may be more expensive than iterating over the vec, but doesn't allocate), and for each item marked ignore, issue a cancel exactty once
- Iterate a final time, waiting for the slab to be empty, which will occur once all the completions come in.

FrankReh:
> I just don't know it its safe to drop the rings if there are in-fligh AsyncCancels, which we don't account for

I've never read a warning in the liburing stuff about having to wait for quiescence before closing a uring fd. I believe we should proceed like it is safe. If you want, I or one of us can pose the question to the liburing author as an issue on their git repo but I really think the answer is the kernel protects itself from the uring fd being closed and doesn't need the user land to wait for things to be cleaned up.

So to be clear, I think it's a good idea we make the drop wait to see the active operations completed but we don't need to wait to see our async cancels themselves being completed. It's only against the kernel modifying userland data that we are trying to protect ourselves but async cancels don't use userland buffers.

ollie-etl:
@FrankReh The number of defunct processes I've had whilst doing dev this week say otherwise

FrankReh:
> @FrankReh Cancel is called exactly once for each outstanding op. The logic is pretty much the same as the orignal, only without a vec
> 
> * Iterate over the slab, and change each lifecycle to either completed or ignored
> * iterate again (which may be more expensive than iterating over the vec, but doesn't allocate), and for each item marked ignore, issue a cancel exactty once for each item marked Ignored
> * Iterate a final time, waiting for the slab to be empty, which will occur once all the completions come in.

Fair enough. Let me look through it again.

ollie-etl:
Although, maybe AsyncCanels are ok. I don't know. I've been ungracefully exiting with send_zc buffers in flight

src/driver/mod.rs
```diff
@@ -144,40 +144,73 @@ impl Drop for Driver {
...
+            let remove = if let (id, Lifecycle::Completed(..)) =
+                &self.ops.lifecycle.iter().next().unwrap()
+            {
+                Some(*id)
+            } else {
+                None
+            };
+
+            // Remove completed lifecycles from Slab
+            // This prevents worst case quadratic processing
```
FrankReh, line 205: I like the intention but I think it is still quadratic because the lifecycle.iter() will run through its Vec until finding a Some, always starting from the beginning.

And if the first entry found by the lifecycle.iter().next() is always an ignored, we have to do the search again to find its location again, after we've called wait and tick once below.

When things work normally, this will be fine. I like to consider the if the slab had grown to 10_000 simultaneous operations. I don't know if that is small or considered too large by tokio's standards for a single threaded runtime.

But when there is a problem, like I put into the other PR, I wish there was logic to trigger at least a warning to stderr that there appears to be no progress being made in getting something canceled. But Noah didn't think it that important at the time so it's not more important now either.

FrankReh (changes requested):
I'm not convinced we should use a want=1 when waiting. The io-uring crate may provide a wait with timeout however. Or a scheme that just busy waits seems safer.

src/driver/mod.rs
```diff
@@ -144,40 +144,73 @@ impl Drop for Driver {
...
+            if let Lifecycle::Ignored(..) = cycle {
+                unsafe {
+                    while self
+                        .uring
+                        .submission()
+                        .push(&AsyncCancel::new(id as u64).build().user_data(u64::MAX))
+                        .is_err()
+                    {
+                        self.uring
+                            .submit_and_wait(1)
```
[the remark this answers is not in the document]
ollie-etl: There is no change from the current state here. Current calls `self.wait()`. This is defined as
```rust
    fn wait(&self) -> io::Result<usize> {
        self.uring.submit_and_wait(1)
    }
```
I can't call self.wait here due to the borrow checker.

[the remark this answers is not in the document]
FrankReh: You are right. The slipped by me in the review of that PR I guess. I would have deferred to Noah as it was trying to fix a problem and there was room for improvement but I think it would have raised some alarm bells for me too. So this isn't any worse. I agree.

FrankReh:
@ollie-etl I don't think you should be trying to get this mergable today. I'm assuming you want it in the release but if a release goes out this weekend like Noah wondered about, I think we should have a follow up release very soon after that with the PRs that have been consuming so much of our time over the last two weeks. I wouldn't even try to get any of my work into the follow on. It would be purely to catch up to with the current PRs that were ready.

src/driver/mod.rs
ollie-etl: Ok: I have worked out a scheme for single traversal. Will implement

ollie-etl:
@FrankReh I think I've addressed the quadratic behaviour

src/driver/mod.rs
```diff
@@ -144,40 +144,75 @@ impl Drop for Driver {
...
-            .iter()
-            .all(|(_, cycle)| matches!(cycle, Lifecycle::Completed(_)))
-        {
-            // If waiting fails, ignore the error. The wait will be attempted
-            // again on the next loop.
-            let _ = self.wait();
-            self.tick();
+        // Wait until all in flight cancellations have stopped
+        // Manually iterate the slab to process it in a single pass
+        let mut id = 0;
```
FrankReh, line 192: That's a good idea. Manually iterate.

FrankReh: I like it. Now my only concern would be the one I should have had for the earlier version too. ;) I'm going to accept but probably have a look later this afternoon at the lifecycle changes going on here. CompletionList to either Ignored or Completed, and here from Ignored to completed because tick is called, and anything else to being removed.

Nice.

FrankReh (approved):
I approve but I want @Noah-Kennedy to have a look and approve too. This is significant.

src/driver/mod.rs
```diff
@@ -144,40 +144,75 @@ impl Drop for Driver {
         }
 
         // pre-determine what to cancel
-        let mut cancellable_ops = Vec::new();
-        for (id, cycle) in self.ops.lifecycle.iter() {
-            // don't cancel completed items
-            if !matches!(cycle, Lifecycle::Completed(_)) {
-                cancellable_ops.push(id);
+        // All Cancallable Ops are marked ignored
```
FrankReh, line 147: This particular comment is in the wrong  place now. This line could read something like
// Leaving Completed as Completed, CompletionList with final already there also made Completed, turning the rest into Ignored.

src/driver/mod.rs
```diff
@@ -144,40 +144,75 @@ impl Drop for Driver {
...
+                        // This op is complete. Replace with a null Completed entry
+                        *cycle = Lifecycle::Completed(op::CqeResult {
+                            result: Ok(0),
+                            flags: 0,
+                        });
+                    }
+                }
+
+                _ => {
+                    // All other states need cancelling.
```
FrankReh, line 167: Maybe change to
// All other state need cancelling so leave as Ignored.

src/driver/mod.rs
```diff
@@ -144,40 +144,75 @@ impl Drop for Driver {
...
-            .ops
-            .lifecycle
-            .iter()
-            .all(|(_, cycle)| matches!(cycle, Lifecycle::Completed(_)))
-        {
-            // If waiting fails, ignore the error. The wait will be attempted
-            // again on the next loop.
-            let _ = self.wait();
-            self.tick();
+        // Wait until all in flight cancellations have stopped
```
FrankReh, line 190: Rather than say "Wait until all in flight cancellations have stopped"
how about something like
// waiting and looping while a slab index remains Ignored, and removing for anyth other states (which will have to be Completed)

oh wait ... what if the top loop made an entry Ignored, when more CompletionList More entries were going to come in. Then a More does come. Because it is not Final, it stays as Ignored, right? Only the Final cqe makes Ignored go to removed.

FrankReh (changes requested):
I hate to ask for comment fixes when Noah hasn't reviewed yet. But the comments in the right place might make it easier for him to reason through too.

FrankReh:
> Looking at the wrong section

@ollie-etl Yes, easy to look in the wrong section. So for @Noah-Kennedy 's benefit, we no longer think this fixes his earlier version. This just makes it possible to handle CompletionList entries too. But I probably like the last loop here better, even without counting the handling of CompletionList being included here.

src/driver/mod.rs
ollie-etl: Yes, that logic is already in Ignored, in the send_zc branch. Its not in here because we have no consumers, and its specialized by the CqeFuture marker struct

ollie-etl:
@FrankReh Comments addressed

FrankReh (approved):
I would like to see this go in if @Noah-Kennedy finds it passes all the stress tests he had run before. I would expect this to be tweaked down the road, at least as we look for a way that avoids a possible wait hang with want=1. But maybe not before what could be today's release.

Noah-Kennedy (approved):

Noah-Kennedy:
@ollie-etl please run `cargo fmt`, then I will merge.

ollie-etl:
@Noah-Kennedy Done.

ollie-etl:
@Noah-Kennedy I think this needs an approval from you, again it shows as blocked on approval. I guess the old one is stale?

Noah-Kennedy (approved):

PR #165, opened 2022-11-04
Defer work until polled

### LINKED ISSUE #120: possibly deferring work in the creation of an operation
Currently, some actual uring work is done when an *Op\<T>* is created in the *submit_with* function, as seen here.
https://github.com/tokio-rs/tokio-uring/blob/913aa14c567cfae026b7d4f88a5155921cfb9575/src/driver/op.rs#L61-L91

So a uring submission queue entry is created and added to the actual uring that is shared with the kernel when *submit_with* is called. It is only waiting for a response that is accomplished with an *await* on the returned result.

Given the discussion in #115, I wanted to ask if it is really desirable to do this work at creation time?

An alternative would be to return a Future that would first perform the submission and then await the response.

I ask because I can see the design working either way, and this seems as good a place as any to document why the crate design in this regards is what it is.

Having read through what the try_join! and select! macros are doing and seeing how a good crate will have documented what is cancel safe, this seemed like the right time to ask and get the reason down in writing.

I don't see how we can make the uring operations cancel safe anyway, operations where the kernel has begun reading data seem inherently *not* cancel safe if waiting for the result is canceled (in select! terminology). So I think I like the design the way it is, but wondering what others see are the tradeoffs.

ollie-etl:
This PR closes #120. It changes the semantics of all oeprations to do no work unless polled, which is generally the expected behavior of Futures.

As an aside, the API change gets rid of quite a few unwraps around Op creation, which are now mostly infallible, unless they have fallible steps during their creation


ollie-etl:
This isn't ready, but I thought I'd solicit feedback early

ollie-etl:
@Noah-Kennedy I have a question: What are the semantics for Ops polled outside of the runtime? Failure or panic?
In this test https://github.com/tokio-rs/tokio-uring/blob/4b0e07e9cdd9f6cd941ee2c2abedc099961be045/tests/driver.rs#L75 we are to using io_uring resources out of context, and there appears to be significant special casing for this Op.

src/driver/op.rs
```diff
@@ -40,14 +42,18 @@ pub(crate) struct Op<T: 'static, CqeType = SingleCQE> {
 /// A Marker for Ops which expect only a single completion event
 pub(crate) struct SingleCQE;
 
+/// A Marker for Ops which may be submitted when the ring is not present
+/// The only user of this is `Close` currently
+pub(crate) struct Fallible;
```
ollie-etl, line 47: I'm unconvinced this is needed. changin t5ry_submit_with to try_submit_with_immediate or similar, which would submit the operation immediately would be sufficient

ollie-etl:
Summary:

Ops no longer immediately submit. Instead, when a new op is created, the Sqe is calculated as before, and placed in a new Lifecycle variant, `Lifecycle::Pending`. If the op is then dropped, the Lifecycle is removed, but no transaction occurs on the ring. If instead the Op is Polled, the Lifecycle transitions to `Lifecycle::Waiting`, and the sqe placed on the submission ring. From this point onward, things proceed as before.

This has some nice properties. 
  - Op construction is infallible, unless other IO is required in the creation.
  - Errors that occur when pushing the op to the queue can now be handled. They cause the op to complete with io::Error
  - Futures behave as advertised / expected.
  
The question re `Close` remains, although now I've dug into it, i have that question anyway. Its mostly orthogonal to this PR
I also need to do something with the unit tests on Op. Because Polling now does the submit, not creation, its much harder (impossible) to create a fake Op which doesn't submit as a test. Or rather I could, but it'd need its own future impl, which seems a bit pointless.

Noah-Kennedy:
> @Noah-Kennedy I have a question: What are the semantics for Ops polled outside of the runtime? Failure or panic? In this test
> 
> https://github.com/tokio-rs/tokio-uring/blob/4b0e07e9cdd9f6cd941ee2c2abedc099961be045/tests/driver.rs#L75
> we are to using io_uring resources out of context, and there appears to be significant special casing for this Op.
> 
> This PR keeps that special casing (Close is the only caller of try_submit_with). I actually think try_submit_with with should be `try_immediate_submit_with` in this case, to force the submission into a submitted state. This would be the only user of Submitted then also.
> 
> However, I'm unsure why we're going to such lengths to allow Ops to execute outside of a runtime?

I actually think that this special casing of close needs to go. We should really be doing a synchronous close anyways.

That said, this is an issue that needs some discussion. Right now "panic" is fine, but if we add a `Handle` type, this would change, but we can cross that bridge if we get there.

ollie-etl:
All ops are now deferred until Polled. The special case handling for Close is left alone, although `submit_try_with` has been renamed `enqueue`, and the signature changed to seperate Op creation from submission. As a summary of performance, this causes either no change or maybe a slight performance improvement vs master

### Benchmarks

```
Benchmarking no_op/1: Warming up for 3.0000 s
Warning: Unable to complete 100 samples in 5.0s. You may wish to increase target time to 22.1s, or reduce sample count to 20.
no_op/1                 time:   [215.09 ms 215.73 ms 216.45 ms]
                        thrpt:  [462.00 Kelem/s 463.55 Kelem/s 464.91 Kelem/s]
                 change:
                        time:   [-2.6494% -2.0570% -1.4711%] (p = 0.00 < 0.05)
                        thrpt:  [+1.4930% +2.1002% +2.7215%]
                        Performance has improved.
Found 7 outliers among 100 measurements (7.00%)
  4 (4.00%) high mild
  3 (3.00%) high severe
no_op/32                time:   [31.378 ms 31.428 ms 31.481 ms]
                        thrpt:  [3.1765 Melem/s 3.1818 Melem/s 3.1870 Melem/s]
                 change:
                        time:   [-2.1162% -1.9144% -1.7006%] (p = 0.00 < 0.05)
                        thrpt:  [+1.7300% +1.9517% +2.1619%]
                        Performance has improved.
Found 7 outliers among 100 measurements (7.00%)
  6 (6.00%) high mild
  1 (1.00%) high severe
no_op/64                time:   [28.056 ms 28.098 ms 28.146 ms]
                        thrpt:  [3.5529 Melem/s 3.5590 Melem/s 3.5643 Melem/s]
                 change:
                        time:   [-1.4561% -1.2270% -0.9898%] (p = 0.00 < 0.05)
                        thrpt:  [+0.9997% +1.2423% +1.4777%]
                        Change within noise threshold.
Found 8 outliers among 100 measurements (8.00%)
  5 (5.00%) high mild
  3 (3.00%) high severe
no_op/256               time:   [25.777 ms 25.821 ms 25.868 ms]
                        thrpt:  [3.8658 Melem/s 3.8728 Melem/s 3.8795 Melem/s]
                 change:
                        time:   [-0.3218% -0.0816% +0.1423%] (p = 0.51 > 0.05)
                        thrpt:  [-0.1421% +0.0817% +0.3228%]
                        No change in performance detected.
Found 3 outliers among 100 measurements (3.00%)
  3 (3.00%) high mild
```

examples/cat.rs
```diff
@@ -43,6 +43,6 @@ fn main() {
         }
 
         // Include a new line
-        println!("");
```
ollie-etl, line 46: I ran `cargo clippy --fix`. I can revert if preferred

examples/tcp_listener.rs
```diff
@@ -38,7 +38,7 @@ fn main() {
                     }
 
                     let (res, slice) = stream.write_all(buf.slice(..read)).await;
-                    let _ = res.unwrap();
+                    res.unwrap();
```
ollie-etl, line 41: I ran `cargo clippy --fix`. I can revert if preferred

src/driver/accept.rs
```diff
@@ -9,7 +9,7 @@ pub(crate) struct Accept {
 }
 
 impl Op<Accept> {
-    pub(crate) fn accept(fd: &SharedFd) -> io::Result<Op<Accept>> {
```
ollie-etl, line 12: Op creation is infallible

src/driver/close.rs
```diff
@@ -1,29 +1,34 @@
...
-use std::io;
-use std::os::unix::io::RawFd;
+use std::{io, os::unix::io::RawFd};
 
 pub(crate) struct Close {
     fd: RawFd,
 }
 
 impl Op<Close> {
+    /// Close a file descriptor
```
ollie-etl, line 11: Comments unrelated to the PR, but helped me understand the whole Close shutdown saga whilst implementing

src/driver/close.rs
```diff
@@ -1,29 +1,34 @@
...
+    ///
+    /// Close is special, in that it does not wait until it is Polled
+    /// before placing the Op on the submission queue. This is to ensure
+    /// that if the Driver is Dropped with the Operation incomplete,
+    /// the drop logic of the driver will ensure we do not leak resources
     pub(crate) fn close(fd: RawFd) -> io::Result<Op<Close>> {
         use io_uring::{opcode, types};
 
-        Op::try_submit_with(Close { fd }, |close| {
+        let op = Op::submit_with(Close { fd }, |close| {
```
ollie-etl, line 20: Create a Close Op: infallible

src/driver/close.rs
```diff
@@ -1,29 +1,34 @@
...
             opcode::Close::new(types::Fd(close.fd)).build()
-        })
+        });
+
+        op.enqueue()
```
ollie-etl, line 24: immediately enqueue the Op: fallible

src/driver/connect.rs
```diff
@@ -13,7 +13,7 @@ pub(crate) struct Connect {
 
 impl Op<Connect> {
     /// Submit a request to connect.
-    pub(crate) fn connect(fd: &SharedFd, socket_addr: SockAddr) -> io::Result<Op<Connect>> {
+    pub(crate) fn connect(fd: &SharedFd, socket_addr: SockAddr) -> Op<Connect> {
```
ollie-etl, line 16: Op creation is infallible

src/driver/mod.rs
```diff
@@ -258,9 +258,11 @@ impl Ops {
...
-        assert!(self
-            .lifecycle
-            .iter()
-            .all(|(_, cycle)| matches!(cycle, Lifecycle::Completed(_))))
+        for (_, cycle) in self.lifecycle.iter() {
+            match cycle {
+                Lifecycle::Completed(_) | Lifecycle::Pending(_) => {}
+                c => unreachable!("Lifecycle {:?} found during driver drop", c),
+            }
+        }
```
ollie-etl, line 266: Stricvly unrelated, but gives better panic messages when something goes wrong during debug. can revert or PR elsewhere if preferred.

src/driver/op.rs
```diff
@@ -46,7 +48,10 @@ pub(crate) trait Completable {
 }
 
 pub(crate) enum Lifecycle {
-    /// The operation has been submitted to uring and is currently in-flight
+    /// The operation has been created, and will be submitted when polled
+    Pending(Entry),
```
ollie-etl, line 52: The meat of the PR

src/driver/op.rs
```diff
@@ -64,6 +69,19 @@ pub(crate) enum Lifecycle {
...
+        match self {
+            Self::Pending(_) => f.debug_tuple("Pending").finish(),
+            Self::Submitted => f.debug_tuple("Submitted").finish(),
+            Self::Waiting(_) => f.debug_tuple("Waiting").finish(),
+            Self::Ignored(_) => f.debug_tuple("Ignored").finish(),
+            Self::Completed(_) => f.debug_tuple("Completed").finish(),
+            Self::CompletionList(_) => f.debug_tuple("CompletionList").finish(),
+        }
+    }
+}
```
ollie-etl, line 83: Debug only. Again, can revert or PR elsewhere if preferred.

src/driver/op.rs
```diff
@@ -83,57 +101,72 @@ impl From<cqueue::Entry> for CqeResult {
...
     }
 
-    /// Submit an operation to uring.
+    /// Submit an operation to the driver.
     ///
-    /// `state` is stored during the operation tracking any state submitted to
+    /// `data` is stored during the operation tracking any state for submission to
     /// the kernel.
-    pub(super) fn submit_with<F>(data: T, f: F) -> io::Result<Self>
+    pub(super) fn submit_with<F>(mut data: T, f: F) -> Self
```
ollie-etl, line 119: Submission to the driver is infallible. This now places the seq for submission into the Pending state

src/driver/op.rs
```diff
@@ -83,57 +101,72 @@ impl From<cqueue::Entry> for CqeResult {
...
-        if CONTEXT.with(|cx| cx.is_set()) {
-            Op::submit_with(data, f)
-        } else {
-            Err(io::ErrorKind::Other.into())
-        }
+    /// Enqueue an operation on the submission ring, without `await`ing the future
+    ///
+    /// This is useful when any failure in io_uring submission needs
+    /// to be handled at the point of submission.
+    pub(super) fn enqueue(self) -> io::Result<Self> {
```
ollie-etl, line 144: Does nothing if called on an already submitted operation. TODO: improve docs

src/driver/op.rs
```diff
@@ -202,11 +254,11 @@ impl<T, CqeType> Drop for Op<T, CqeType> {
                     }
                 };
 
-                match mem::replace(lifecycle, Lifecycle::Submitted) {
+                match mem::replace(lifecycle, Lifecycle::Ignored(Box::new(()))) {
                     Lifecycle::Submitted | Lifecycle::Waiting(_) => {
                         *lifecycle = Lifecycle::Ignored(Box::new(self.data.take()));
                     }
-                    Lifecycle::Completed(..) => {
+                    Lifecycle::Completed(..) | Lifecycle::Pending(_) => {
```
ollie-etl, line 261: If the op is dropped before it is submitted to the ring, no work on the ring is scheduled.

src/driver/shared_fd.rs
```diff
@@ -74,61 +93,68 @@ impl Inner {
...
+        // we guarantee all in-flight operations have completed. The most
+        // common cause for an error is attempting to close the FD while
+        // off runtime.
+        //
+        // This is done by initializing a `File` with the FD and
+        // dropping it.
+        //
+        // TODO: Should we warn?
+        *state = match CONTEXT.try_with(|cx| cx.is_set()) {
+            Ok(true) => match Op::close(self.fd) {
```
ollie-etl, line 109: As Close is special cased (preserving the current  behavior to immediately submitting to the ring) , we can fallback to synchronous Close as required.

src/driver/shared_fd.rs
```diff
@@ -74,61 +93,68 @@ impl Inner {
...
                     Poll::Pending
                 }
-                State::Closing(op) => {
-                    // Nothing to do if the close opeation failed.
-                    let _ = ready!(Pin::new(op).poll(cx));
+                State::Closing(mut op) => {
+                    let r = ready!(Pin::new(&mut op).poll(cx));
                     *state = State::Closed;
-                    Poll::Ready(())
+                    Poll::Ready(r)
```
ollie-etl, line 152: Propagates any error in Close from io_uring

src/driver/op.rs
```diff
@@ -83,57 +101,72 @@ impl From<cqueue::Entry> for CqeResult {
...
-    pub(super) fn try_submit_with<F>(data: T, f: F) -> io::Result<Self>
-    where
-        F: FnOnce(&mut T) -> squeue::Entry,
-    {
-        if CONTEXT.with(|cx| cx.is_set()) {
-            Op::submit_with(data, f)
-        } else {
-            Err(io::ErrorKind::Other.into())
-        }
+    /// Enqueue an operation on the submission ring, without `await`ing the future
```
FrankReh, line 140: Could end the sentence with a period, to be consistent with other changes to this file.

ollie-etl:
@mzabaluev As you're pretty active in PR's currently, do you have a view on deferred operations?

src/driver/op.rs
```diff
@@ -83,57 +101,72 @@ impl From<cqueue::Entry> for CqeResult {
...
+                        .ops
+                        .get_mut(self.index)
+                        .expect("invalid internal state");
+
+                    match std::mem::replace(lifecycle, Lifecycle::Submitted) {
+                        Lifecycle::Pending(sqe) => {
+                            // Try to push the new operation
+                            while unsafe { driver.uring.submission().push(&sqe).is_err() } {
+                                // Fail with an IoError if encountered
+                                driver.submit()?;
```
mzabaluev, line 158: The lifecycle entry will stay in the `Submitted` state even if this fails?

src/driver/socket.rs
```diff
@@ -40,12 +40,12 @@ impl Socket {
     }
 
     pub(crate) async fn write<T: IoBuf>(&self, buf: T) -> crate::BufResult<usize, T> {
-        let op = Op::write_at(&self.fd, buf, 0).unwrap();
+        let op = Op::write_at(&self.fd, buf, 0);
```
mzabaluev, line 43: I believe the `async` transformation works so that this is not run until the future returned by `write` is polled for the first time. If `Op` creation only happens in such places, we have nothing to worry about?

src/driver/op.rs
oliverbunting: Hmm, yes. Bug. It should be completed

src/driver/socket.rs
oliverbunting: I don't follow, sorry

mzabaluev: Line 43 is not executed synchronously, that happens on the first poll of the future returned by `write`. So this method does not actually do any work until polled.
Is there any instance where `submit_with` is called inside a synchronous fn?

Noah-Kennedy:
@oliverbunting as a result of the convo we had in the issue thread as well as discord, I think it might be better to move this to draft

Noah-Kennedy:
Or potentially even close this since its actually part of a larger discussion we are having.

oliverbunting:
In my experience, PRs marked draft get ignored, and Id like to invite feedback

Issue #150, opened 2022-10-26
Close on Drop Semantics


Noah-Kennedy:
What do we think about our current semantics regarding close-on-drop?

I'm not sure if asynchronously pushing a close op is actually the best approach on drop. It complicates our code a surprising amount, as we need to be able to handle the case here where we drop outside of a runtime context, and thus we need to do a bit of working around this with try_submit and other hacks.

Personally, I think that we should change the semantics to synchronously close the fd (like tokio does) on drop and allow users to explicitly invoke an async close operation if they want to close an fd with uring.

FrankReh:
How async should handle close on drop seems to be one of the questions that comes up again and again, even for tokio.

What drop are you referring to? Don't Op drops have to happen while still in the driver's block_on? At least normally. Do you mean when the user has to Op reference passed out of the context it was created in on purpose?

The uring driver gives us two other options, plus there is the final fallback to using std, as you say. The uring driver allows making a submission with no completion, and since 6.0, allows for a synchronous close.

There is one other complication. We should eventually support uring direct descriptors. When we drop one of those, if the ring exists, we have to use one of the ring's close mechanisms and if the uring has already been closed and there is no driver anymore, we can ignore the direct descriptor value because it won't represent anything and the kernel resource will have already been cleaned up when the uring interface was closed.

If we don't want to try and support pre 6.0 kernels the best we can, in other words, if we want to support 6.0+ kernels, we can say we'll only support direct descriptors for 6.0+ kernels and then we can use the nice synchronous close the uring interface gives us, or even the async close because it could mean one less system call.

Noah-Kennedy:
So, with the off-reference drop, I was referring to the semantics of, say, dropping a TcpListener.

FrankReh:
> So, with the off-runtime drop, I was referring to the semantics of, say, dropping a TcpListener.

We could plan on a sync/std close for something like that, when the fd is a file descriptor and reverting to a uring sync close when it is a uring direct descriptor.

Noah-Kennedy:
We should probably create a dedicated issue for uring direct descriptor support.

PR #168, opened 2022-11-06
Simplify the fallback logic for shared_fs

ollie-etl:
This PR is a light refactoring, which moves the special case handling for `SharedFd` which falls back to synchronous file close if the runtime or driver is unavailable. It moves it from being implemented in Op, and available to all Ops, to within `SharedFd`, to make its purpose more clear.

Relates to #150 (doesn't close)

### DIFF
--- a/src/driver/op.rs
+++ b/src/driver/op.rs
@@ -123,18 +123,6 @@ where
             })
         })
     }
-
-    /// Try submitting an operation to uring
-    pub(super) fn try_submit_with<F>(data: T, f: F) -> io::Result<Self>
-    where
-        F: FnOnce(&mut T) -> squeue::Entry,
-    {
-        if CONTEXT.with(|cx| cx.is_set()) {
-            Op::submit_with(data, f)
-        } else {
-            Err(io::ErrorKind::Other.into())
-        }
-    }
 }
 
 impl<T> Future for Op<T, SingleCQE>
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,7 +1,100 @@
+# 0.4.0 (November 5th, 2022)
+
+### Fixed
+
+- Fix panic in Deref/DerefMut for Slice extending into uninitialized part of the buffer ([#52])
+- docs: all-features = true ([#84])
+- fix fs unit tests to avoid parallelism ([#121])
+- Box the socket address to allow moving the Connect future ([#126])
+- rt: Fix data race ([#146])
+
+### Added
+
+- Implement fs::File::readv_at()/writev_at() ([#87])
+- fs: implement FromRawFd for File ([#89])
+- Implement `AsRawFd` for `TcpStream` ([#94])
+- net: add TcpListener.local_addr method ([#107])
+- net: add TcpStream.write_all ([#111])
+- driver: add Builder API as an option to start ([#113])
+- Socket and TcpStream shutdown ([#124])
+- fs: implement fs::File::from_std ([#131])
+- net: implement FromRawFd for TcpStream ([#132])
+- fs: implement OpenOptionsExt for OpenOptions ([#133])
+- Add NoOp support ([#134])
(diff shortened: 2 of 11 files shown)


ollie-etl:
@Noah-Kennedy @FrankReh This is ready for review

FrankReh:
Does this fix a panic? Is there a reproducer for that panic or is this help for the discussion in #150? Just trying to understand the context when looking at the diffs.

FrankReh:
Totally unrelated to this PR. Is anyone else getting confused by the src/driver directory structure?

There is an driver/op subdirectory but its not used for defining driver operations. The driver operations get their own files under src/driver but there's also bind and  pool and util. When I saw the driver/close.rs mentioned in this PR, my mind when to "oh, the close of the driver". Actually it's the "close" operation the driver supports, having nothing to do with closing the driver.

src/driver/op.rs
```diff
@@ -123,18 +123,6 @@ where
             })
         })
     }
-
```
FrankReh, line 126: Every Op\<T> had come with a *try_submit_with* but the only one that used it was the *close* op? That's interesting.

FrankReh:
Time to throw my hat into the ring about file descriptors.

I don't have a clear picture yet of where it will affect the model that tokio-uring creates, but soon, we should be supporting the uring direct descriptors, available since the 5.15 kernel, that can be created with the openat, openat2, or accept operations and closed with the close operation. I suspect anywhere we are using a fd that we think of as a normal file descriptor, it could also be a direct descriptor that was created by one of the ops mentioned above or that was created with the register_files uring register variant.

The liburing manpage for io_uring_register_files gives a short explanation why an app would want to use direct descriptors. Of course, it boils down to getting better performance.

oliverbunting:
@FrankReh this PR is purely cosmetic, it gives no new functionality, and changes no behaviour. I was trying to understand the special case logic on Close, and realised the logic for it was contained in try_submit_with, which is only called in Close.

I just moved the logic into Close to make it clearer what's going on

FrankReh:
> this PR is purely cosmetic, it gives no new functionality, and changes no behaviour. I was trying to understand the special case logic on Close, and realised the logic for it was contained in try_submit_with, which is only called in Close.
> 
> I just moved the logic into Close to make it clearer what's going on

Thanks. That context helps. I'm for it but let's wait some for @Noah-Kennedy to free up and have a look too.

And regarding the file layout, if you also think something could be done, let's try to convince @Noah-Kennedy. My suggestion would be that we come up with a proposal ahead of time for file movement, but we get all the old PRs taken care of before we make those changes and we warn any fresh PRs coming in that files are about to change - I don't trust git to merge changes well when file names change and content changes. I mean Noah might feel its not worth changing and everything is clear in his head and over time, they will be in ours too, and the churn would be too much work for PRs we didn't resolve ahead of time. Let's see.

Noah-Kennedy:
I've been wanting to restructure the file layout for everything, but am waiting for a break in PRs to quickly move things around.

Noah-Kennedy:
What I'm wondering is if we even want to do this, or if we just want to do a synchronous close on drop. I really don't like the whole deferred cleanup semantics this drop behavior has created.

oliverbunting:
> What I'm wondering is if we even want to do this, or if we just want to do a synchronous close on drop. I really don't like the whole deferred cleanup semantics this drop behavior has created.

I think the alternative is to tie the lifetime of the runtime to that of all Ops, or at least, ops you want to have shutdown semantics

oliverbunting:
Because it's not just Close, it's also multi poll ops

ollie-etl:
@Noah-Kennedy How do you feel about merging this until we get round to working out a better scheme for Close/cancellation semantics?

Noah-Kennedy:
I'd be down with merging while we work out a better scheme.

Noah-Kennedy (approved):

PR #172, opened 2022-11-07
`BoundedBuf` trait to take `.slice` API out of `IoBuf`

mzabaluev:
Introduce new traits, `BoundedBuf` and `BoundedBufMut`, taking the `slice` method and related infrastructure off `IoBuf`.
`Slice` no longer implements `IoBuf`/`IoBufMut`, instead it does `BoundedBuf` and `BoundedBufMut`.
These traits are used as bounds in most generic I/O methods in place of `IoBuf`/`IoBufMut`.

The purpose of this is to get rid of `Slice<Slice<T>>` arising from chained slice calls.

* Any buffer the user has can be defined to be an IoBuf (subject to the safety constraints of IoBuf)
* An IoBuf is by definition a BoundedBuf
* BoundedBuf<IoBuf>.slice(range) -> Slice<IoBuf> (i.e. slicing an IoBuf BoundedBuf produces an IoBuf Slice)
* A Slice<IoBuf> is also a BoundedBuf by definition
* And to void the problem of a Slice<Slice<>> hierarchy building up
  * A BoundedBuf<Slice<IoBuf>>.slice(range) -> Slice<IoBuf> (keeping the slice flat)

All calls that take a BoundedBuf input type can take an IoBuf and a Slice<IoBuf>.

And a brief explanation for the few functions that were split into two layers, a generic public layer and a slice specific private layer:

The purpose of factoring out the bulk of the public method's body into the private method working over Slice is twofold:

* Unify the code path recovering the buffer passed in for the BufResult, rather than have it replicated in multiple return expressions as in https://github.com/tokio-rs/tokio-uring/pull/163;
* Reduce monomorphization bloat, so that most of the machine code to work with buffers passed in as, say, Vec<u8> and Slice<Vec<u8>> is emitted in one instance, rather than replicated instances of exactly the same code.


FrankReh:
@mzabaluev How would you describe the trait relationships that make this work. Such a description could be good to put into the top description of this feature. I'm trying to boil this down into a short description of how this works, why two layers of trait helps, and how it can still be easy for the user to understand and appreciate. The three lines of code required in File.read_exact_at to wrap File.read_exact_slice_at cause me to look for a simple paradigm to have in mind when implementing a feature at this level.

Is it

* Any buffer the user has can be defined to be an IoBuf (subject to the safety constraints of IoBuf)
* An IoBuf is by definition a BoundedBuf
* BoundedBuf\<IoBuf>.slice(range) -> Slice\<IoBuf> (i.e. slicing an IoBuf BoundedBuf produces an IoBuf Slice)
* A Slice\<IoBuf> is also a BoundedBuf by definition
* And to void the problem of a Slice\<Slice\<>> hierarchy building up
  * A BoundedBuf\<Slice\<IoBuf>>.slice(range) -> Slice\<IoBuf> (keeping the slice flat)

So all calls that take a BoundedBuf input type can take an IoBuf and a Slice\<IoBuf>?

Maybe a std::convert::From impl would even help.

I'm not going to try to boil down the Mut versions yet.

FrankReh:
Note: this isn't critical for the PR, but an Into would make some parameter passing cleaner.

I tried to impl the Into trait for a BoundedBufMut into a Slice<IoBufMut> but I
get this compile error. I haven't figured out how to resolve. It's supposed to mean one of the types I'm using isn't from the local trait, but except for the Into trait, the traits are from the tokio-uring crate.

```rust
use std::convert::Into;

impl<T, B> Into<Slice<T>> for B
where
    T: IoBufMut,
    B: BoundedBufMut<BufMut = T>,
 
{
    fn into(self) -> Slice<T> {
        self.slice(..)
    }
}
```

compiles with error:
```bash
  Compiling tokio-uring v0.4.0 (/home/frank/forks/tokio-uring)
error[E0210]: type parameter `B` must be covered by another type when it appears before the first local type (`Slice<T>`)
   --> src/buf/bounded.rs:160:9
    |
160 | impl<T, B> Into<Slice<T>> for B
    |         ^ type parameter `B` must be covered by another type when it appears before the first local type (`Slice<T>`)
    |
    = note: implementing a foreign trait is only possible if at least one of the types for which it is implemented is local, and no uncovered type parameters appear before that first local type
    = note: in this case, 'before' refers to the following order: `impl<..> ForeignTrait<T1, ..., Tn> for T0`, where `T0` is the first and `Tn` is the last

For more information about this error, try `rustc --explain E0210`.
error: could not compile `tokio-uring` due to previous error
```

mzabaluev:
@FrankReh 

> I'm trying to boil this down into a short description of how this works, why two layers of trait helps, and how it can still be easy for the user to understand and appreciate. [...]
> 
> Is it

The bullet points are correct, as is this:

> So all calls that take a `BoundedBuf` input type can take an `IoBuf` and a `Slice<IoBuf>`?

:heavy_check_mark:

> The three lines of code required in File.read_exact_at to wrap File.read_exact_slice_at cause me to look for a simple paradigm to have in mind when implementing a feature at this level.

The purpose of factoring out the bulk of the public method's body into the private method working over `Slice` is twofold:

* Unify the code path recovering the buffer passed in for the `BufResult`, rather than have it replicated in multiple return expressions as in #163;
* Reduce monomorphization bloat, so that most of the machine code to work with buffers passed in as, say, `Vec<u8>` and `Slice<Vec<u8>>` is emitted in one instance, rather than replicated instances of exactly the same code.

> Maybe a `std::convert::From` impl would even help.

I don't see a particular need for it, the I/O methods taking buffer arguments are as polymorphic as they were prior to this change.

mzabaluev:
> 
> ```rust
> impl<T, B> Into<Slice<T>> for B
> where
>     T: IoBufMut,
>     B: BoundedBufMut<BufMut = T>,
> ```

You only really need an `IoBuf` bound on T.
The signature of that implementation is too loose for the orphan rule.
And you should always try to implement `From` rather than `Into`, as the latter then works via its blanket impl.
Try this:

```rust
impl<T: IoBuf> From<T> for Slice<T> {
    // ...
}
```

FrankReh:
@mzabaluev Congrats on fixing my attempt at an Into, or a From also. Even with it implemented, the code in read_exact_at couldn't be changed without making it specific to the IoBufMut case it seemed. (Of course I could have done something else wrong too.) I thought the case where a BoundedBufMut was already a Slice<IoButMut> would be covered because I had read we get the From and Into for free for types that are already the type of the parameter.

No matter. I agree there is no call to create the From implementation because the wrapper function serves an important function of reducing the monomorphism bloat.

src/fs/file.rs
```diff
@@ -338,50 +335,57 @@ impl File {
...
     ///
-    /// [`ErrorKind::Interrupted`]: std::io::ErrorKind::Interrupted
     /// [`ErrorKind::UnexpectedEof`]: std::io::ErrorKind::UnexpectedEof
-    pub async fn read_exact_at<T: IoBufMut>(
+    pub async fn read_exact_at<T>(&self, buf: T, pos: u64) -> crate::BufResult<(), T>
+    where
+        T: BoundedBufMut,
+    {
+        let orig_bounds = buf.bounds();
+        let (res, buf) = self.read_exact_slice_at(buf.slice(..), pos).await;
```
FrankReh, line 344: Instead of calling buf.slice(..) which I think still involves work, even when the compiler knows buf is an IoBufMut, could it call buf.slice_full()? Where slice_full is defined for both IoBuf and Slice<T> trivially? Or does the compiler already deduce all that from the '..' range? Would even the asserts be considered dead code?

FrankReh (changes requested):
Most are just observations, as I become enlightened. This is cool.

But one question about how buf.slice(..) is used in the wrapper functions and one about where the right place to call into_inner() is.

src/buf/bounded.rs
```diff
@@ -0,0 +1,156 @@
...
+/// Because buffers are passed by ownership to the runtime, Rust's slice API
+/// (`&buf[..]`) cannot be used. Instead, `tokio-uring` provides an owned slice
+/// API: [`.slice()`]. The method takes ownership of the buffer and returns a
+/// [`Slice`] value that tracks the requested range.
+///
+/// This trait provides a generic way to use buffers and `Slice` views
+/// into such buffers with `io-uring` operations.
+///
+/// [`.slice()`]: BoundedBuf::slice
+pub trait BoundedBuf: Sized + Unpin + 'static {
```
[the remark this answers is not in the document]
mzabaluev: Yes, surprisingly the `Self: Sized` bound is not needed here. I thought it must be somewhere for the `slice` method definition (like it was in `IoBuf` prior to this change, though that one had a default body), and also for `BoundedBuf::from_buf_bounds`, but the compiler does not require it. I think this is [an explanation](https://stackoverflow.com/a/54468427/2217151).

Since the `BoundedBuf` trait implementations are sealed in this crate, I think it's OK to remove the `Sized` supertrait. Regardless, it cannot be implemented on unsized types, e.g. changing the blanket impl's signature to `impl<T: ?Sized + IoBuf> BoundedBuf for T` will fail.
But I don't think it impedes any known usage either, since `IoBuf` is only implemented on sized types (including `&'static [u8]`).

src/fs/file.rs
```diff
@@ -338,50 +335,57 @@ impl File {
...
-
-        if pos.checked_add(buf_len as u64).is_none() {
+        if pos.checked_add(buf.bytes_total() as u64).is_none() {
             return (
                 Err(io::Error::new(
                     io::ErrorKind::InvalidInput,
                     "buffer too large for file",
                 )),
-                buf,
+                buf.into_inner(),
```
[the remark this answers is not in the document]
mzabaluev: It would result in _cleaner_ code, but I was semi-conscious of the micro-optimization impact: the fat `Slice` struct would be returned by the internal function (or rather, the `Future` impl after the async transform), only to be stripped to the inner buffer by the caller. With unified monomorphization between `T: IoBuf` and `Slice<T>`, 
it is likely that the call is not inlined and the return is done on stack.

src/fs/file.rs
mzabaluev: Indeed, it's possible to add such a method as an optimization, similarly to #53. Some of the asserts in an inlined `buf.slice(..)` call can be eliminated as dead code, but not all I think.

src/buf/bounded.rs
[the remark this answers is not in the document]
FrankReh: Thanks for the link. Very informative.

FrankReh (approved):
LVGTM. Thanks!

FrankReh:
@Noah-Kennedy Can you take another look?

src/buf/bounded.rs
```diff
@@ -0,0 +1,167 @@
...
+    fn stable_ptr(&self) -> *const u8;
+
+    /// Number of initialized bytes available via this view.
+    fn bytes_init(&self) -> usize;
+
+    /// Total size of the view, including uninitialized memory, if any.
+    fn bytes_total(&self) -> usize;
+}
+
+impl<T: IoBuf> BoundedBuf for T {
```
Noah-Kennedy, line 78: Is it possible to do the reverse here, and accepting IoBuf for all of our APIs, with IoBuf being implemented for BoundedBuf implementations?

mzabaluev: In the currently proposed design, `IoBuf` is an open trait which represents whole buffers, so it excludes `Slice`, while `BoundedBuf` is a sealed extension trait, adds slicing and is also implemented for `Slice`. @Noah-Kennedy are you suggesting to rename these traits, or to make more substantial changes?

Noah-Kennedy: Ah, that makes sense.

I was thinking that IoBuf would semantically represent anything which can be used as a buffer for an operation (as is currently the case).

Is there a reason that this can't be the case?

mzabaluev: > I was thinking that IoBuf would semantically represent anything which can be used as a buffer for an operation (as is currently the case).

That role is to be taken by `BoundedBuf`. `IoBuf` is dedicated to representing basic buffers that have stable addresses and length and can maintain the `.bytes_init()` marker, but are otherwise opaque to this crate. `IoBuf` remains an extension point, open for implementation by other crates, while the implementations of `BoundedBuf`  are effectively sealed by tokio-uring.

> Is there a reason that this can't be the case?

We need to be able to pass all types that `slice` can be called on, but for `Slice` this method needs to return `Self`. So we can't have the same trait be an open-ended abstraction for buffers and also generalize the slicing functionality.

Noah-Kennedy: Ah, makes sense then!

Noah-Kennedy (approved):

FrankReh (approved):
What an achievement. Let's see how long this slice implementation lasts. It has certainly been contentious and remains so in some discussions with rust long timers. Big thanks!

PR #173, opened 2022-11-13
rt: use try_with to avoid panic

wllenyj:
The CONTEXT thread_local may have been destructed. Use try will panic. If put Runtime into thread_local, the order of Drop calls is platform-dependent.

Signed-off-by: wanglei01 <wllenyj@linux.alibaba.com>


wllenyj:
Hi, @Noah-Kennedy. 
Any suggestions for this PR? 
Thanks.

wllenyj:
Thanks @FrankReh.

> Does that unit test show a failure without the try change?
> And under what platform did the baseline form cause a problem?

Yes, the GitHub-hosted runners will fail.
You can check https://github.com/wllenyj/test-thread-local/pull/2 for details, I created a repo to test this case.

FrankReh:
@wllenyj Thanks for pointing us to the failing repo. As @Noah-Kennedy is the one who understands all things driver/runtime and CURRENT, I will still defer to him. And thanks for splitting off that other PR. I'll ask for your patience as we work through some long standing PRs that maybe ready to merge and we don't want to ask them for rebases if we don't have to.

wllenyj:
@FrankReh Ok, able to understand. Thanks.

Noah-Kennedy:
I'm not confident that this is enough to make thread-local runtimes work soundly. I'm currently working on something which may be a better solution.

PR #174, opened 2022-11-14
add probe io-uring support

wllenyj:
It is a useful wrapper. Avoid users of the `tokio-uring` to rely on the
`io-uring` library themselves.

Signed-off-by: wanglei01 <wllenyj@linux.alibaba.com>


Cargo.toml
```diff
@@ -18,12 +18,12 @@ keywords = ["async", "fs", "io-uring"]
 
 [dependencies]
 tokio = { version = "1.2", features = ["net", "rt"] }
-scoped-tls = "1.0.0"
 slab = "0.4.2"
 libc = "0.2.80"
 io-uring = { version = "0.5.8", features = ["unstable"] }
 socket2 = { version = "0.4.4", features = ["all"] }
 bytes = { version = "1.0", optional = true }
+once_cell = "1.16.0"
```
FrankReh, line 26: Why use this crate when tokio::sync::OnceCell exists? I haven't looked at the details. You may very well have a good reason.

FrankReh (commented):
I agree letting the user use tokio-uring without having to depend on io-uring is worthwhile.

I am curious at the structure of this implementation though. Why are so many files being touched and why isn't the probe done in a way that the caller could then use the results for checking any other opcodes of interest to them?

I guess exposing uring opcodes is tricky too if io-uring isn't creating an easy way for tokio-uring to expose them.

The APIs for tokio-uring aren't being cast in stone yet so I guess anything is better than nothing and it can always evolve or change before a 1.0.

This PR raises a very good point. Will let @Noah-Kennedy weigh in and decide what to do.

Cargo.toml
```diff
@@ -18,12 +18,12 @@ keywords = ["async", "fs", "io-uring"]
 
 [dependencies]
 tokio = { version = "1.2", features = ["net", "rt"] }
-scoped-tls = "1.0.0"
```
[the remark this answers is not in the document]
wllenyj: Thanks, fixed

Cargo.toml
wllenyj: I just think that the initialization function of OnceCell does not use async, so there is no need to use the async version. Just like Mutex. 
Use tokio::sync::OnceCell can reduce extra dependencies. And it's more tokio. It is also a nice choice. But tokio::sync::OnceCell will add an await dispatch.

I'd love to hear your opinion.

FrankReh: I think I understood the reason, but as you've guessed, adding an await for a one time initialization feature isn't a concern, or not nearly as much as introducing a dependency. Extra dependencies for examples, benchmarks and tests aren't nearly the concern but it is something that everyone using the library would give some brain cycles to. I would say it's better to stay in the tokio ecosystem where possible. Your point that even tokio recommends using the std Mutex when it is appropriate I agree with. We don't have an extensive benchmarking suite yet, it's one of the PRs waiting to be merged; not to make it more extensive yet, but to make what's there more useful.

src/driver/mod.rs
```diff
@@ -42,12 +42,48 @@ mod write;
...
+
+    let mut probe = Probe::new();
+
+    // Check we can register a probe to validate supported operations.
+    if let Err(_) = submitter.register_probe(&mut probe) {
+        return false;
+    }
+
+    // Check IORING_OP_FSYNC is supported
+    if !probe.is_supported(opcode::Fsync::CODE) {
```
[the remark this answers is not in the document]
wllenyj: I just simply check if io-uring is supported, if the kernel supports io-uring, then usually  most of the opcode is supported. It might be more friendly to add an interface here to probe specific opcode.

Cargo.toml
wllenyj: Thanks for your patient reply, I'll fix it.

wllenyj:
> Why are so many files being touched.

The src/driver/mod.rs file is a wrapper around io-uring, all operations on io-uring should be implemented in this module. But is it appropriate to release the exposed interface to Runtime? Or under lib.rs? I don't have any good ideas.

FrankReh:
The work could be in driver but the API could be exposed by the top lib.rs if I understand correctly. But before doing more actual implementation, maybe pass on a thought of what the API for keeping all the probe results would be. Personally, I would be okay if the state weren't kept at all by the library and if the user wants to ask more than once, they have the work done more than once. The less state kept by the library, the easier it is to understand and the fewer opinions it needs to have baked in. There's also the option of a helper module that the driver doesn't rely on but that the user can invoke if it helps standardize a common pattern, like probe results.

It's small, but there probably isn't a need for the library to hold many memory of what the probe results were. Some users will want to keep the data, perhaps for sharing between current threads, perhaps for deciding at runtime what calls to make for a given feature. Others will check once and then not need the data any more.

So I guess I'm leaning towards just supporting the call to probe and returning the results with some kind of boolean tests coded against the probe results, again so the io_uring driver and its opcodes, don't need to be exposed to the user.

wllenyj:
Hi @FrankReh.
I have re-implemented it. Have time to review it? Thanks.

But when I used `tokio::sync::OnceCell`, I got a problem. The user uses the `io_uring_probe` function, there may not be an async context. It may happen before the user creates the `tokyo-uring::Runtime`. If the kernel does not support io-uring, the user may choose to create tokio::Runtime.

Noah-Kennedy:
I'm not willing to accept this PR right now, as we haven't had much discussion on how probing should work semantically. I'd recommend converting this into a draft PR and opening an issue for us to discuss probe semantics. In particular, I'm wondering if probing may be something best done at startup and stored in the runtime, and used to potentially modify runtime behavior or enable/disable certain APIs.

PR #175, opened 2022-11-17
bug: send_zc should be a MultiCQEFuture

ollie-etl:
Somewhere, in the myriad of refactorings that occurred in #123, I dropped the marker type. This leads to wrongly associating the Future for handling SingleCQE types with SendZc, which is a multi completion type.

### DIFF
--- a/src/driver/send_zc.rs
+++ b/src/driver/send_zc.rs
@@ -5,6 +5,7 @@ use crate::{
     BufResult,
 };
 use std::io;
+use crate::driver::op::MultiCQEFuture;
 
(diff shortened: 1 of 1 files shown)


ollie-etl:
@FrankReh @Noah-Kennedy could we get this fixed?

FrankReh (approved):
Would a unit test have caught this, since we would have gotten *more* completions that were unexpected?

ollie-etl:
I just raised an issue for that. THe answer is _yes_, but its unlikely the ci environment can run them currently

PR #180, opened 2022-11-20
chore: refactor driver module

Noah-Kennedy:
This change renames the top-level driver module to "io" and moves the driver and Op to a driver module within the runtime module.


Noah-Kennedy:
This is a precursor to several other PRs. After this, I'm doing some more work on the runtime and putting up my draft with Prepared.

src/runtime/driver/op/mod.rs
```diff
@@ -99,7 +99,7 @@ where
     T: Completable,
 {
     /// Create a new operation
-    fn new(data: T, inner: &mut driver::Driver) -> Self {
+    fn new(data: T, inner: &mut runtime::driver::Driver) -> Self {
```
FrankReh, line 102: Not a comment on this reorg. I wonder why this parameter is named `inner`. Such a common term doesn't mean much to my brain. I would almost rather see index being passed in here, but that's not what this reorg is about and everyone else might like it the way it is anyway. JTOL.

src/fs/directory.rs
```diff
@@ -1,5 +1,4 @@
-use crate::driver::Op;
-
+use crate::runtime::driver::op::Op;
```
FrankReh, line 1: Again, JTOL. I've never noticed this remove_dir function before. Odd we don't have other directory functions yet and odd to me the argument to this isn't like the arg to open in file.rs.

src/io/fsync.rs
```diff
@@ -28,7 +27,7 @@ impl Op<Fsync> {
 impl Completable for Fsync {
     type Output = io::Result<()>;
 
-    fn complete(self, cqe: op::CqeResult) -> Self::Output {
+    fn complete(self, cqe: CqeResult) -> Self::Output {
```
FrankReh, line 30: Nice simplification.

FrankReh:
I think pool.rs should be deleted in the next go-round.

src/io/rename_at.rs
```diff
@@ -20,8 +18,8 @@ impl Op<RenameAt> {
     pub(crate) fn rename_at(from: &Path, to: &Path, flags: u32) -> io::Result<Op<RenameAt>> {
         use io_uring::{opcode, types};
 
-        let from = driver::util::cstr(from)?;
-        let to = driver::util::cstr(to)?;
+        let from = super::util::cstr(from)?;
```
FrankReh, line 21: Maybe six of one, half a dozen of the other. I could see using *crate* instead of *super* here since we'll want to access the util modules from all over, whether or not they are just one crate up and over or not.

FrankReh: Oh man. My comment is wrong. This util isn't at the top level, it's under io/, so *super* is totally correct.

src/io/socket.rs
```diff
@@ -1,6 +1,7 @@
+use crate::runtime::driver::op::Op;
```
FrankReh, line 1: Maybe socket.rs could go somewhere else? It doesn't exactly represent IO like the other files in this directory do. (I think?)

FrankReh (approved):
LGTM. A few comments but nothing important really. So many changes and not a comment in sight for me to nitpick over. 😼

Noah-Kennedy:
> I think pool.rs should be deleted in the next go-round.

Agree.

src/io/socket.rs
Noah-Kennedy: I'd agree, I just need to figure out what to do with it

PR #183, opened 2022-11-24
Example of tcp_listener using fixed buffers

FrankReh:
Just a draft. Meant to show some working code to elicit feedback on what a useful example would be. I particularly don't like that it can't be used as a self contained test. I'm aiready tired of manually starting up netcat clients two or three at a time to verify the results. This one doesn't need it but for newer kernel features, I'm often sending the binary to a test runner and ssh'ing over to it to get other clients started up. Would be nice to have a standard way to control several processes, capturing their respective outputs, sometimes with delays in their startup and input scripts. And having them wrapped in a tmux session with panes so their collection is nicely self contained and can be manually prodded occasionally. Maybe this binary doesn't include the self test option, but another program could serve to start/stop a tmux session and programs within the panes when called with some yaml file.

------

The example is the classic TCP echo server that in this case can serve up to two connections simultaneously. The example creates a fixed buffer register with two entries.

The example does not yet use a write_all function for ensuring everything is written back. It is expected a write_all helper will be provided later for the write_fixed call.

### DIFF
--- a/examples/tcp_listener_fixed_buffers.rs
+++ b/examples/tcp_listener_fixed_buffers.rs
@@ -1,0 +1,95 @@
+// An example of using fixed buffers for reading and writing with TCP streams.
+// A buffer registry size of two is created, to allow a maximum of two simultaneous connections.
+
+use std::{env, iter, net::SocketAddr};
+
+use tokio_uring::buf::fixed::FixedBufRegistry;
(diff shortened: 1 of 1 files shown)


FrankReh:
I'm closing this. It will be part of a write_all PR anyway. We can review there.

PR #184, opened 2022-11-25
write_fixed_all

FrankReh:
Define write_all functions that take fixed buffers.
    
Adds write_fixed_all for TCP and Unix streams.
    
Adds write_fixed_all_at for File.

Also adds an example: tcp_listener_fixed_buffers that uses read_fixed and write_fixed_all.

### DIFF
--- a/examples/tcp_listener_fixed_buffers.rs
+++ b/examples/tcp_listener_fixed_buffers.rs
@@ -1,13 +1,14 @@
-// An example of using fixed buffers for reading and writing with TCP streams.
+// An example of an echo server using fixed buffers for reading and writing TCP streams.
 // A buffer registry size of two is created, to allow a maximum of two simultaneous connections.
 
 use std::{env, iter, net::SocketAddr};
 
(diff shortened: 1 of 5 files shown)


ollie-etl:
@FrankReh This should also be applicable to udp and (I think) also the zero copy variants

FrankReh:
@ollie-etl Turns out the UdpSocket can take a write, as it can be connected separately first, so even write_fixed should be defined for the UdpSocket. Just realized that today.

FrankReh:
@ollie-etl Actually, could you clarify what you meant? An *_all* option doesn't seem necessary for UDP sockets as the writes have to be all or nothing anyway.

oliverbunting:
@FrankReh I meant write_fixed, rather than write_fixed_all.

An explosion in number of methods is occurring. I think the case for a builder API overhaul is growing stronger

oliverbunting:
And realised my initial comment was irrelevant. Please disregard

Noah-Kennedy (approved):

Issue #185, opened 2022-11-26
Allow for thread safe `FixedBuf` handles


Nehliin:
With the introduction of Fixed buffers I was curious if it would be possible to also expose thread safe handles to those buffers? I.e supporting operations on different threads (and runtimes) but still utilizing the same buffer pool. The use case  I have in mind for example is having a separate disk-io thread which would be distinct from the thread reading data from a socket. It would be neat if both could share the same buffer pool by propagating the handles from the socket thread to the disk thread and then use write_fixed there.

After discussions on discord it seems like at a minimum  `register_buffers` and `unregister_buffers` needs to be exposed publicly for it to be possible for end users to create their own buffer registries.

Nehliin:
I realize now that another problem is that all the io objects like TcpStream etc only accepts the `FixedBuf` handle making it impossible to use a custom registry that would with thread safe handles.

mzabaluev:
The reference counting mechanics would need to be redone with something like `Arc<Mutex<FixedBuffers>>`, which would cause unnecessary overhead in the single threaded case.

I think it should be possible to have thread-safe fixed buffer collections alongside the current ones. For operations, it should be possible to generalize with a marker trait.

Nehliin:
Yes! I do not want us to replace the existing ones. Just make the operations more flexible

PR #187, opened 2022-11-26
Expose `buf_index`

Nehliin:
This makes it easier to build abstractions on top of the `FixedBufferRegistry`. I can work around this not being exposed it but it's a bit unnecessary to track the same information in multiple places and I don't see a reason from hiding this information from the users in the first place.

### DIFF
--- a/src/buf/fixed/handle.rs
+++ b/src/buf/fixed/handle.rs
@@ -54,7 +54,7 @@ impl FixedBuf {
             index,
         }
     }
-    
+
     /// Index of the underlying registry buffer
     pub fn buf_index(&self) -> u16 {
         self.index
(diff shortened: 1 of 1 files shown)


mzabaluev (approved):
I don't see a reason why buffer index should not be public.

In a future change to resolve #185 and enable other alternative fixed buffer collections, this method will be provided by a trait, which should be made public to enable off-crate implementations.

FrankReh (approved):
In for a penny, in for a pound.

FrankReh:
Looks like fmt is not happy. Maybe a space or tab left on that blank line above the doc comment?

FrankReh (changes requested):
Seems time to make cargo fmt happy.

FrankReh (approved):

PR #189, opened 2022-11-30
net: UDP socket gets local_addr, read_fixed and write_fixed

FrankReh:
The type UdpSocket gets methods local_addr, read_fixed and write_fixed.

local_addr is useful when binding the socket to 127.0.0.1:0 to be able to report what port the OS assigned.

The read_fixed  and write_fixed are useful, like read and write, when the UDP socket has been connected to an address.


FrankReh:
@Noah-Kennedy Can you review? The additions are innocuous. I'll do a separate sweep of all the code and change functions to use the socket2::SockRef::from call in a separate PR.

Noah-Kennedy (approved):

PR #190, opened 2022-11-30
`FixedBufPool`: a dynamic fixed buffer collection

mzabaluev:
Make the internal infrastructure for fixed buffers more flexible, allowing `FixedBuf` handles to be backed by alternative collections.
Add one new collection as `FixedBufPool`, enabling dynamic retrieval of free buffers.


src/buf/fixed/pool.rs
```diff
@@ -0,0 +1,341 @@
...
+impl Inner {
+    fn new(bufs: impl Iterator<Item = Vec<u8>>) -> Self {
+        let bufs = bufs.take(cmp::min(UIO_MAXIOV as usize, u16::MAX as usize));
+        let (size_hint, _) = bufs.size_hint();
+        let mut iovecs = Vec::with_capacity(size_hint);
+        let mut states = Vec::with_capacity(size_hint);
+        let mut free_buf_head_by_cap = HashMap::new();
+        for (index, mut buf) in bufs.enumerate() {
+            let cap = buf.capacity();
+            let next = free_buf_head_by_cap.insert(cap, index as u16);
```
mzabaluev, line 241: The free list is initially constructed back-to-front, so that for the same capacity bucket, the buffer yielded last will be retrieved first. I did not feel like adding more work or bookkeeping to populate front to back or reverse the list in place, as I don't know of any ill effects this would cause.

src/buf/fixed/pool.rs
```diff
@@ -0,0 +1,341 @@
...
+    }
+
+    fn check_in(&mut self, index: u16, init_len: usize) {
+        let cap = self.iovecs()[index as usize].iov_len;
+        let state = &mut self.states[index as usize];
+        debug_assert!(
+            matches!(state, BufState::CheckedOut),
+            "the buffer must be checked out"
+        );
+        let next = self.free_buf_head_by_cap.insert(cap, index);
```
mzabaluev, line 319: This, however, is beneficial: as recently released buffers will be first to be reused, their data will more likely remain in the caches.

mzabaluev:
Huh, the CI test runner machine does not seem to allow locking memory, uninitialized by the test, for the requested total capacity of 8 kB. The test succeeds on my machine.
Elsewhere, the `FixedBufRegistry::new` test succeeds with fully initialized buffers.

mzabaluev:
Reproduced the failure locally by setting `ulimit -l 1`. Will try to make the tests push less against the resource limits.

FrankReh:
This is substantial.

I'll take a look tomorrow. Today I saw an Inner that is used for the old Registry and the new Pool. Interesting that neither type needed extra fields. Wonder if Inner should get its own file in buf/fixed/. I thought I saw a buf.len() being passed to a drop, but then the code refreshed. Not sure if it's still there. The buf.len() I guess is fixed, there is no difference between the len and the capacity.

Of course, I'm always looking for a little design in the source comments. To explain the structure and the reason for the structure. Maybe it's there. I'll look with fresh eyes tomorrow. If CI is failing because of memory constraints, are the test buffer lengths being made short?

FrankReh:
Started looking at it. Good thing I realized the Inner for Registry and Pool are different. Funny how our assumptions we don't even know we're making can color what the code looks like.

FrankReh:
So general question. Why should the pool be part of the crate and not provided as an extra?

Reasons I might guess are the API surface is so small, it's a no-brainer, or provides a valuable base that some part of the crate should be relying on anyway,

mzabaluev:
> So general question. Why should the pool be part of the crate and not provided as an extra?

I think it makes sense to include at least a few simple collections that cover the most important use cases.
I started out adding the free list into `FixedBufRegistry`, but realized that this will impose performance and bookkeeping impact on both cases, compared to the separate collections dedicated to each purpose.

I'm totally on board with making the fixed buffers API public to allow off-crate implementations. This PR is one step towards that goal, and I'm going to do more while working on #185.

src/buf/fixed/pool.rs
```diff
@@ -0,0 +1,353 @@
...
+            let registry = Rc::clone(&self.inner);
+            // Safety: the validity of buffer data is ensured by
+            // Inner::try_next
+            unsafe { FixedBuf::new(registry, data) }
+        })
+    }
+}
+
+// Internal state shared by FixedBufPool and FixedBuf handles.
+pub(crate) struct Inner {
```
mzabaluev, line 215: I just realized I did not make the `Inner` structs module-private, which may cause confusion. This is just a leftover from earlier code. But they are not going to stay private (or named `Inner`) for long, as they will be shared with the upcoming `Sync + Send` fixed buffer implementations.

.github/workflows/ci.yml
```diff
@@ -54,6 +54,8 @@ jobs:
   test:
     runs-on: ubuntu-latest
     steps:
+      - name: Print locked memory limits
+        run: ulimit -l; ulimit -H -l
```
FrankReh, line 58: Can this change be explained? Why are soft and hard limits printed?

FrankReh (changes requested):
Still not convinced the Pool should be part of the src/buf/fixed proper. It's not a deal breaker but I like having the bare bones Registry in there. If @Noah-Kennedy is also on the fence and the author feels strongly, then that's enough for me. But otherwise, this would seem a good time to branch into an extras folder. It would give folks a clearer picture of what they could bring to the table themselves, kind of a template. But I guess the trait is enough to do that. So I don't know.

I would prefer to see the two Inners in their own files, so that would mean different names already I guess. Separate files also makes the separation of duties easier on the mind, and like I said in the other chat, it's easier on the mind, and using Vec and RawVec as the example. Again, I don't feel super strongly about it.

I shouldn't leave this without saying the code is super impressive.

.github/workflows/ci.yml
mzabaluev: I was troubleshooting some tests that were failing in CI, where both limits turn out to be set to 64 k (that's only 16 pages). I printed the hard limit on the off chance that we could raise the soft limit to something bigger than it's set to in the runner shell.

This limit gives you the total size of memory pages a process is allowed to lock, which the memory referenced in `register_buffers` is counted to. I've adjusted the tests to not require more than a fraction of that limit, and to not use very small buffers that may be laid out on many more pages than their total size suggests due to fragmentation. The printouts can be removed now.

src/buf/fixed/buffers.rs
```diff
@@ -1,113 +1,12 @@
...
-                // as checked by Vec::get_mut above, called on the array of
-                // states that has the same length.
-                let iovec = unsafe { iovecs_ptr.as_ptr().add(index).read() };
-                Some((iovec, init_len))
-            }
-            BufState::CheckedOut => None,
-        })
-    }
+// Abstracts management of fixed buffers in a buffer registry.
+pub(crate) trait FixedBuffers {
```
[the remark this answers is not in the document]
mzabaluev: No, that's exactly the trait to expose, once other pieces that are needed for custom implementations fall into place.

src/buf/fixed/handle.rs
```diff
@@ -8,45 +8,54 @@ use std::mem::ManuallyDrop;
...
 
 impl FixedBuf {
-    pub(super) unsafe fn new(
-        registry: Rc<RefCell<FixedBuffers>>,
-        iovec: iovec,
-        init_len: usize,
-        index: u16,
-    ) -> Self {
+    // Safety: Validity constrants must apply to CheckedOutBuf members:
+    // - iovec must refer to an array allocated by Vec<u8>;
```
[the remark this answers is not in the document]
mzabaluev: You are right, there is nothing in the `FixedBuf` handle itself that needs this contract.

> Probably mmap would usually be used in Anonymous mode, but there should be no reason why the buffer space couldn't be backed by a file either.

Currently an anonymous mapping is required, but the kernel will tell you about that with an error code.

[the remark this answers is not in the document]
mzabaluev: > there is nothing in the FixedBuf handle itself that needs this contract.

...um, except for the letter of the safety constraints on `Vec::from_raw_parts`, but I can replace the `ManuallyDrop<Vec<u8>>` member with raw array data. Good point looking forward to custom mmap-backed collections.

src/buf/fixed/handle.rs
```diff
@@ -8,45 +8,54 @@ use std::mem::ManuallyDrop;
...
 }
 
 impl FixedBuf {
-    pub(super) unsafe fn new(
-        registry: Rc<RefCell<FixedBuffers>>,
-        iovec: iovec,
-        init_len: usize,
-        index: u16,
-    ) -> Self {
+    // Safety: Validity constrants must apply to CheckedOutBuf members:
```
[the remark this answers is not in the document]
mzabaluev: https://github.com/tokio-rs/tokio-uring/pull/190/commits/d470b888e96ebbe2b2754194054f7822d012c87c

.github/workflows/ci.yml
mzabaluev: Removed in https://github.com/tokio-rs/tokio-uring/pull/190/commits/91ce0dcbbf15892cac4dd903e7e3d0009f5c8fa4

src/buf/fixed/handle.rs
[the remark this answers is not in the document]
mzabaluev: Meanwhile, taking in `Vec` leaves these basic collections somewhat usable as a proxy for mmap, as this is what the default allocator goes to for larger allocations. There is still the initialization overhead of many successive allocations and you may want to fine-tune your mmap, use hugepages, etc. So I'll go ahead and dismantle the virtual `Vec`.

mzabaluev:
> Still not convinced the Pool should be part of the src/buf/fixed proper. It's not a deal breaker but I like having the bare bones Registry in there.

¿Por qué no los dos?
It's not possible to reuse the Registry and build free lists on top of it without some parallel bookkeeping, or cracking into some heady inner generics, and you need to be mindful of not exposing `check_out` while providing `try_next` as your primary API, or otherwise you'd need the lists to be doubly linked. So I decided that the pool use case warrants its own internal implementation.

src/buf/fixed/buffers.rs
```diff
@@ -1,113 +1,12 @@
...
-                    let cap = iovec.iov_len;
-                    let v = unsafe { Vec::from_raw_parts(ptr, init_len, cap) };
-                    mem::drop(v);
-                }
-                BufState::CheckedOut => unreachable!("all buffers must be checked in"),
-            }
-        }
-    }
+    // state, otherwise this function will panic.
+    fn check_in(&mut self, buf_index: u16, init_len: usize);
```
mzabaluev, line 11: I wonder if this method needs to be declared unsafe.

Surely the implementation does not need to do anything unsafe, but on the next check out, the buffer's initialized length will be set to the provided value.
So the caller of this method needs a speed bump like the one you get with `Vec::set_len`.

FrankReh: That sounds good to me. I like speed bumps.

FrankReh:
> It's not possible to reuse the Registry and build free lists on top of it

I wasn't clear. I didn't mean Registry could be built upon by others. I meant Registry was a good example to leave in because it is pretty bare bones, without a lot of opinions. Better sets, like the new Pool, will usually be preferred and they require their own raw parts.

Would you mind asking Noah about this? If he doesn't care, I won't either. It's a good tool (FixedBufPool) and I don't want to slow you down. I guess I'm more afraid of triggering a proliferation of alternate implementations within `tokio-uring` and this feels like the first, at least the first of this caliber. (We have many ways to write and read.)

FrankReh (changes requested):
Looking very good. I'm looking forward to getting this in.

src/buf/fixed/handle.rs
```diff
@@ -8,45 +8,54 @@ use std::mem::ManuallyDrop;
...
 /// application code or owned by an I/O operation at any given time,
 /// thus avoiding data races between `io-uring` operations in flight and
 /// the application accessing buffer data.
 ///
 /// [`FixedBufRegistry`]: super::FixedBufRegistry
+/// [`FixedBufPool`]:     super::FixedBufPool
 ///
 pub struct FixedBuf {
-    registry: Rc<RefCell<FixedBuffers>>,
+    registry: Rc<RefCell<dyn FixedBuffers>>,
```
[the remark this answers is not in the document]
mzabaluev: No, it does not introduce extra allocations, but the trait object type is unsized, so the `Rc` turns into a fat pointer (there needs to be a vtable to dispatch to the trait methods and call `drop`). `RefCell` is optionally sized and implements `CoerceUnsized`, so it propagates the unsizing.

src/buf/fixed/pool.rs
```diff
@@ -0,0 +1,355 @@
...
+/// use std::mem;
+///
+/// # #[allow(non_snake_case)]
+/// # fn main() -> Result<(), std::io::Error> {
+/// # let (memlock_limit, _) = rlimit::Resource::MEMLOCK.get()?;
+/// # let BUF_SIZE_LARGE = memlock_limit as usize / 8;
+/// # let BUF_SIZE_SMALL = memlock_limit as usize / 16;
+/// let pool = FixedBufPool::new(
+///     iter::once(Vec::with_capacity(BUF_SIZE_LARGE))
+///         .chain(iter::repeat_with(|| Vec::with_capacity(BUF_SIZE_SMALL)).take(2))
```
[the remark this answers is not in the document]
mzabaluev: True. This pool is confirmed to be exhausted later on by the test.

src/buf/fixed/pool.rs
```diff
@@ -0,0 +1,355 @@
...
+            raw_bufs,
+            states,
+            orig_cap,
+            free_buf_head_by_cap,
+        }
+    }
+
+    // If the free buffer list for this capacity is not empty, checks out the first buffer
+    // from the list and returns its data. Otherwise, returns None.
+    fn try_next(&mut self, cap: usize) -> Option<CheckedOutBuf> {
```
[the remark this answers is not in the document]
mzabaluev: I got annoyed passing 3-ary tuples around. But this will make a nice public API when we open it up.

src/buf/fixed/registry.rs
```diff
@@ -73,7 +125,7 @@ impl FixedBufRegistry {
     /// of the `tokio-uring` runtime this call is made in, the function returns
     /// an error.
     pub fn register(&self) -> io::Result<()> {
-        crate::io::register_buffers(&self.inner)
+        crate::io::register_buffers(Rc::clone(&self.inner) as _)
```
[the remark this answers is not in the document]
mzabaluev: `<Rc as Clone>::clone` is shallow, it only increases the reference count. The method is invoked in type-qualified way to highlight this. The `as _` part is needed to prevent type inference from choosing the wrong clonee type, as references to different types of `Rc` cannot be coerced to one another, but the `Rc` instances themselves do coerce to unsized. This quirk is explained [here](https://stackoverflow.com/a/63894064/2217151).

src/io/register_buffers.rs
```diff
@@ -5,24 +5,24 @@ use std::cell::RefCell;
 use std::io;
 use std::rc::Rc;
 
-pub(crate) fn register_buffers(buffers: &Rc<RefCell<FixedBuffers>>) -> io::Result<()> {
+pub(crate) fn register_buffers(buffers: Rc<RefCell<dyn FixedBuffers>>) -> io::Result<()> {
```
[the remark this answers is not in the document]
mzabaluev: As explained in the previous comment, we cannot borrow a reference to a fat `Rc<RefCell<dyn FixedBuffers>>` pointer, without that fat pointer being instantiated somewhere, and the `Rc<RefCell<Inner>>` member of the collection struct is not it. We have to first clone and coerce it to a temporary fat pointer, and since we have done that, why not spare an extra indirection and pass the temporary `Rc` in by value?

Cargo.toml
```diff
@@ -32,6 +32,7 @@ futures = "0.3.25"
 criterion = "0.4.0"
 # we use joinset in our tests
 tokio = "1.21.0"
+rlimit = "0.8.3"
```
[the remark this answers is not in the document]
mzabaluev: https://github.com/tokio-rs/tokio-uring/pull/190/commits/002517aba99ed545ec5ffb7dbbabf92ad9429626

src/buf/fixed/pool.rs
```diff
@@ -323,15 +348,8 @@ impl FixedBuffers for Inner {
...
-    fn check_in(&mut self, index: u16, init_len: usize) {
-        let cap = self.iovecs()[index as usize].iov_len;
-        let state = &mut self.states[index as usize];
-        debug_assert!(
-            matches!(state, BufState::CheckedOut),
-            "the buffer must be checked out"
-        );
-        let next = self.free_buf_head_by_cap.insert(cap, index);
-        *state = BufState::Free { init_len, next };
+    unsafe fn check_in(&mut self, index: u16, init_len: usize) {
```
FrankReh, line 351: No SAFETY comment?

src/buf/fixed/registry.rs
```diff
@@ -264,6 +257,19 @@ impl FixedBuffers for Inner {
...
 
+impl FixedBuffers for Inner {
+    fn iovecs(&self) -> &[iovec] {
+        // Safety: the raw_bufs pointer is valid for the lifetime of self,
+        // the length of the states array is also the length of buffers array
+        // by construction.
+        unsafe { slice::from_raw_parts(self.raw_bufs.as_ptr(), self.states.len()) }
+    }
+
+    unsafe fn check_in(&mut self, index: u16, init_len: usize) {
```
FrankReh, line 268: Same. No SAFETY comment?

FrankReh (changes requested):
Looks good. Very good. Maybe just describe for me about those two unsafe methods. Then I think this is ready.

FrankReh (approved):

FrankReh:
Thanks for reworking with nix.

src/buf/fixed/pool.rs
mzabaluev: The implementation is delegated to an all-safe method, no additional concerns here.
The caller's safety requirements are documented on the trait.

src/buf/fixed/registry.rs
mzabaluev: Ditto.

FrankReh:
@mzabaluev I'm getting buffers out with init_len set to zero. Does that make sense?

I followed the example of using iterators that use with_capacity, which creates a zero length Vec by design.
Where should the length be filled in? Is there another call that would zero fill efficiently?

FrankReh:
And maybe this Pool should have come with a rezero flag so when buffers are reclaimed, they can optionally be reinitialized to zero. Well, the Registry doesn't come with it either. I don't know, I feel like we should be zero filling buffers that are allocated from these pools, but I'm not sure. Maybe there should be two calls for getting. We're not getting the benefits of filled vs init markers in the buffers like Nick is pushing in his async IO traits work.

mzabaluev:
@FrankReh The buffers you get initially are the ones you put in. It's normal to have buffers at zero initialized data length until they are initialized with something. `init_len` tracks the initialized amount not because we care about having precisely this much data in the buffer after any one op, but so that there is no UB with using uninitialized data. If not for that, we'd have to zero-fill every buffer out right, which is the pessimized old way to achieve safety with reads.

Currently in the crate, there is nothing better than `IoBufMut` for writing bytes into a partially initialized buffer and advancing the initialized water mark. The read operations do it for you - the kernel writes into the buffer, the library's op implementation calls `set_init` to make the data available safely. But to prepare a possibly uninitialized buffer for writing, the application has to do the same on its side. There should be utilities on the `BoundedBufMut` trait for safely putting byte slices, integer values in specified byte order, etc. into the buffer. Maybe `std::fmt::Write` impls as well so you can format strings directly in a buffer.

> And maybe this Pool should have come with a rezero flag so when buffers are reclaimed, they can optionally be reinitialized to zero

Zero-filling is unnecessary. It shouldn't be a problem to overwrite existing data in subsequent reads into the buffer. You get the amount of bytes read by the last operation so you use a slice (a built-in, `Deref`-provided one) with that many bytes to get at the data of the latest read. In async-io-traits, I see an effort to put the read cursor into the returned buffer, but it might just complicate things too much. We also went through similar ideas in https://github.com/tokio-rs/tokio-uring/pull/163#discussion_r1013187082.

FrankReh:
How does a user of the FixedBuf write data to, extending its length if the first operation with a fixed buffer is to write it to a file descriptor? Is there a method I'm overlooking?

FrankReh:
I think this is a bit of the kind of opinion we'll want to keep out of the mainline crate. The opinion, in this case, mine.

I'm all for a fast implementation and providing a crate that is safe to use, by rust/UB standards. But I would want a pool of buffers that zeros out the data, either when the message gets put back into the pool or when it is pulled out. Too many smart people dissecting binaries and finding flaws in their design and then using those flaws to spill private data from another account. Go provides that level of security by forcing all data to be zeroed when it is allocated from the heap. They don't try to save time by tracking what bytes are initialized and what are not. They have assembler that is quite fast at zeroing memory.

I'll hope the LLVM compiler has a similar zeroing assembler routine it can call upon when it finds a suitable loop that writes 0 from beginning to end.

But a pool with the opinion that it doesn't retain packet data once the packet is dropped should be kept out of the main crate.

mzabaluev:
There is often a tradeoff between security and performance, and I think a software component using fixed buffers with io-uring would lean heavily towards the latter, internally. I envision that the applications would not let `FixedBuf` handles get around too far from the few I/O loops that use them. That said, there may be a case for a fill marker separate from the initialized bytes marker similar to one maintained by the internal buffer of a `std::io::BufReader`, so that it does not expose data from past reads. But this perhaps should be rearchitected on the general API level, because the use pattern currently provided by `IoBufMut` does not care about pre-existing data in the buffer. If you reuse a `Vec<u8>` as your buffer in tokio-uring, you get the same problem.

mzabaluev:
> How does a user of the FixedBuf write data to, extending its length if the first operation with a fixed buffer is to write it to a file descriptor? Is there a method I'm overlooking?

No, as said [above](#issuecomment-1336285856), currently there isn't an easily usable method. I have only prototyped a function that does it [in a test](https://github.com/tokio-rs/tokio-uring/blob/3b2174ea8e48d898ba6f8aede11717559e5d73bb/tests/fs_file.rs#L316). Perhaps it's high time to add it as a method to `BoundedBufMut`.

PR #192, opened 2022-12-02
Add TcpStream::set_nodelay

kaimast:
Exposes `net::TcpStream::set_nodelay` like in `tokio::net` and `std::net`.


src/io/socket.rs
```diff
@@ -206,6 +206,18 @@ impl Socket {
...
         result
     }
+
+    pub fn set_nodelay(&self, nodelay: bool) -> io::Result<()> {
+        use std::os::unix::io::FromRawFd;
+
+        let fd = self.as_raw_fd();
+
+        // SAFETY: See note for Self::shutdown
+        let s = unsafe { socket2::Socket::from_raw_fd(fd) };
```
Noah-Kennedy, line 216: Use SockRef

Noah-Kennedy (changes requested):

src/io/socket.rs
kaimast: Done! You could probably make the same change for `shutdown ` as well.

Noah-Kennedy: Yeah, I forgot it was done like that.

Noah-Kennedy (approved):

PR #193, opened 2022-12-02
Add create_dir 

### LINKED ISSUE #48: Add missing filesystem operations

kaimast:
This adds a `create_dir` function call to `tokio_uring::fs` and partially fixes #48.

I also renamed `tests/directory.rs` to `tests/fs_directory.rs` so the name is in sync with that of `tests/fs_file.rs`.


src/io/mkdir_at.rs
```diff
@@ -0,0 +1,40 @@
...
+        })
+    }
+}
+
+impl Completable for Mkdir {
+    type Output = io::Result<()>;
+
+    fn complete(self, cqe: CqeResult) -> Self::Output {
+        cqe.result?;
+        Ok(())
```
FrankReh, line 38: Also, like unlink_at.rs,
```rust
cqe.result.map(|_| ())
```

FrankReh (changes requested):
Nice. Just minor stuff.

FrankReh:
@kaimast Would it be alright with you if I made the two tweaks and then merged?

kaimast:
Yes, of course.

kaimast:
I made the changes and also resolved conflicts with more recent changes in master.

FrankReh (approved):
Looks good. Thanks for getting to this.

PR #196, opened 2022-12-04
driver: refactor the driver to be stored in a shared set of handles

Noah-Kennedy:
This opens up a lot of further cleanup opportunities, especially around shutdown logic. It also opens up the opportunity for us to eventually provide a public Handle API.

I am trying to make it so that we don't need to orchestrate our entire shutdown logic around the thread locals rather than purely the CONTEXT thread local, which is slightly broken to begin with (I believe that there are currently issues around creating two runtimes at the same time).

Now the CONTEXT behaves more like in tokio. When you enter a runtime context via block_on, we set the context. When we leave that call, we unset the context.

This also allows us to have a bit more flexibility with things like sending ops between multiple runtimes as well, so you can now in theory await ops created from another runtime.

Part of the goal of this is to add new options around things like fixed or provided buffers. This potentially allows us to work towards things like transferring buffers between uring runtimes.

More importantly however, it is a crucial first step towards having a handle to a uring runtime.

More refactoring to come.


Noah-Kennedy:
I still need to fix the tests.

Noah-Kennedy:
Might be a good idea to hide whitespace when reviewing.

src/io/accept.rs
```diff
@@ -17,21 +18,23 @@ impl Op<Accept> {
...
-                opcode::Accept::new(
-                    types::Fd(accept.fd.raw_fd()),
-                    &mut accept.socketaddr.0 as *mut _ as *mut _,
-                    &mut accept.socketaddr.1,
-                )
-                .flags(libc::O_CLOEXEC)
-                .build()
-            },
-        )
+        CONTEXT.with(|x| {
```
Noah-Kennedy, line 21: For many of these op constructors, I plan to have the weak handle in the socket object.

src/runtime/driver/handle.rs
```diff
@@ -0,0 +1,272 @@
...
+
+    pub(crate) fn tick(&self) {
+        self.inner.borrow_mut().tick()
+    }
+
+    pub(crate) fn flush(&self) -> io::Result<usize> {
+        self.inner.borrow_mut().uring.submit()
+    }
+
+    pub(crate) fn register_buffers(
```
Noah-Kennedy, line 37: I'm going to move a bunch of these methods into the Driver and do delegated calls here, but in another PR.

src/runtime/driver/mod.rs
```diff
@@ -52,7 +52,7 @@ impl Driver {
 
     // only used in tests rn
     #[allow(unused)]
-    fn num_operations(&self) -> usize {
+    pub(super) fn num_operations(&self) -> usize {
```
FrankReh, line 55: Would be nice if things like this were exposed to the tracing tool, tokio console.

src/runtime/driver/handle.rs
```diff
@@ -0,0 +1,272 @@
+use io_uring::squeue;
```
FrankReh, line 1: Is there a comment block somewhere describing the design? Both for reviewers now but more importantly in the long run for readers and developers in the future?

FrankReh, line 1: Maybe put cqueue up here too, for the symmetry.

src/runtime/driver/handle.rs
```diff
@@ -0,0 +1,272 @@
...
+                    driver.ops.remove(op.index);
+                }
+            }
+            Lifecycle::Ignored(..) => unreachable!(),
+        }
+    }
+}
+
+impl WeakHandle {
+    pub(crate) fn upgrade(&self) -> Option<Handle> {
```
FrankReh, line 245: Output lifetime not tied to &self's lifetime. By design?

FrankReh: Imagine I'll run into reasons soon enough when I looking at more code. Ignore my question.

src/runtime/driver/mod.rs
Noah-Kennedy: I think that this is a good idea to expose eventually.

src/runtime/driver/handle.rs
Noah-Kennedy: Hmm that is a good idea. I'll add it in the Handle file.

Noah-Kennedy: Oh I missed this

src/runtime/driver/handle.rs
Noah-Kennedy: This retrieves a Handle from WeakHandle, and basically copies from the Rc/Arc API

src/runtime/driver/handle.rs
Noah-Kennedy: Done.

Noah-Kennedy: Done.

Noah-Kennedy:
I've fixed the tests and marked this and removed the draft label.

FrankReh:
Not as part of this PR, but perhaps a quick follow-on. What would you think about moving the definitions from all the mod.rs files into their own files? So the mod files are just lines of `use` and `mod` themselves? Like most of runtime/mod.rs goes into runtime/runtime.rs and most of runtime/driver/mod.rs goes in runtime/driver/driver.rs and so on.

FrankReh:
Please remind me, why does the `tokio_uring::spawn` call `tokio::task::spawn_local` and not use the `Runtime`'s `local`? https://github.com/tokio-rs/tokio-uring/blob/ca8887c2faf0499a907bcb3b7684693a1ffa807a/src/runtime/mod.rs#L55-L57

Noah-Kennedy:
> Not as part of this PR, but perhaps a quick follow-on. What would you think about moving the definitions from all the mod.rs files into their own files? So the mod files are just lines of `use` and `mod` themselves? Like most of runtime/mod.rs goes into runtime/runtime.rs and most of runtime/driver/mod.rs goes in runtime/driver/driver.rs and so on.

This is a really good idea.

Noah-Kennedy:
> Please remind me, why does the `tokio_uring::spawn` call `tokio::task::spawn_local` and not use the `Runtime`'s `local`?
> 
> https://github.com/tokio-rs/tokio-uring/blob/ca8887c2faf0499a907bcb3b7684693a1ffa807a/src/runtime/mod.rs#L55-L57

Because this spawn method is static and thus happens via the thread local context in tokio.

FrankReh:
> Because this spawn method is static and thus happens via the thread local context in tokio.

But Tokio still adds it to this crate's runtime's localset because that localset is in some sense active, because it all happens in the context of calling rt.block_on(self.local.run_until(..))?

src/runtime/driver/handle.rs
```diff
@@ -0,0 +1,286 @@
...
+        // Push the new operation
+        while unsafe { driver.uring.submission().push(&sqe).is_err() } {
+            // If the submission queue is full, flush it to the kernel
+            driver.submit()?;
+        }
+
+        Ok(op)
+    }
+
+    pub(crate) fn poll_op<T>(&self, op: &mut Op<T>, cx: &mut Context<'_>) -> Poll<T::Output>
```
FrankReh, line 112: I don't see the benefit of moving the two big Future functionalities into the new handle.rs file but maybe I just haven't gotten used to it yet. The idea of a Handle is going to get muddy if all the little details have to be put into this file though (it seems to me). Please take all this with a grain of salt, I haven't tried to get it to compile any other way.

Noah-Kennedy:
Entering a LocalSet context (e.g run_until) sets thread locals within tokio to denote the "current" runtime, kinda like we now are with the drivers.

src/runtime/context.rs
```diff
@@ -1,30 +1,27 @@
-use crate::runtime::driver::Driver;
-use crate::util::PhantomUnsendUnsync;
+use crate::runtime::driver;
```
FrankReh, line 1: Maybe a good time to add a few file comments about this RuntimeContext. How it sort of sits behind the scenes and is ready to potentially give access to the Runtime and gets installed by the Runtime's block_on pub(crate) routine. (I think block_on is currently pub though.)

But whatever you think. You're allowed to say, you don't think it's necessary. I understand sometimes I look for too much help from comments when another window and another minute would answer the question for me.

src/runtime/context.rs
```diff
@@ -1,30 +1,27 @@
...
+use crate::runtime::driver::Handle;
 use std::cell::RefCell;
-use std::marker::PhantomData;
 
 /// Owns the driver and resides in thread-local storage.
 pub(crate) struct RuntimeContext {
-    driver: RefCell<Option<Driver>>,
-    _phantom: PhantomUnsendUnsync,
+    driver: RefCell<Option<driver::Handle>>,
```
FrankReh, line 7: Interesting the unsend/unsync type is needed any more.

src/io/writev.rs
```diff
@@ -32,22 +33,24 @@ impl<T: IoBuf> Op<Writev<T>> {
...
-                    types::Fd(fd.raw_fd()),
-                    write.iovs.as_ptr(),
-                    write.iovs.len() as u32,
-                )
-                .offset(offset as _)
-                .build()
-            },
-        )
+        CONTEXT.with(|x| {
+            x.handle().expect("Not in a runtime context").submit_op(
```
FrankReh, line 37: This looks good too. Makes it clear the Op can only be made if there is a local runtime context and the strong handle is available.

src/io/recv_from.rs
```diff
@@ -31,22 +32,24 @@ impl<T: BoundedBufMut> Op<RecvFrom<T>> {
...
-            },
-            |recv_from| {
-                opcode::RecvMsg::new(
-                    types::Fd(recv_from.fd.raw_fd()),
-                    recv_from.msghdr.as_mut() as *mut _,
-                )
-                .build()
-            },
-        )
+        CONTEXT.with(|x| {
```
FrankReh, line 35: I could see a macro serving well here, in all these op files, to reduce boilerplate and make future changes easier.

FrankReh (approved):

src/runtime/driver/handle.rs
```diff
@@ -0,0 +1,286 @@
...
+use std::rc::{Rc, Weak};
+use std::task::{Context, Poll};
+
+use crate::buf::fixed::FixedBuffers;
+use crate::runtime::driver::op::{Completable, Lifecycle, MultiCQEFuture, Op, Updateable};
+use crate::runtime::driver::Driver;
+
+#[derive(Clone)]
+pub(crate) struct Handle {
+    pub(super) inner: Rc<RefCell<Driver>>,
```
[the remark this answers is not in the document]
Noah-Kennedy: > How would you feel about naming this field driver? And the WeakHandle's inner `driver`` too? Just thinking out loud. This isn't even a nit. Just wanting to discuss names, the crate will live with them a long time.

I think that this is a good idea.

> Also where the Handle type is used for other fields in this PR, maybe name the field handle instead of driver? There's some precedent for that in the tokio crate I think. Then one can say or think "the handle to the driver", and it syncs with the code. I guess thinking "the weak handle to the driver" would mean the inner field below could be named driver too.

I named it driver because we might end up with multiple handles potentially. I could see us having handles for things like buffer registries as well.

src/runtime/mod.rs
```diff
@@ -61,7 +62,10 @@ impl Runtime {
         let rt = tokio::runtime::Builder::new_current_thread()
```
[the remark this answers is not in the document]
Noah-Kennedy: I think this is a good idea, but I'm trying to avoid touching too much existing code in this PR. I want to try and scope this refactor PR as narrowly as I can.

src/runtime/mod.rs
```diff
@@ -85,29 +87,44 @@ impl Runtime {
...
-        Ok(Runtime { local, rt })
+        Ok(Runtime { local, rt, driver })
     }
 
     /// Runs a future to completion on the current runtime
     pub fn block_on<F>(&self, future: F) -> F::Output
     where
         F: Future,
     {
+        struct ContextGuard;
```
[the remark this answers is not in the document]
Noah-Kennedy: It's public because there were requests to make it public. I forget why. I actually agree that it should be public ideally, but I think that we made it public prematurely. The current API feels very rough and messy.

src/runtime/mod.rs
```diff
@@ -21,6 +19,9 @@ pub struct Runtime {
     /// LocalSet for !Send tasks
     local: ManuallyDrop<LocalSet>,
 
+    /// Strong reference to the driver.
```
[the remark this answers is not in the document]
Noah-Kennedy: Don't try to make too much sense of this logic right now as it's horrible. I need to do more cleanup once this is in.

src/runtime/mod.rs
```diff
@@ -85,29 +87,44 @@ impl Runtime {
...
         tokio::pin!(future);
 
-        self.rt
+        let res = self
+            .rt
             .block_on(self.local.run_until(std::future::poll_fn(|cx| {
                 // assert!(drive.as_mut().poll(cx).is_pending());
                 future.as_mut().poll(cx)
-            })))
+            })));
```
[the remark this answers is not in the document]
Noah-Kennedy: Possibly, but let's leave that to another PR.

src/runtime/mod.rs
```diff
@@ -85,29 +87,44 @@ impl Runtime {
...
     }
 
     /// Runs a future to completion on the current runtime
     pub fn block_on<F>(&self, future: F) -> F::Output
     where
         F: Future,
     {
+        struct ContextGuard;
+
+        impl Drop for ContextGuard {
```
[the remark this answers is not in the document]
Noah-Kennedy: Yeah, I'd like us to eventually have something like this, but I've left that work to another PR.

src/runtime/driver/handle.rs
Noah-Kennedy: I'm trying to keep the logic around the driver self-contained to the driver and handle. I want the Op to not have to know about the guts of the driver, just how to communicate with APIs provided by the driver.

src/runtime/context.rs
Noah-Kennedy: You bring up a good point here, this absolutely needs documented.

src/runtime/driver/handle.rs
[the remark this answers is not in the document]
FrankReh: I think the Handle for the driver will be the most important so it would be okay to go by the short name handle. If there are other Handle types down the road, and the variable is long lived, it gets a name other than simply handle. JMO.

PR #197, opened 2022-12-04
rt: refactor driver, runtime, and op

Noah-Kennedy:
This cleans up a few different pieces and is a followup change from #196.

### DIFF
--- a/src/runtime/driver/handle.rs
+++ b/src/runtime/driver/handle.rs
@@ -12,15 +12,16 @@
 //! The weak handle should be used by anything which is stored in the driver or does not need to
 //! keep the driver alive for it's duration.
 
-use io_uring::{cqueue, squeue};
+use io_uring::squeue;
 use std::cell::RefCell;
 use std::io;
+use std::ops::Deref;
 use std::os::unix::io::{AsRawFd, RawFd};
 use std::rc::{Rc, Weak};
 use std::task::{Context, Poll};
 
 use crate::buf::fixed::FixedBuffers;
-use crate::runtime::driver::op::{Completable, Lifecycle, MultiCQEFuture, Op, Updateable};
+use crate::runtime::driver::op::{Completable, MultiCQEFuture, Op, Updateable};
 use crate::runtime::driver::Driver;
 
 #[derive(Clone)]
@@ -40,8 +41,8 @@ impl Handle {
         })
     }
 
-    pub(crate) fn tick(&self) {
-        self.inner.borrow_mut().tick()
+    pub(crate) fn dispatch_completions(&self) {
+        self.inner.borrow_mut().dispatch_completions()
     }
 
     pub(crate) fn flush(&self) -> io::Result<usize> {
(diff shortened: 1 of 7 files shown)


src/runtime/driver/handle.rs
```diff
@@ -52,99 +53,29 @@ impl Handle {
...
-        let mut driver = self.inner.borrow_mut();
-
-        driver
-            .uring
-            .submitter()
-            .register_buffers(buffers.borrow().iovecs())?;
-
-        driver.fixed_buffers = Some(buffers);
-        Ok(())
+        self.inner.borrow_mut().register_buffers(buffers)
```
FrankReh, line 56: Lots of layers for getting register_buffers called now. Is this one really necessary or is this a stop gap version?

src/runtime/driver/mod.rs
```diff
@@ -56,7 +57,7 @@ impl Driver {
         self.ops.lifecycle.len()
     }
 
-    pub(crate) fn tick(&mut self) {
+    pub(crate) fn dispatch_completions(&mut self) {
```
FrankReh, line 60: True the name `tick` didn't do this justice since it was not controlled by a clock. How about something without the name completions in it? Complete and completable are all over the place already. Could you go with `cq_sync`? Or even `cq_sync_and_dispatch`?

src/runtime/driver/mod.rs
```diff
@@ -82,7 +83,7 @@ impl Driver {
                     return Ok(());
                 }
                 Err(ref e) if e.raw_os_error() == Some(libc::EBUSY) => {
-                    self.tick();
+                    self.dispatch_completions();
```
FrankReh, line 86: How would you feel about moving this `submit` function above the earlier `dispatch` function? In some ways, it would come first since it would be used first.

src/runtime/driver/mod.rs
```diff
@@ -91,6 +92,198 @@ impl Driver {
...
+                    Poll::Ready(cqe) => {
+                        self.ops.remove(op.index());
+                        Poll::Ready(data.complete(cqe))
+                    }
+                }
+            }
+        }
+    }
+
+    pub(crate) fn remove_op<T, CqeType>(&mut self, op: &mut Op<T, CqeType>) {
```
FrankReh, line 251: Should `remove_op` go right after `submit_op` a little above in this file?

src/runtime/mod.rs
```diff
@@ -130,14 +115,36 @@ impl Runtime {
...
 
+fn start_uring_wakes_task(
+    tokio_rt: &tokio::runtime::Runtime,
+    local: &LocalSet,
+    driver: driver::Handle,
+) {
+    let _guard = tokio_rt.enter();
+    let async_driver_handle = AsyncFd::new(driver).unwrap();
+
+    local.spawn_local(drive_uring_wakes(async_driver_handle));
```
FrankReh, line 134: If I may, I think 'driver' is still being used too much. Everything is a driver. Then nothing is. 😄

src/runtime/mod.rs
```diff
@@ -130,14 +115,36 @@ impl Runtime {
...
+}
+
+async fn drive_uring_wakes(driver: AsyncFd<driver::Handle>) {
```
FrankReh, line 137: `drive_uring_wakes` is not bad. By which I mean, I kind of like it. It's very descriptive. Unfortunately it doesn't call a `uring_wakes`. Maybe `drive_uring_sync` or `drive_cq_sync`?

And would you mind using whatever name you pick for the argument on line 134, as the parameter name here? Or pick a super short name, like the Go developers and old programmers like, since the type is everything and the name not so much. `afd` for example.
I just can't bring myself to thinking of it as a driver too.

In case I seem annoying, first I apologize. Second, look at line 140. The driver is not readable, the async_fd is readable, or becomes readable once the Future is resolved. True the unwrap yields a guard to the driver's Handle. I can think of it as a guard handle, even if the name is just guard.

FrankReh (changes requested):
Very nice. And quick. I'm setting this to 'request changes' really just to get your reaction to everything. If you don't want to change anything that's okay. Then I'll set to approved.

FrankReh:
@Noah-Kennedy Oh, also make clippy happy.

src/runtime/driver/handle.rs
Noah-Kennedy: I'm trying to keep everything which touches the driver's state contained to the driver.

src/runtime/driver/mod.rs
Noah-Kennedy: I feel like this name does a good job of describing what we are doing here. We should probably change some other methods though. You are right about there being too many things with `complete` in the name.

src/runtime/driver/mod.rs
Noah-Kennedy: Sure, I don't really mind the placement currently.

src/runtime/driver/mod.rs
Noah-Kennedy: I don't really care much how these are ordered. I'm fine either way.

src/runtime/mod.rs
Noah-Kennedy: Yeah, we should probably rename a bunch of the `driver` vars to `driver_handle`.

src/runtime/mod.rs
Noah-Kennedy: I think it's probably best for us to describe what we are doing here, which isn't just regarding syncing the cqueue, but more importantly waking tasks.

src/runtime/driver/mod.rs
Noah-Kennedy: Done.

src/runtime/driver/mod.rs
Noah-Kennedy: Done.

Noah-Kennedy:
@FrankReh could you take a re-review?

FrankReh (approved):
Looks good.

PR #198, opened 2022-12-04
Add method `BoundedBuf::put_slice`

mzabaluev:
Provide a generic,  safe way to copy data into a buffer, possibly advancing the initialized length watermark.
This method is especially useful for `FixedBuf`, which lacks other population methods available for the foreign types.

This resolves a discussion started in https://github.com/tokio-rs/tokio-uring/pull/190#issuecomment-1336301837

### DIFF
--- a/src/buf/bounded.rs
+++ b/src/buf/bounded.rs
@@ -154,7 +154,7 @@ pub trait BoundedBufMut: BoundedBuf<Buf = Self::BufMut> {
     /// to `pos` are initialized and owned by the buffer.
     unsafe fn set_init(&mut self, pos: usize);
 
-    /// Copies the given byte slice slice into the buffer, starting at
+    /// Copies the given byte slice into the buffer, starting at
     /// this view's offset.
     ///
     /// # Panics
--- a/src/buf/bounded.rs
+++ b/src/buf/bounded.rs
@@ -1,6 +1,7 @@
 use super::{IoBuf, IoBufMut, Slice};
 
 use std::ops;
+use std::ptr;
 
 /// A possibly bounded view into an owned [`IoBuf`] buffer.
 ///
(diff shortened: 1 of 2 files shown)


src/buf/bounded.rs
```diff
@@ -152,6 +153,29 @@ pub trait BoundedBufMut: BoundedBuf<Buf = Self::BufMut> {
     /// The caller must ensure that all bytes starting at `stable_mut_ptr()` up
     /// to `pos` are initialized and owned by the buffer.
     unsafe fn set_init(&mut self, pos: usize);
+
+    /// Copies the given byte slice slice into the buffer, starting at
```
FrankReh, line 157: 'slice slice' -> 'slice'

FrankReh (changes requested):
Super.

~~Could you see the use cases for providing two other methods, one for zeroing to the init point and one for zeroing to the capacity and setting the init point to the capacity?~~

Seems code like
```rust
ary.iter_mut().for_each(|m| *m = 0)
```
~~is expected to be optimized to a memset call, per https://users.rust-lang.org/t/fastest-way-to-zero-an-array/39222
but that's from a few years ago.~~

One typo, and maybe wait for Noah's second commit today and this is good to go.

FrankReh:
Following the discussion further, I see there is a slice fill method since 1.50.
https://doc.rust-lang.org/std/primitive.slice.html#method.fill

Noah-Kennedy:
Once the typo is fixed I'm in favor of merging.

src/buf/bounded.rs
mzabaluev: https://github.com/tokio-rs/tokio-uring/pull/198/commits/83519e4fc1cd8d58a78487c58fa74fe8e096ca08

FrankReh (approved):
Thanks

PR #199, opened 2022-12-06
`FixedBufPool::next`

mzabaluev:
Add an async API to get buffers out of `FixedBufPool`. This is another way in which `FixedBufPool` will be more useful than `FixedBufRegistry`.

Also reorganize the shared data structures of `FixedBufPool` and `FixedBufRegistry` into separate modules for better readability and future code reuse.

### DIFF
--- a/src/buf/fixed/plumbing/pool.rs
+++ b/src/buf/fixed/plumbing/pool.rs
@@ -6,6 +6,7 @@ use std::collections::HashMap;
 use std::mem;
 use std::ptr;
 use std::slice;
+use std::task::{Context, Poll, Waker};
 
 // Internal state shared by FixedBufPool and FixedBuf handles.
 pub(crate) struct Pool {
@@ -20,6 +21,8 @@ pub(crate) struct Pool {
     states: Vec<BufState>,
     // Table of head indices of the free buffer lists in each size bucket.
     free_buf_head_by_cap: HashMap<usize, u16>,
+    // Wakers for tasks pending on poll_next
+    waiting_on_next: Vec<Waker>,
 }
 
 // State information of a buffer in the registry,
@@ -73,6 +76,7 @@ impl Pool {
             orig_cap,
             states,
             free_buf_head_by_cap,
+            waiting_on_next: vec![],
         }
     }
 
@@ -114,6 +118,17 @@ impl Pool {
         })
     }
 
+    pub(crate) fn poll_next(&mut self, cap: usize, cx: &mut Context<'_>) -> Poll<CheckedOutBuf> {
+        if let Some(buf) = self.try_next(cap) {
+            return Poll::Ready(buf);
+        }
+        let waker = cx.waker();
+        if !self.waiting_on_next.iter().any(|w| w.will_wake(waker)) {
+            self.waiting_on_next.push(waker.clone());
+        }
+        Poll::Pending
+    }
+
     fn check_in_internal(&mut self, index: u16, init_len: usize) {
         let cap = self.iovecs()[index as usize].iov_len;
         let state = &mut self.states[index as usize];
@@ -128,6 +143,13 @@ impl Pool {
         let next = self.free_buf_head_by_cap.insert(cap, index);
 
         *state = BufState::Free { init_len, next };
+
+        if !self.waiting_on_next.is_empty() {
+            // Wake up tasks pending on poll_next
+            for waker in mem::take(&mut self.waiting_on_next) {
+                waker.wake()
+            }
+        }
     }
 }
 
--- a/src/buf/fixed/plumbing/pool.rs
+++ b/src/buf/fixed/plumbing/pool.rs
@@ -23,7 +23,7 @@ pub(crate) struct Pool {
     states: Vec<BufState>,
     // Table of head indices of the free buffer lists in each size bucket.
     free_buf_head_by_cap: HashMap<usize, u16>,
-    // Used to notify tasks pending on Next
+    // Used to notify tasks pending on `next`
     notify_next_by_cap: HashMap<usize, Arc<Notify>>,
 }
 
--- a/src/buf/fixed/plumbing/pool.rs
+++ b/src/buf/fixed/plumbing/pool.rs
@@ -1,12 +1,14 @@
 use crate::buf::fixed::{handle::CheckedOutBuf, FixedBuffers};
 
 use libc::{iovec, UIO_MAXIOV};
+use tokio::sync::Notify;
+
 use std::cmp;
 use std::collections::HashMap;
 use std::mem;
 use std::ptr;
 use std::slice;
-use std::task::{Context, Poll, Waker};
+use std::sync::Arc;
 
 // Internal state shared by FixedBufPool and FixedBuf handles.
 pub(crate) struct Pool {
@@ -21,8 +23,8 @@ pub(crate) struct Pool {
     states: Vec<BufState>,
     // Table of head indices of the free buffer lists in each size bucket.
     free_buf_head_by_cap: HashMap<usize, u16>,
-    // Wakers for tasks pending on poll_next
-    waiting_on_next: Vec<Waker>,
+    // Used to notify tasks pending on Next
+    notify_next_by_cap: HashMap<usize, Arc<Notify>>,
 }
 
 // State information of a buffer in the registry,
@@ -76,7 +78,7 @@ impl Pool {
             orig_cap,
             states,
             free_buf_head_by_cap,
-            waiting_on_next: vec![],
+            notify_next_by_cap: HashMap::new(),
         }
     }
 
@@ -118,15 +120,9 @@ impl Pool {
         })
     }
 
-    pub(crate) fn poll_next(&mut self, cap: usize, cx: &mut Context<'_>) -> Poll<CheckedOutBuf> {
-        if let Some(buf) = self.try_next(cap) {
-            return Poll::Ready(buf);
-        }
-        let waker = cx.waker();
-        if !self.waiting_on_next.iter().any(|w| w.will_wake(waker)) {
-            self.waiting_on_next.push(waker.clone());
-        }
-        Poll::Pending
+    pub(crate) fn notify_on_next(&mut self, cap: usize) -> Arc<Notify> {
+        let notify = self.notify_next_by_cap.entry(cap).or_default();
+        Arc::clone(notify)
     }
 
     fn check_in_internal(&mut self, index: u16, init_len: usize) {
@@ -144,11 +140,9 @@ impl Pool {
 
         *state = BufState::Free { init_len, next };
 
-        if !self.waiting_on_next.is_empty() {
-            // Wake up tasks pending on poll_next
-            for waker in mem::take(&mut self.waiting_on_next) {
-                waker.wake()
-            }
+        if let Some(notify) = self.notify_next_by_cap.get(&cap) {
+            // Wake up a single task pending on `next`
+            notify.notify_one();
         }
     }
 }
--- a/src/buf/fixed/plumbing/pool.rs
+++ b/src/buf/fixed/plumbing/pool.rs
@@ -120,6 +120,8 @@ impl Pool {
         })
     }
 
+    // Returns a `Notify` to use for waking up tasks awaiting a buffer of
+    // the specified capacity.
     pub(crate) fn notify_on_next(&mut self, cap: usize) -> Arc<Notify> {
         let notify = self.notify_next_by_cap.entry(cap).or_default();
         Arc::clone(notify)
--- a/src/buf/fixed/plumbing/pool.rs
+++ b/src/buf/fixed/plumbing/pool.rs
@@ -1,0 +1,164 @@
+use crate::buf::fixed::{handle::CheckedOutBuf, FixedBuffers};
+
+use libc::{iovec, UIO_MAXIOV};
+use std::cmp;
+use std::collections::HashMap;
+use std::mem;
+use std::ptr;
+use std::slice;
+
+// Internal state shared by FixedBufPool and FixedBuf handles.
+pub(crate) struct Pool {
+    // Pointer to an allocated array of iovec records referencing
+    // the allocated buffers. The number of initialized records is the
+    // same as the length of the states array.
+    raw_bufs: ptr::NonNull<iovec>,
+    // Original capacity of raw_bufs as a Vec.
+    orig_cap: usize,
+    // State information on the buffers. Indices in this array correspond to
+    // the indices in the array at raw_bufs.
+    states: Vec<BufState>,
+    // Table of head indices of the free buffer lists in each size bucket.
+    free_buf_head_by_cap: HashMap<usize, u16>,
+}
+
+// State information of a buffer in the registry,
+enum BufState {
+    // The buffer is not in use.
+    Free {
+        // This field records the length of the initialized part.
+        init_len: usize,
+        // Index of the next buffer of the same capacity in a free buffer list, if any.
+        next: Option<u16>,
+    },
+    // The buffer is checked out.
+    // Its data are logically owned by the FixedBuf handle,
+    // which also keeps track of the length of the initialized part.
+    CheckedOut,
+}
+
+impl Pool {
+    pub(crate) fn new(bufs: impl Iterator<Item = Vec<u8>>) -> Self {
+        let bufs = bufs.take(cmp::min(UIO_MAXIOV as usize, u16::MAX as usize));
+        let (size_hint, _) = bufs.size_hint();
+        let mut iovecs = Vec::with_capacity(size_hint);
+        let mut states = Vec::with_capacity(size_hint);
+        let mut free_buf_head_by_cap = HashMap::new();
+        for (index, mut buf) in bufs.enumerate() {
+            let cap = buf.capacity();
+
+            // Link the buffer as the head of the free list for its capacity.
+            // This constructs the free buffer list to be initially retrieved
+            // back to front, which should be of no difference to the user.
+            let next = free_buf_head_by_cap.insert(cap, index as u16);
+
+            iovecs.push(iovec {
+                iov_base: buf.as_mut_ptr() as *mut _,
+                iov_len: cap,
+            });
+            states.push(BufState::Free {
+                init_len: buf.len(),
+                next,
+            });
+            mem::forget(buf);
+        }
+        debug_assert_eq!(iovecs.len(), states.len());
+
+        // Safety: Vec::as_mut_ptr never returns null
+        let raw_bufs = unsafe { ptr::NonNull::new_unchecked(iovecs.as_mut_ptr()) };
+        let orig_cap = iovecs.capacity();
+        mem::forget(iovecs);
+        Pool {
+            raw_bufs,
+            orig_cap,
+            states,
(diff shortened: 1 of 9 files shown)


Noah-Kennedy (approved):

Noah-Kennedy:
ugh, I did not mean to remove the draft status

src/buf/fixed/plumbing/pool.rs
```diff
@@ -0,0 +1,186 @@
...
+            index: index as u16,
+        })
+    }
+
+    pub(crate) fn poll_next(&mut self, cap: usize, cx: &mut Context<'_>) -> Poll<CheckedOutBuf> {
+        if let Some(buf) = self.try_next(cap) {
+            return Poll::Ready(buf);
+        }
+        let waker = cx.waker();
+        if !self.waiting_on_next.iter().any(|w| w.will_wake(waker)) {
```
FrankReh, line 126: Sounds like will_wake won't catch all duplicates. Is it possible this list will grow too large? And what's the danger of keeping an old waker that the scheduler had replaced in a subsequent poll?

src/buf/fixed/plumbing/pool.rs
```diff
@@ -0,0 +1,186 @@
...
+        // Link the buffer as the new head of the free list for its capacity.
+        // Recently checked in buffers will be first to be reused,
+        // improving cache locality.
+        let next = self.free_buf_head_by_cap.insert(cap, index);
+
+        *state = BufState::Free { init_len, next };
+
+        if !self.waiting_on_next.is_empty() {
+            // Wake up tasks pending on poll_next
+            for waker in mem::take(&mut self.waiting_on_next) {
```
FrankReh, line 149: Is waking them all necessary because one or more may have been canceled?

FrankReh (commented):
Nice.

src/buf/fixed/plumbing/pool.rs
mzabaluev: > Sounds like will_wake won't catch all duplicates. Is it possible this list will grow too large?

It is theoretically, but I don't know a way to prevent this.

> And what's the danger of keeping an old waker that the scheduler had replaced in a subsequent poll?

I trust the `Waker` API to do the right thing in this case.

src/buf/fixed/plumbing/pool.rs
mzabaluev: I don't get this question, sorry. All stored wakers are used up to wake any of the tasks that still remain, and `Waker::wake` is a no-op if its task has terminated. If a task has dropped the `Next` future (or otherwise stopped calling `poll_next`), it will get a spurious wakeup once, but I think this is normal in the async world.

FrankReh: Oh, you got my question. You just couldn't hear my tone of voice of my body language. Like many of my questions, it came half from naiveté and half from just wanting to make conversation and wanting to double check and keep my brain thinking during the review.

It's the first time I've seen it in action but I agree it's necessary. It's maybe a small blemish on the async world where the wake call does not return the state of whether it found a Future to put into the `Ready` queue.

Thank you.

src/buf/fixed/plumbing/pool.rs
FrankReh: Tokio still has an open issue about this. https://github.com/rust-lang/rust/issues/66281

How about keeping an optional clone in the future itself, to allow searching the `Vec` with the older one first, with simple address comparison? I'm afraid the `Vec`'s growth could be unbounded if the future were part of a select in a loop as long as Tokio's waker API doesn't guarantee will_wake doesn't have false negatives.

FrankReh (changes requested):
The direction is great. Please consider my two outstanding questions. Thank you.

src/buf/fixed/plumbing/pool.rs
```diff
@@ -20,6 +21,8 @@ pub(crate) struct Pool {
     states: Vec<BufState>,
     // Table of head indices of the free buffer lists in each size bucket.
     free_buf_head_by_cap: HashMap<usize, u16>,
+    // Wakers for tasks pending on poll_next
+    waiting_on_next: Vec<Waker>,
```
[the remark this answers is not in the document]
mzabaluev: I don't think this requires an enum, can you sketch some code for that? Also, keeping the entry with an empty `Vec` value, even when all wakers for a given capacity are consumed, may be better than removing the `Vec` from the table and re-adding it later for another poll.

I thought about making this a hash table indexed per capacity. But then, the waker list(s) is not going to be populated unless the pool becomes exhausted, and when it does, we're anyway up for some latency, re-polling, spurious wakeups, etc.
So I would not like to make this logic more complex and branchy without having some performance data to understand whether this would make a significant difference.

src/buf/fixed/plumbing/pool.rs
mzabaluev: > How about keeping an optional clone in the future itself, to allow searching the `Vec` with the older one first, with simple address comparison?

How could that work with cloned wakers that do not expose any underlying pointers in stable API?

FrankReh: If you store a waker in the Future, don't you have a stable address you can compare against?

mzabaluev: No. The reference returned by `cx.waker()` is not guaranteed to live longer than the poll call that the `Context` is passed into. Cloning the waker is possible, but what address comparison against other `Waker` instances is going to achieve then? The only API we have to avoid storing redundant wakers is `Waker::will_wake`. That's not guaranteed to always work, but I think in practice no executor will repeatedly produce wakers that can't be correlated to one another with `will_wake` when the task is repeatedly polled on a pending future. Correct me please if there are examples to the contrary.

FrankReh: I'm sorry I don't see the problem yet with storing a cloned waker in the future when the cloned waker is also pushed onto the `Vec` so when poll is called again, before pushing a new cloned waker onto the `Vec`, the cloned waker held by the future is used to get the old waker's address and that address is searched in the `Vec` to see if a `Vec` location can be swapped out with a new clone. I have not seen any structures in the std library where a list of wakers is held. I'm prepared to write a demo to see if this idea would work, I don't see why it wouldn't - the waker is a function and it has an address that can be compared, but I haven't had time yet.

I know `will_wake` is used for the first best shot at deciding when a waker doesn't need to be saved because an equally valid waker is already held. But in all the cases I've seen, when `will_wake` returns false, it means a waker is replaced with a clone, not appended as a clone. I'm just trying to find the location in the `Vec` to perform the replacement by comparing pointers. Another approach, not relying on the waker address, would be to store a generation counter along with the waker in the `Vec` cell and keep the generation counter in the future, for searching the `Vec` on a subsequent poll.

mzabaluev: I have rewritten the wakeup implementation using `tokio::sync::Notify`. It's non-allocating, besides the synchronization objects created on demand for each capacity bucket.

src/buf/fixed/plumbing/pool.rs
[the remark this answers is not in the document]
mzabaluev: With the new implementation, wakeups do happen per capacity bucket, and a single task is awoken after each buffer check-in, if one is waiting for a buffer of this capacity.

src/buf/fixed/plumbing/pool.rs
```diff
@@ -0,0 +1,180 @@
...
+        let iovec = unsafe { self.raw_bufs.as_ptr().add(index).read() };
+        debug_assert_eq!(iovec.iov_len, cap);
+        Some(CheckedOutBuf {
+            iovec,
+            init_len,
+            index: index as u16,
+        })
+    }
+
+    pub(crate) fn notify_on_next(&mut self, cap: usize) -> Arc<Notify> {
```
mzabaluev, line 123: The `Arc` is not necessary for the single-threaded pool, but I'm going to reuse this for the thread-safe implementation. `Notify` contains a mutex and uses atomic pointers for tracking waiters, so it's going to be slow regardless when we hit the waiting branch. Performance of `next` in the depleted pool case is not high on my current priority list for micro-optimizations.

tests/fixed_buf.rs
```diff
@@ -137,6 +139,54 @@ fn slicing() {
...
+            let mut buf = buffers.next(BUF_SIZE).await;
+            println!(
+                "[main] iteration {}: obtained buffer {}",
+                i,
+                buf.buf_index()
+            );
+            let cloned_file = file.try_clone().unwrap();
+
+            let handle = tokio_uring::spawn(async move {
+                let file = File::from_std(cloned_file);
```
mzabaluev, line 167: There should be a way to share `File` and socket objects between tasks, since the descriptor is already in a `SharedFd` internally.

src/buf/fixed/plumbing/pool.rs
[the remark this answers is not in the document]
FrankReh: Thanks for this clarification. I'll have to look at the code another time though to see how waking one and only one task is safe. Is the buffer ownership handed to the Future as part of being awoken, even before another poll? No need to answer, I'll look for the answer this weekend.

src/buf/fixed/pool.rs
```diff
@@ -223,164 +232,57 @@ impl FixedBufPool {
...
-                BufState::CheckedOut => unreachable!("all buffers must be checked in"),
+            if let Some(data) = self.inner.borrow_mut().try_next(cap) {
+                // Safety: the validity of buffer data is ensured by
+                // plumbing::Pool::try_next
+                let buf = unsafe { FixedBuf::new(Rc::clone(&self.inner) as _, data) };
+                return buf;
             }
+
+            // Reset the `Notified` future in case another call to `try_next`
+            // got the buffer before us.
```
FrankReh, line 284: Maybe the comment could add that the permit may have been held even though there was no buffer, which is likely once the exhaustion case has gone away and then comes back.

FrankReh (approved):
I'm not going to check the unit tests tonight. But the use of Notify looks very nice. That is one powerful `waiting` mechanism `Tokio` has to offer. I ended up liking this implementation more than I thought I would because the Notify does all the heavy lifting of avoiding the pitfall of dropping an awake call. And as result, *all* the waiters don't have to be awoken to be on the safe side. The `Notify` implementation makes it safe to begin with, hence `notify_one` can be used very effectively.

src/buf/fixed/plumbing/pool.rs
```diff
@@ -0,0 +1,180 @@
...
+        let iovecs = unsafe {
+            Vec::from_raw_parts(self.raw_bufs.as_ptr(), self.states.len(), self.orig_cap)
+        };
+        for (i, iovec) in iovecs.iter().enumerate() {
+            match self.states[i] {
+                BufState::Free { init_len, next: _ } => {
+                    let ptr = iovec.iov_base as *mut u8;
+                    let cap = iovec.iov_len;
+                    let v = unsafe { Vec::from_raw_parts(ptr, init_len, cap) };
+                    mem::drop(v);
```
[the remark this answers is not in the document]
mzabaluev: Each buffer must be individually allocated since it's supplied to the constructor as a `Vec`.
I don't think io-uring requires the buffers to be aligned; tests pass with buffers of size 80, for example.

I put a `mem::drop` call in there to illustrate that the `Vec` is recovered from raw pointer and size specifically to be dropped and thus to get the buffer properly deallocated.

src/buf/fixed/plumbing/pool.rs
```diff
@@ -0,0 +1,180 @@
...
+    raw_bufs: ptr::NonNull<iovec>,
+    // Original capacity of raw_bufs as a Vec.
+    orig_cap: usize,
+    // State information on the buffers. Indices in this array correspond to
+    // the indices in the array at raw_bufs.
+    states: Vec<BufState>,
+    // Table of head indices of the free buffer lists in each size bucket.
+    free_buf_head_by_cap: HashMap<usize, u16>,
+    // Used to notify tasks pending on `next`
+    notify_next_by_cap: HashMap<usize, Arc<Notify>>,
```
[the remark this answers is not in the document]
mzabaluev: Yes, it's possible that a task calling `next` will consume the permit, not get a buffer from `try_next`, and continue waiting. It's also possible in other scenarios, for example when tasks are calling `next` and `try_next` on the same pool concurrently.

src/buf/fixed/pool.rs
```diff
@@ -223,164 +232,57 @@ impl FixedBufPool {
...
+            notified.set(notify.notified());
```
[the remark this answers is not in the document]
mzabaluev: I lifted this from the code example on `Notified` :grin:

mzabaluev:
@FrankReh
> But we are left without a simple/fast pool, that doesn't allocate, for the current-thread case.

While the pool is not exhausted, tasks call `next` at near zero cost with no allocations.

src/buf/fixed/plumbing/pool.rs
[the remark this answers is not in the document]
FrankReh: Is that worth a comment somewhere, to let the reader know it has been considered, or perhaps that it is by design? When reading something new, it's nice to see that all the bases were covered by a design, before having to figure it out myself.

src/buf/fixed/plumbing/pool.rs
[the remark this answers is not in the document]
FrankReh: I'm not sure about this. If the constructor didn't do the individual allocators, it doesn't seem right that the drop would assume individual deallocations.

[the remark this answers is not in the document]
FrankReh: At least some kernel ops require page aligned buffers. I don't recall which right now but why shouldn't the interface allow the caller to decide for themselves how the allocations are done? Does that make the API too onerous?

FrankReh:
@mzabaluev 
> While the pool is not exhausted, tasks call `next` at near zero cost with no allocations.
That's a very good point. I think I overlooked that while digging into the weeds of the implementation.

src/buf/fixed/plumbing/pool.rs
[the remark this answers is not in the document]
mzabaluev: > If the constructor didn't do the individual allocators, it doesn't seem right that the drop would assume individual deallocations.

The constructor takes ownership of a sequence of `Vec` that have to be allocated using the global allocator, so drop deallocates these `Vec` objects. If the application wants a more precise control over the allocations, perhaps it should consider its own implementation of a `FixedBuffers` collection (when we get around to make the trait public).

src/buf/fixed/plumbing/pool.rs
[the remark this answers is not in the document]
mzabaluev: https://github.com/tokio-rs/tokio-uring/pull/199/commits/d83eacb9a484be4cb9567a9977ea117fa833fc56

src/buf/fixed/pool.rs
mzabaluev: https://github.com/tokio-rs/tokio-uring/pull/199/commits/d83eacb9a484be4cb9567a9977ea117fa833fc56

src/buf/fixed/plumbing/pool.rs
```diff
@@ -0,0 +1,180 @@
...
+        let iovec = unsafe { self.raw_bufs.as_ptr().add(index).read() };
+        debug_assert_eq!(iovec.iov_len, cap);
+        Some(CheckedOutBuf {
+            iovec,
+            init_len,
+            index: index as u16,
+        })
+    }
+
+    pub(crate) fn notify_on_next(&mut self, cap: usize) -> Arc<Notify> {
```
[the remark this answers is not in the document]
mzabaluev: https://github.com/tokio-rs/tokio-uring/pull/199/commits/091cb4de5548a6311b2d351c6c358ad1d90461d8

ollie-etl:
@Noah-Kennedy @mzabaluev Is something blocking this? It looks like all points have been resolved

Noah-Kennedy:
You don't need my approval here. I'm alright with @FrankReh signing off on this and merging it whenever he feels it is ready.

src/buf/fixed/plumbing/pool.rs
ollie-etl: I'd just like to clarify my understanding here: once a notify has been registered for a capacity, then every time a buffer of that capacity is commited, notify_one() is called (because the notify_next_by_cap entry is never removed)?

oliverbunting:
> @mzabaluev 
> > While the pool is not exhausted, tasks call `next` at near zero cost with no allocations.
> 
> That's a very good point. I think I overlooked that while digging into the weeds of the implementation.
> 

Is this not a common case though. I.e, to keep some pipeline of tasks fully utilized, do work whenever there is a buffer available?

mzabaluev:
> Is this not a common case though. I.e, to keep some pipeline of tasks fully utilized, do work whenever there is a buffer available?

Depends if the task creation is only throttled by availability of the buffers.

src/buf/fixed/plumbing/pool.rs
ollie-etl: I think this could be improved by keeping a `(usize, Notify)` instead, to explicitly keep a count for the number of waiting tasks, and removing  the entry from the hashset once count == zero. That way, you only pay the `SeqCst` cost of calling `notify()` when you need to.

FrankReh: Not saying this is needed, I will let @mzabaluev set the path on this one. But I would prefer something that had a little hysteresis to it. Like removing the Notify if the pool size reached something like 10 buffers again. (But I have no idea why I would pick 10. Just something larger than 0.) But I haven't tried to design anything around this idea so feel free to disregard entirely.

ollie-etl: My concern is that with the current design, once `try_next()` has failed once in the `next()` method. every checking from then on for that buffer size is going to issue a `SeqCst` memory barrier with the call to `notify_one()`. Thats quite a penalty. Its unavoidable with the surrent Notify, which is designed for `Send` operations, but we could at least limit it to when there is actually a consumer waiting.

FrankReh: @ollie-etl With your idea, I don't see how the usize field is decremented properly? In the case of a future being cancelled, there is no future to do the work of updating the count (from my understanding).

ollie-etl: You'd need to return a `Next` struct wrapper, which decrements on drop

FrankReh: Oh, that's much more like @mzabaluev 's original idea I think, of returning a future that does work when it is polled, and does work when it is dropped. But using Notify.

FrankReh:
I'm going to merge later tonight so it makes it into the next release. This is some really great work that we want in the repo but I had to do a manual merge on four files because code moved and code changed at the same time.

Actually the condensed diff that this page makes available isn't too bad to read. Most of the changes are straight forward. 

I guess the drop code, and the intra-doc links are where there could be problems so another pair of eyes or two or three wouldn't hurt. It compiles cleanly, and there is a nice unit test that passes. But rarely do our unit tests check that things were dropped properly so that's why looking at the drop again would make sense.

But also, nothing we can't change yet again with more iterations.

FrankReh:
Thank you @mzabaluev .

PR #200, opened 2022-12-10
Restore off-runtime creation of fixed buffer collections

mzabaluev:
#196 broke the ability to populate a collection of fixed buffers off-runtime and then have it registered inside a runtime.

There is no need for the driver when a fixed buffer collection is populated. The driver comes in when registering and unregistering the buffers, and then as well, we don't currently need the collection to be affine to the driver. The driver always compares the pointer passed in to prevent unregistering the wrong buffers, so we can just grab the driver from the context.

Change the doc examples to show and exercise off-runtime creation of `FixedBufPool` and `FixedBufRegistry`.

### DIFF
--- a/src/buf/fixed/pool.rs
+++ b/src/buf/fixed/pool.rs
@@ -1,7 +1,6 @@
 use super::handle::CheckedOutBuf;
 use super::{FixedBuf, FixedBuffers};
 
-use crate::runtime::driver::WeakHandle;
 use crate::runtime::CONTEXT;
 use libc::{iovec, UIO_MAXIOV};
 use std::cell::RefCell;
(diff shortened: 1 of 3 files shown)


FrankReh (approved):
This is a good cleanup. I would merge but if @Noah-Kennedy wants to take a look too, he should. I don't believe this would get in the way of anything Noah is doing for multi-thread support.

Noah-Kennedy (approved):
I'm still not entirely happy with these semantics, but we can't do anything better without a thread-safe internal handle that can be used for this sort of stuff. Still, this is definitely better than what we just had.

At some point I want to figure out a thread-safe handle for the io driver so that we don't have the footgun with leaking kernel resources.

PR #202, opened 2022-12-22
Replace squeue::Entry with squeue::Entry128, Add UringCmd16 and UringCmd80

thomasbarrett:
This is definitely not in a ready-to-merge state, but I would like to open a conversation about integrating squeue::Entry128, and io_uring methods that require squeue::Entry128, into tokio-uring. For context, I am working on implementing a rust-native ublk server, and would like to make use of tokio-uring.


src/fs/mod.rs
```diff
@@ -11,3 +11,4 @@ pub use file::File;
 
 mod open_options;
 pub use open_options::OpenOptions;
+mod cmd;
```
FrankReh, line 14: Missing newline.

FrankReh (commented):
No comments. No unit tests. What kind of cloud machine could this UringCmd16 be tested on? Is there something in qemu that would simulate such a device too?

thomasbarrett:
I have been testing this on a Linux 6.1 QEMU VM on the “/dev/ublk-control” device created by the ublk module.

in general, IORING_OP_URING_CMD is used for device specific operations, very similar to ioctl. In addition to ublk, I believe that NVMe devices and /dev/null also support IORING_OP_URING_CMD. I will add some tests using /dev/null since it doesn’t require any specific devices.

FrankReh:
I have a VM with an NVMe root partition but I don't know if that can be used for these URING_CMD operations. I'll have to read some more soon.

FrankReh:
Do the feature flags mean one can't have both forms of sqe and cqe in the same app? I always thought we would end up with the main driver being the smaller version and when a user wanted the larger version, there would be a secondary ring created but I believe Noah was leaning towards supporting multiple threads, each with their own uring potentially. The uring interface doesn't require all urings from one app be the same size so we probably shouldn't be enforcing something like that either. But @Noah-Kennedy needs to get involved and say what he wants in this feature.

thomasbarrett:
> I have a VM with an NVMe root partition but I don't know if that can be used for these URING_CMD operations. I'll have to read some more soon.

Hmm, you I have not personally played around with uring-cmd + NVMe drives, but [https://github.com/espoal/uring_examples/blob/main/examples/nvme/src/main.rs](this) seems to be a good example for anyone interested. You could probably get away with some read-only operations to your root partition... but you'd be playing a dangerous game.

FrankReh:
Thanks for the link. I'll figure out why it didn't work in the morning. Also, my VM that is supposed to have an NVMe root doesn't show up as NVMe so I admit I may have spoken too soon. Maybe through virtio it doesn't give the kernel an NVMe interface anyway.

thomasbarrett:
> Thanks for the link. I'll figure out why it didn't work in the morning. Also, my VM that is supposed to have an NVMe root doesn't show up as NVMe so I admit I may have spoken too soon. Maybe through virtio it doesn't give the kernel an NVMe interface anyway.

If you are using plain QEMU, I have personally found [this](https://qemu-project.gitlab.io/qemu/system/devices/nvme.html) feature good for emulating NVMe devices.

Noah-Kennedy:
Why is there a binary file?

FrankReh:
> Why is there a binary file?

Good catch. I totally missed that when looking through the diffs.

thomasbarrett:
Good catch @Noah-Kennedy! Removed. Perhaps a cargo cross artifact? Struggling to run CI pipeline on macOS 😄

FrankReh:
Why is the command behind a feature flag? We have other commands that don't work in older kernels.

FrankReh (changes requested):
The code itself looks good.

Why is uring_cmd16 behind a feature flag? And @Noah-Kennedy will have to weigh in on the use of the other feature flag for the other ring entry sizes. I think the crate should support both forms at one time.

thomasbarrett:
Good point, the `uring-cmd` operation doesn't need its own feature flag... the `sqe128` flag should suffice. As for the `sqe128` and `cqe32` flags... I will take another look at adding them as generic arguments.

thomasbarrett:
Hmm @FrankReh ... so I took a look at making the runtime generic w.r.t SQE and CQE size. The current holdup is that the thread_local CONTEXT variable cannot be made generic. I don't believe that there is any zero-cost way to make the runtime generic. Rather than add a bunch of boxed traits, the best move IMO is probably just to keep SQE and CQE size as a feature flag.

thomasbarrett:
Also, the uring_cmd tests will fail unless run on a sufficiently recent version of Linux. What is the established standard way to include version specific tests if not feature flags?

FrankReh:
> Also, the uring_cmd tests will fail unless run on a sufficiently recent version of Linux. What is the established standard way to include version specific tests if not feature flags?

I guess the test needs to start by checking the feature or command is supported at runtime and quit the test gracefully if the most rudimentary aspect fails or can't be supported. Maybe leave a comment in the test. I had to put safeguards into the io-uring crate for this but don't recall having to do that for tokio-uring yet. Whatever you come up with is fine.

FrankReh:
@Noah-Kennedy Once this passes all CI tests, even if it means the tests were commented out, I'm inclined to approve something like this as it doesn't appear to change what any of us are currently doing and for the those that want the new feature, they can have it immediately - and maybe with a little experience, someone will figure out a good way of having both rings - I know other rust project(s) have already done that.

Also this is a good time to mention that something going in now might be changed for a more general solution down the road. And it's too early to be holding anyone to benchmark results from one release to the next.

thomasbarrett:
Ugh sorry... should have checked all of the CI after making my changes. Should be good now.

src/runtime/driver/mod.rs
```diff
@@ -343,7 +355,15 @@ impl Drop for Driver {
...
                         .submission()
-                        .push(&AsyncCancel::new(id as u64).build().user_data(u64::MAX))
+                        .push(
+                            // Conversion to configured squeue::EntryMarker is useless when
+                            // sqe128 feature is disabled.
+                            #[allow(clippy::useless_conversion)]
+                            &AsyncCancel::new(id as u64)
+                                .build()
+                                .user_data(u64::MAX)
+                                .into(),
```
FrankReh, line 365: The comment above helps, but I still don't know what this into() is converting to. I could clone the branch tomorrow and figure it out I guess.

src/runtime/driver/op/mod.rs
```diff
@@ -90,6 +90,19 @@ impl From<cqueue::Entry> for CqeResult {
     }
 }
 
+impl From<cqueue::Entry32> for CqeResult {
```
FrankReh, line 93: I have to clone this tomorrow to figure out what this block is for too.

FrankReh (changes requested):
Well, it looks good, without my understanding every aspect of it this evening. I didn't seem to be changing anything for those who aren't enabling the feature flags. So there aren't really any changes being asked for. I guess there could have been a comment block or two about how/why the features exist and something about passing in byte arrays as commands but that's more for the novice person who wouldn't know what to do with these commands - so maybe not that useful. For someone wanting to pass these arrays of bytes through the uring to a device, I imagine it would be pretty obvious.

So I'll fork the branch tomorrow and play around with the compile to get a better feel but I don't have a device to test it on I think. But I think it could go in, with the warning that future versions of the crate might do away with the feature and provide a second ring of this size or sizes at runtime when asked for. It's just that we don't have a good mechanism for using one ring to support other rings at the moment but I think we should.

jiangliu:
Thanks for your great work:)
We are planning to implement a ulbk server in rust, and this definitely help us much!

FrankReh:
I'm going to leave this for @Noah-Kennedy to comment on. It's not the direction I was hoping for but also it's not next in line for me to figure out. The project is at an interesting point. The APIs are still in flux and there's no corporate sponsorship of course. Do we put something in that we most likely will change later or do we hold off putting anything about it in until we think we know what we want in the end.

thomasbarrett:
> Thanks for your great work:)
> We are planning to implement a ulbk server in rust, and this definitely help us much!

Very cool! I actually have an early stage (but working) ublk server implementation that I will probably make public in the next day or so. If this is for an open source project, I would be happy to contribute.

thomasbarrett:
Feel free to close out this MR. It turns out that I need a bit more manual control then `tokio-uring` provides at the moment. I am going [this](https://github.com/thomasbarrett/io-uring-async) project moving forward. It essentially takes the guts of `tokio-uring` and creates a IoUringAsync device that can run directly on top of normal single-threaded Tokio runtime. It could be interesting to integrate something like this into `tokio-uring` in the future. Let me know what you think.

Noah-Kennedy:
@thomasbarrett what additional manual control do you need? We're working on providing more manual control of a number of things to users.

thomasbarrett:
Hey @Noah-Kennedy. It would be great if 3rd party libraries had the ability to send manual io-uring commands to the Runtime. This would allow libraries to be built w/ support for io-uring features that are not yet supported by this runtime. I have started using this project for my own work (it essentially just wraps the runtime portion of tokio-uring). I have started using this repository for my own work, but I would love to contribute some of these ideas back into tokio-uring.

https://github.com/thomasbarrett/io-uring-async

PR #203, opened 2022-12-29
Update docs to conform to Tokio standards

rrichardson:
This  PR does ~3~ 2 things: 
~1. Alters the Runtime crate::Builder API slightly to make it easier to correctly set sq and cq entry counts.~
2. Updates method documentation to use indicative mood.  This is not a huge change, as most public method docs were conformant.
3. Enhances other documentation in a few places for clarity.

### DIFF
--- a/src/fs/directory.rs
+++ b/src/fs/directory.rs
@@ -3,12 +3,17 @@ use std::io;
 use std::path::Path;
 
 /// Creates a directory on the local filesystem.
+///
 /// Returns a future which resolves to unit on success, and ['std::io::Error'] on failure.
 ///
-/// Preconditions:
-/// 1. The specified directory must not already exist.
-/// 2. The parent directory must exist.
-/// 3. The user must have sufficient permissions to create the directory on the filesystem.
+/// # Errors
+///
+/// This function will return an error in the following situations, but is not
+/// limited to just these cases:
+///
+/// * User lacks permissions to create a directory at `path`.
+/// * A parent of the given path doesn't exist.
+/// * `path` already exists.
 ///
 /// # Examples
 ///
@@ -28,12 +33,18 @@ pub async fn create_dir<P: AsRef<Path>>(path: P) -> io::Result<()> {
 }
 
 /// Removes a directory on the local filesystem.
+///
 /// Returns a future which resolves to unit on success, and ['std::io::Error'] on failure.
 ///
-/// Preconditions:
-/// 1. The specified directory must exist.
-/// 2. The specified directory must be empty.
-/// 3. The user must have sufficient permissions to modify/remove a directory on the filesystem.
+/// # Errors
+///
+/// This function will return an error in the following situations, but is not
+/// limited to just these cases:
+///
+/// * `path` doesn't exist.
+/// * `path` isn't a directory.
+/// * The user lacks permissions to modify/remove the directory at the provided `path`.
+/// * The directory isn't empty.
 ///
 /// # Examples
 ///
--- a/src/fs/directory.rs
+++ b/src/fs/directory.rs
@@ -4,17 +4,19 @@ use std::path::Path;
 
 /// Creates a directory on the local filesystem.
 ///
-/// Returns unit on success, and ['std::io::Error'] on failure.
-///
 /// # Errors
 ///
 /// This function will return an error in the following situations, but is not
 /// limited to just these cases:
 ///
-/// * User lacks permissions to create a directory at `path`.
+/// * User lacks permissions to create a directory at `path`
+///      * [`io::ErrorKind`] would be set to `PermissionDenied`
 /// * A parent of the given path doesn't exist.
+///      * [`io::ErrorKind`] would be set to `NotFound` or `NotADirectory`
 /// * `path` already exists.
+///      * [`io::ErrorKind`] would be set to `AlreadyExists`
 ///
+/// [`ErrorKind`]: std::io::ErrorKind
 /// # Examples
 ///
 /// ```no_run
@@ -34,7 +36,11 @@ pub async fn create_dir<P: AsRef<Path>>(path: P) -> io::Result<()> {
 
 /// Removes a directory on the local filesystem.
 ///
-/// Returns unit on success, and ['std::io::Error'] on failure.
+/// This will only remove empty directories with no children. If you want to destroy the entire
+/// contents of a directory, you may try [`remove_dir_all`] which uses the standard Tokio executor.
+/// There currently is no implementation of `remove_dir_all` in tokio-uring.
+///
+/// [`remove_dir_all`]: https://docs.rs/tokio/latest/tokio/fs/fn.remove_dir_all.html
 ///
 /// # Errors
 ///
@@ -42,9 +48,13 @@ pub async fn create_dir<P: AsRef<Path>>(path: P) -> io::Result<()> {
 /// limited to just these cases:
 ///
 /// * `path` doesn't exist.
+///      * [`io::ErrorKind`] would be set to `NotFound`
 /// * `path` isn't a directory.
+///      * [`io::ErrorKind`] would be set to `NotADirectory`
 /// * The user lacks permissions to modify/remove the directory at the provided `path`.
+///      * [`io::ErrorKind`] would be set to `PermissionDenied`
 /// * The directory isn't empty.
+///      * [`io::ErrorKind`] would be set to `DirectoryNotEmpty`
 ///
 /// # Examples
 ///
--- a/src/fs/directory.rs
+++ b/src/fs/directory.rs
@@ -4,7 +4,7 @@ use std::path::Path;
 
 /// Creates a directory on the local filesystem.
 ///
-/// Returns a future which resolves to unit on success, and ['std::io::Error'] on failure.
+/// Returns unit on success, and ['std::io::Error'] on failure.
 ///
 /// # Errors
 ///
@@ -34,7 +34,7 @@ pub async fn create_dir<P: AsRef<Path>>(path: P) -> io::Result<()> {
 
 /// Removes a directory on the local filesystem.
 ///
-/// Returns a future which resolves to unit on success, and ['std::io::Error'] on failure.
+/// Returns unit on success, and ['std::io::Error'] on failure.
 ///
 /// # Errors
 ///
--- a/src/fs/directory.rs
+++ b/src/fs/directory.rs
@@ -2,6 +2,14 @@ use crate::runtime::driver::op::Op;
 use std::io;
 use std::path::Path;
 
+/// Creates a directory on the local filesystem.
+/// Returns a future which resolves to unit on success, and ['std::io::Error'] on failure.
+///
+/// Preconditions:
+/// 1. The specified directory must not already exist.
+/// 2. The parent directory must exist.
+/// 3. The user must have sufficient permissions to create the directory on the filesystem.
+///
 /// # Examples
 ///
 /// ```no_run
@@ -19,7 +27,13 @@ pub async fn create_dir<P: AsRef<Path>>(path: P) -> io::Result<()> {
     Op::make_dir(path.as_ref())?.await
 }
 
-/// Removes an empty directory.
+/// Removes a directory on the local filesystem.
+/// Returns a future which resolves to unit on success, and ['std::io::Error'] on failure.
+///
+/// Preconditions:
+/// 1. The specified directory must exist.
+/// 2. The specified directory must be empty.
+/// 3. The user must have sufficient permissions to modify/remove a directory on the filesystem.
 ///
 /// # Examples
 ///
--- a/src/fs/file.rs
+++ b/src/fs/file.rs
@@ -905,6 +905,7 @@ pub async fn remove_file<P: AsRef<Path>>(path: P) -> io::Result<()> {
 /// Renames a file or directory to a new name, replacing the original file if
 /// `to` already exists.
 ///
+/// #Errors
 ///
 /// * `path` doesn't exist.
 ///      * [`io::ErrorKind`] would be set to `NotFound`
--- a/src/fs/file.rs
+++ b/src/fs/file.rs
@@ -877,6 +877,14 @@ impl fmt::Debug for File {
 
 /// Removes a File
 ///
+/// This function will return an error in the following situations, but is not
+/// limited to just these cases:
+///
+/// * `path` doesn't exist.
+///      * [`io::ErrorKind`] would be set to `NotFound`
+/// * The user lacks permissions to modify/remove the file at the provided `path`.
+///      * [`io::ErrorKind`] would be set to `PermissionDenied`
+///
 /// # Examples
 ///
 /// ```no_run
@@ -897,7 +905,13 @@ pub async fn remove_file<P: AsRef<Path>>(path: P) -> io::Result<()> {
 /// Renames a file or directory to a new name, replacing the original file if
 /// `to` already exists.
 ///
-/// This will not work if the new name is on a different mount point.
+///
+/// * `path` doesn't exist.
+///      * [`io::ErrorKind`] would be set to `NotFound`
+/// * The user lacks permissions to modify/remove the file at the provided `path`.
+///      * [`io::ErrorKind`] would be set to `PermissionDenied`
+/// * The new name/path is on a different mount point.
+///      * [`io::ErrorKind`] would be set to `CrossesDevices`
 ///
 /// # Example
 ///
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -191,9 +191,6 @@ impl Builder {
     /// Replaces the default [`io_uring::Builder`], which controls the settings for the
     /// inner `io_uring` API.
     ///
-    /// This interface offers many more settings and parameters,
-    /// but requires directly depending upon the `io_uring` crate.
-    ///
     /// Refer to the [`io_uring::Builder`] documentation for all the supported methods.
     pub fn uring_builder(&mut self, b: &io_uring::Builder) -> &mut Self {
         self.urb = b.clone();
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -84,7 +84,7 @@ pub use runtime::Runtime;
 use crate::runtime::driver::op::Op;
 use std::future::Future;
 
-/// Start an `io_uring` enabled Tokio runtime.
+/// Starts an `io_uring` enabled Tokio runtime.
 ///
 /// All `tokio-uring` resource types must be used from within the context of a
 /// runtime. The `start` method initializes the runtime and runs it for the
@@ -146,7 +146,7 @@ pub fn start<F: Future>(future: F) -> F::Output {
     rt.block_on(future)
 }
 
-/// Create and return an io_uring::Builder that can then be modified
+/// Creates and returns an io_uring::Builder that can then be modified
 /// through its implementation methods.
 ///
 /// This function is provided to avoid requiring the user of this crate from
@@ -156,50 +156,67 @@ pub fn uring_builder() -> io_uring::Builder {
     io_uring::IoUring::builder()
 }
 
-/// Builder API to allow starting the runtime and creating the io_uring driver with non-default
-/// parameters.
+/// Builder API that can create and start the `io_uring` runtime with non-default parameters,
+/// while abstracting away the underlying io_uring crate.
 // #[derive(Clone, Default)]
 pub struct Builder {
-    entries: u32,
+    sq_entries: u32,
+    cq_entries: u32,
     urb: io_uring::Builder,
 }
 
-/// Return a Builder to allow setting parameters before calling the start method.
-/// Returns a Builder with our default values, all of which can be replaced with the methods below.
+/// Constructs a [`Builder`] with default settings.
+/// Use this to alter submission and completion queue parameters, and to `start` the io_uring runtime.
 ///
-/// Refer to Builder::start for an example.
+/// Refer to [`Builder::start`] for an example.
 pub fn builder() -> Builder {
     Builder {
-        entries: 256,
+        sq_entries: 256,
+        cq_entries: 512,
         urb: io_uring::IoUring::builder(),
     }
 }
 
 impl Builder {
-    /// Set number of submission queue entries in uring.
+    /// Sets the number of submission queue entries in uring. The default is 256.
     ///
-    /// The kernel will ensure it uses a power of two and will round this up if necessary.
-    /// The kernel requires the number of completion queue entries to be larger than
-    /// the submission queue entries so generally will double the sq entries count.
-    ///
-    /// The caller can specify even a larger cq entries count by using the uring_builder
-    /// as shown in the start example below.
-    pub fn entries(&mut self, e: u32) -> &mut Self {
-        self.entries = e;
+    /// The kernel requires the number of submission queue entries to be a power of two,
+    /// and that it be less than the number of completion queue entries.
+    /// This function will adjust the `cq_entries` value to be at least 2 times `sq_entries`
+    pub fn sq_entries(&mut self, sqe: u32) -> &mut Self {
+        self.sq_entries = sqe.next_power_of_two();
+        if self.cq_entries / 2 < self.sq_entries {
+            self.cq_entries = self.sq_entries * 2;
+        }
+        self.urb.setup_cqsize(self.cq_entries);
         self
     }
 
-    /// Replace the default io_uring Builder. This allows the caller to craft the io_uring Builder
-    /// using the io_uring crate's Builder API.
+    /// Sets the number of completion queue entries in uring. The default is 512.
     ///
-    /// Refer to the Builder start method for an example.
-    /// Refer to the io_uring::builder documentation for all the supported methods.
+    /// The kernel requires the number of completion queue entries to be a power of two,
+    /// and that it be greater than the number of submission queue entries.
+    /// This function will adjust the `cq_entries` value to be at least 2 times `sq_entries`
+    pub fn cq_entries(&mut self, cqe: u32) -> &mut Self {
+        self.cq_entries = cqe.next_power_of_two();
+        if (self.cq_entries / 2) < self.sq_entries {
+            self.sq_entries = self.cq_entries / 2;
+        }
+        self.urb.setup_cqsize(self.cq_entries);
+        self
+    }
+
+    /// Replaces the default [`io_uring::Builder`], which controls the settings for the
+    /// inner `io_uring` API. This interface offers many more settings and parameters,
+    /// but requires directly depending upon the `io_uring` crate.
+    ///
+    /// Refer to the [`io_uring::builder`] documentation for all the supported methods.
     pub fn uring_builder(&mut self, b: &io_uring::Builder) -> &mut Self {
         self.urb = b.clone();
         self
     }
 
-    /// Start an `io_uring` enabled Tokio runtime.
+    /// Starts an `io_uring` enabled Tokio runtime.
     ///
     /// # Examples
     ///
@@ -211,10 +228,8 @@ impl Builder {
     ///
     /// fn main() -> Result<(), Box<dyn std::error::Error>> {
     ///     tokio_uring::builder()
-    ///         .entries(64)
-    ///         .uring_builder(tokio_uring::uring_builder()
-    ///             .setup_cqsize(1024)
-    ///             )
+    ///         .sq_entries(64)
+    ///         .cq_entries(1024)
     ///         .start(async {
     ///             let listener = TcpListener::bind("127.0.0.1:8080").await?;
     ///
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -166,7 +166,8 @@ pub struct Builder {
 
 /// Constructs a [`Builder`] with default settings.
 ///
-/// Use this to alter submission and completion queue parameters, and to `start` the io_uring runtime.
+/// Use this to alter submission and completion queue parameters, and to create the io_uring
+/// Runtime.
 ///
 /// Refer to [`Builder::start`] for an example.
 pub fn builder() -> Builder {
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -160,57 +160,41 @@ pub fn uring_builder() -> io_uring::Builder {
 /// while abstracting away the underlying io_uring crate.
 // #[derive(Clone, Default)]
 pub struct Builder {
-    sq_entries: u32,
-    cq_entries: u32,
+    entries: u32,
     urb: io_uring::Builder,
 }
 
 /// Constructs a [`Builder`] with default settings.
+///
 /// Use this to alter submission and completion queue parameters, and to `start` the io_uring runtime.
 ///
 /// Refer to [`Builder::start`] for an example.
 pub fn builder() -> Builder {
     Builder {
-        sq_entries: 256,
-        cq_entries: 512,
+        entries: 256,
         urb: io_uring::IoUring::builder(),
     }
 }
 
 impl Builder {
-    /// Sets the number of submission queue entries in uring. The default is 256.
+    /// Sets the number of Submission Queue entries in uring.
     ///
+    /// The default value is 256.
     /// The kernel requires the number of submission queue entries to be a power of two,
     /// and that it be less than the number of completion queue entries.
     /// This function will adjust the `cq_entries` value to be at least 2 times `sq_entries`
-    pub fn sq_entries(&mut self, sqe: u32) -> &mut Self {
-        self.sq_entries = sqe.next_power_of_two();
-        if self.cq_entries / 2 < self.sq_entries {
-            self.cq_entries = self.sq_entries * 2;
-        }
-        self.urb.setup_cqsize(self.cq_entries);
-        self
-    }
-
-    /// Sets the number of completion queue entries in uring. The default is 512.
-    ///
-    /// The kernel requires the number of completion queue entries to be a power of two,
-    /// and that it be greater than the number of submission queue entries.
-    /// This function will adjust the `cq_entries` value to be at least 2 times `sq_entries`
-    pub fn cq_entries(&mut self, cqe: u32) -> &mut Self {
-        self.cq_entries = cqe.next_power_of_two();
-        if (self.cq_entries / 2) < self.sq_entries {
-            self.sq_entries = self.cq_entries / 2;
-        }
-        self.urb.setup_cqsize(self.cq_entries);
+    pub fn entries(&mut self, sqe: u32) -> &mut Self {
+        self.entries = sqe;
         self
     }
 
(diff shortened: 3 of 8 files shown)


src/fs/directory.rs
```diff
@@ -2,6 +2,14 @@ use crate::runtime::driver::op::Op;
 use std::io;
 use std::path::Path;
 
+/// Creates a directory on the local filesystem.
+/// Returns a future which resolves to unit on success, and ['std::io::Error'] on failure.
+///
+/// Preconditions:
```
Noah-Kennedy, line 8: I'd generally prefer indicating these as modes of failure (see the docs in tokio)

src/fs/directory.rs
```diff
@@ -19,7 +27,13 @@ pub async fn create_dir<P: AsRef<Path>>(path: P) -> io::Result<()> {
     Op::make_dir(path.as_ref())?.await
 }
 
-/// Removes an empty directory.
+/// Removes a directory on the local filesystem.
+/// Returns a future which resolves to unit on success, and ['std::io::Error'] on failure.
+///
+/// Preconditions:
```
Noah-Kennedy, line 33: I'd generally prefer indicating these as modes of failure (see the docs in tokio)

src/runtime/mod.rs
```diff
@@ -13,7 +13,9 @@ thread_local! {
     pub(crate) static CONTEXT: RuntimeContext = RuntimeContext::new();
 }
 
-/// The Runtime executor
+/// The Runtime Executor
```
Noah-Kennedy, line 16: Add a whitespace line in-between. Rustdoc relies on this line for how it formats the docs.

src/fs/directory.rs
```diff
@@ -19,7 +27,13 @@ pub async fn create_dir<P: AsRef<Path>>(path: P) -> io::Result<()> {
     Op::make_dir(path.as_ref())?.await
 }
 
-/// Removes an empty directory.
+/// Removes a directory on the local filesystem.
```
Noah-Kennedy, line 30: Add a line of whitespace between the initial overview line and the body.

src/fs/directory.rs
```diff
@@ -2,6 +2,14 @@ use crate::runtime::driver::op::Op;
 use std::io;
 use std::path::Path;
 
+/// Creates a directory on the local filesystem.
```
Noah-Kennedy, line 5: Add a line of whitespace between the initial overview line and the body.

src/lib.rs
```diff
@@ -156,50 +156,67 @@ pub fn uring_builder() -> io_uring::Builder {
...
 pub struct Builder {
-    entries: u32,
+    sq_entries: u32,
+    cq_entries: u32,
     urb: io_uring::Builder,
 }
 
-/// Return a Builder to allow setting parameters before calling the start method.
-/// Returns a Builder with our default values, all of which can be replaced with the methods below.
+/// Constructs a [`Builder`] with default settings.
```
Noah-Kennedy, line 168: Add a whitespace line.

src/net/udp.rs
```diff
@@ -92,6 +92,8 @@ pub struct UdpSocket {
 
 impl UdpSocket {
     /// Creates a new UDP socket and attempt to bind it to the addr provided.
```
Noah-Kennedy, line 94: Add a newline in between

src/runtime/mod.rs
```diff
@@ -56,7 +58,8 @@ pub fn spawn<T: Future + 'static>(task: T) -> tokio::task::JoinHandle<T::Output>
 }
 
 impl Runtime {
-    /// Create a new tokio_uring runtime on the current thread
+    /// Creates a new tokio_uring runtime on the current thread
```
Noah-Kennedy, line 61: Add a period and a new line.

src/runtime/mod.rs
```diff
@@ -83,7 +86,13 @@ impl Runtime {
         })
     }
 
-    /// Runs a future to completion on the current runtime
+    /// Runs a future to completion on the current runtime.
```
Noah-Kennedy, line 89: Add a newline.

Noah-Kennedy (changes requested):
I left some feedback, but overall this is a great improvement!

Noah-Kennedy:
I'd strongly recommend using `cargo doc` to look at how any changes in the docs would render.

src/lib.rs
```diff
@@ -156,50 +156,67 @@ pub fn uring_builder() -> io_uring::Builder {
...
 
-/// Builder API to allow starting the runtime and creating the io_uring driver with non-default
-/// parameters.
+/// Builder API that can create and start the `io_uring` runtime with non-default parameters,
+/// while abstracting away the underlying io_uring crate.
 // #[derive(Clone, Default)]
 pub struct Builder {
-    entries: u32,
+    sq_entries: u32,
+    cq_entries: u32,
```
FrankReh, line 164: Sorry, I can't get behind adding a cq entry here.

src/lib.rs
```diff
@@ -156,50 +156,67 @@ pub fn uring_builder() -> io_uring::Builder {
...
+/// Constructs a [`Builder`] with default settings.
+/// Use this to alter submission and completion queue parameters, and to `start` the io_uring runtime.
 ///
-/// Refer to Builder::start for an example.
+/// Refer to [`Builder::start`] for an example.
 pub fn builder() -> Builder {
     Builder {
-        entries: 256,
+        sq_entries: 256,
+        cq_entries: 512,
```
FrankReh, line 175: I also can't get behind defining a default here.

src/lib.rs
```diff
@@ -156,50 +156,67 @@ pub fn uring_builder() -> io_uring::Builder {
...
-    /// the submission queue entries so generally will double the sq entries count.
+    /// The kernel requires the number of submission queue entries to be a power of two,
+    /// and that it be less than the number of completion queue entries.
+    /// This function will adjust the `cq_entries` value to be at least 2 times `sq_entries`
+    pub fn sq_entries(&mut self, sqe: u32) -> &mut Self {
+        self.sq_entries = sqe.next_power_of_two();
+        if self.cq_entries / 2 < self.sq_entries {
+            self.cq_entries = self.sq_entries * 2;
+        }
+        self.urb.setup_cqsize(self.cq_entries);
```
FrankReh, line 191: I also can't get behind using this value to set a value in the default urb here. But maybe this is fixing something that was broken and I'm mistaken?

src/lib.rs
```diff
@@ -156,50 +156,67 @@ pub fn uring_builder() -> io_uring::Builder {
...
-        self.entries = e;
+    /// The kernel requires the number of completion queue entries to be a power of two,
+    /// and that it be greater than the number of submission queue entries.
+    /// This function will adjust the `cq_entries` value to be at least 2 times `sq_entries`
+    pub fn cq_entries(&mut self, cqe: u32) -> &mut Self {
+        self.cq_entries = cqe.next_power_of_two();
+        if (self.cq_entries / 2) < self.sq_entries {
+            self.sq_entries = self.cq_entries / 2;
+        }
+        self.urb.setup_cqsize(self.cq_entries);
```
FrankReh, line 205: Setting the value in the urb seems odd.

src/lib.rs
```diff
@@ -156,50 +156,67 @@ pub fn uring_builder() -> io_uring::Builder {
...
-    /// Refer to the Builder start method for an example.
-    /// Refer to the io_uring::builder documentation for all the supported methods.
+    /// Refer to the [`io_uring::builder`] documentation for all the supported methods.
     pub fn uring_builder(&mut self, b: &io_uring::Builder) -> &mut Self {
         self.urb = b.clone();
         self
     }
 
-    /// Start an `io_uring` enabled Tokio runtime.
+    /// Starts an `io_uring` enabled Tokio runtime.
```
FrankReh, line 219: above on line 215, the self.urb gets replaced with a b.clone(), even if the user had specified the other cq parameter first. The proposed api looks easier but without more explanation, it can go wrong.

FrankReh (changes requested):
Generally I'm not for PRs that mix comments with changes to an API. Why not propose the comment improvements as a separate PR?

Noah-Kennedy:
@FrankReh that's a good point, maybe this should be narrowed to only doc changes, and API changes can go in another PR.

rrichardson:
> Generally I'm not for PRs that mix comments with changes to an API. Why not propose the comment improvements as a separate PR? 

I am with you on this.  I started off working only on documentation, but I couldn't get past the weird/inconsistent settings for sq and cq entries. 

I agree that the implementation is a bit weird, it is a side effect of (sometimes) transparently using the `urb` for the actual initialization.  I'm not super happy with it either.  

As you suggest,  I will submit the doc changes as a PR. 
Regarding the code changes:   I think the Builder API(s) need(s) some refactoring to make them more ergonomic and foolproof, so it will probably require some design discussion.  I'll create a ticket for discussion.

Noah-Kennedy:
It's probably best to leave this open as a PR for the doc changes, due to the feedback already present.

src/lib.rs
rrichardson: It wasn't broken per-se, but the crate::Builder is just a thin wrapper over the io_uring::Builder, and io_uring::Builder.build() takes 1 parameter, the SQ Entries parameter.  It doesn't offer any other interface to set SQ entries size, but it does offer a method to set the CQ Entries size.  

Also, the Builders are used on start() in a couple different ways, so I wanted to be sure that the cq entries setting was saved before  io_uring::Builder.build() is called. 

What I think needs to happen is we need to refactor the io_uring::Builder API a bit, because there is no foolproof way to use it. 


https://github.com/rrichardson/tokio-uring/blob/47342a872fb90ea82236f9a506bdafc67036b424/src/runtime/driver/mod.rs#L40-L41

src/lib.rs
rrichardson: I added the cq entry because I couldn't read the values in the inner`io_uring::Builder`. They're write-only.  
However, I wanted to be able to enforce the invariants regarding SQ and CQ entries.

rrichardson:
> I'd strongly recommend using cargo doc to look at how any changes in the docs would render. 

I typically follow this workflow, however, in this case, it is not terribly convenient.  My work computer is a macbook, and my dev box is a corporate managed Linux VM with limited access.  I'll see if I can come up with a convenient way to sync the docs from my dev box (I need to create a static webserver for documentation anyways)

FrankReh:
This PR would be good to use for the comments. The title or description can be changed to just reference improving comments when it is squash-merged.

src/runtime/mod.rs
```diff
@@ -83,7 +88,13 @@ impl Runtime {
...
+    ///
+    /// Returns a Future that resolves to the output of the
+    /// supplied future.
+    ///
+    /// Note: Unlike most [`Futures`](std::future::Future), this future will
```
FrankReh, line 96: This `Note` also strikes me as kind of opposite the kind of advice `Tokio` would give about its `block_on`. It should be understood that calling `block_on` is a task operation.

Also, which `this future` is being referred to? If such a note really is worthwhile, it should probably leave no room for doubt.

In fact, if we want to clean this public function documentation, it should be written as carefully as the `Tokio runtime block_on`.  Or just reference it and say what our wrapping of it accomplishes. https://github.com/tokio-rs/tokio/blob/048049f8883f1bc73ee5d3fa3c94e61cf41d1e38/tokio/src/runtime/runtime.rs#L223-L252

FrankReh (changes requested):
Lots of very good improvements to the comments, thank you.

Sorry, I know my comments about comments sound harsh today. Not having a quiet/let-the-chips-fall-where-they-may-day I guess.

Also, I know the interface for setting the io-uring builder isn't idiomatic. I wasn't aware of any problems with it though. And certainly before ever getting to 1.0, we would probably just copy all the io-uring builder methods and pass them through. I expected the io-uring builder API to still be changing frequently so didn't want to duplicate the efforts yet. Maybe the uring API isn't changing so much anymore these days either.

rrichardson:
Ok.  I've removed the API changes, and also found a few more spots where the docs needed updating.  

Also, I've created https://github.com/tokio-rs/tokio-uring/issues/204  to discuss changes to the Builder API

src/lib.rs
```diff
@@ -176,30 +177,30 @@ pub fn builder() -> Builder {
...
         self
     }
 
-    /// Replace the default io_uring Builder. This allows the caller to craft the io_uring Builder
-    /// using the io_uring crate's Builder API.
+    /// Replaces the default [`io_uring::Builder`], which controls the settings for the
+    /// inner `io_uring` API.
+    ///
+    /// This interface offers many more settings and parameters,
+    /// but requires directly depending upon the `io_uring` crate.
```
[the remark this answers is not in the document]
rrichardson: There is the `uring_builder` method which does return io_uring::Builder.  I am assuming that io_uring is re-exported then.  

With that in mind, maybe the right thing to do is to eliminate the tokio_uring::Builder method since it isn't really a Builder, but a convenience wrapper for supplying the SQ Entries parameter to `io_uring::Builder::build()`

We can discuss that in #204

src/net/udp.rs
```diff
@@ -92,6 +92,9 @@ pub struct UdpSocket {
...
+    ///
+    /// Returns a future that resolves to a new instance of [`UdpSocket`] on success,
```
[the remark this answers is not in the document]
rrichardson: That is rather redundant. I also don't consistently do it in the other pieces.  Removed.

src/runtime/mod.rs
rrichardson: >  Here, I agree it is warranted that it says a future is being returned.

Actually. I'm wrong.   That function is not async.  It doesn't return a future.  

I think the most accurate documentation would be to copy paste *most of* the tokio docs you reference, (omitting the multithreaded runtime bits)

rrichardson: Looking more closely at the comments,  neither the `Current thread scheduler` or `Multi thread scheduler` sections are correct for tokio-uring. 

I'll leave them out for now, and in the future, someone with more insight into its precise behavior can comment.

src/lib.rs
```diff
@@ -156,18 +156,19 @@ pub fn uring_builder() -> io_uring::Builder {
...
     entries: u32,
     urb: io_uring::Builder,
 }
 
-/// Return a Builder to allow setting parameters before calling the start method.
-/// Returns a Builder with our default values, all of which can be replaced with the methods below.
+/// Constructs a [`Builder`] with default settings.
 ///
-/// Refer to Builder::start for an example.
+/// Use this to alter submission and completion queue parameters, and to `start` the io_uring runtime.
```
FrankReh, line 169: `and to start the io_uring runtime`. It's better than what we had before. But is it accurate? It is used to start the `tokio-uring` runtime isn't it? I like how the emphasis is on getting to `start` at the end.

FrankReh: I'll admit to being confused at the moment between our public `start` and our public `block_on`. I could go look at the code, but maybe it says something about our API or about our documentation here. At least the example of start that comes with the `start` method makes it clear although it is verbose.

src/lib.rs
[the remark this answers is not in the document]
FrankReh: yes. Two levels of builder. Not idiomatic at all.

FrankReh, line 195: I still think this comment is wrong. It doesn't depend on the `io_uring` crate. It does depend on the `io_uring` API. Maybe I'm splitting hairs. But I don't think any of the examples required pulling in the `io-uring` crate as a dependency.

src/net/udp.rs
```diff
@@ -183,19 +186,22 @@ impl UdpSocket {
...
+    ///
+    /// This enables `write` and `read` syscalls to be used on this instance.
+    /// It also constrains the `read` to receive data only from the specified remote peer.
     ///
-    /// Note that usually, a successful `connect` call does not specify
-    /// that there is a remote server listening on the port, rather, such an
-    /// error would only be detected after the first send.
+    /// Note: UDP is connectionless, so a successful `connect` call does not execute
+    /// a handshake or validation of the remote peer of any kind.
+    /// Any errors would not be detected until the first send.
```
FrankReh, line 196: These comments are very good.

src/net/udp.rs
```diff
@@ -183,19 +186,22 @@ impl UdpSocket {
...
     pub async fn connect(&self, socket_addr: SocketAddr) -> io::Result<()> {
         self.inner.connect(SockAddr::from(socket_addr)).await
     }
 
-    /// Sends data on the socket to the given address. On success, returns the
-    /// number of bytes written.
+    /// Sends data on the socket to the given address.
+    ///
+    /// On success, returns the number of bytes written.
+    /// This can be used on a socket regardless of whether it has been "connected".
```
FrankReh, line 204: This aside, line 204, isn't really necessary. The `connect` method already calls out what it enables. Everything else is normal for a UDP socket.

src/net/udp.rs
```diff
@@ -253,12 +264,17 @@ impl UdpSocket {
         self.inner.read_fixed(buf).await
     }
 
-    /// Write some data to the socket from the buffer, returning the original buffer and
-    /// quantity of data written.
+    /// Writes data into the socket from the specified buffer.
+    ///
+    /// Takes ownership of the buffer for the duration of the Future.
```
FrankReh, line 269: It's true that it takes ownership of the buffer for the duration of the future. But are we going to say that for each of our ops that uses a supplied buffer? Maybe a separate commit should address a nice comment like this for each of the output methods of the library?

src/runtime/mod.rs
```diff
@@ -13,7 +13,10 @@ thread_local! {
     pub(crate) static CONTEXT: RuntimeContext = RuntimeContext::new();
 }
 
-/// The Runtime executor
+/// The Runtime Executor
+///
+/// io_uring instances are always single-threaded. Correspondingly, this runtime will
```
FrankReh, line 18: @Noah-Kennedy Is this correct? Is it over stating that `io-uring` instances are always single-threaded? I just got the impression all its calls are synchronous. And the uring interface in the kernel isn't single-threaded. In fact there is at least one parameter we can set when we know it will be single threaded as an optimization.

FrankReh (changes requested):

src/lib.rs
rrichardson: It seems that the "One True" entrypoint of tokio-uring is `Runtime::block_on`.  

`crate::start` just creates a new tokio_uring::runtime::Runtime then calls `block_on`  so it saves 1 function call. 

The special sauce of `Runtime::block_on` is that it creates an instance of the Uring driver,   and plugs it into the context.  After that it seems to be just vanilla Tokio stuff.

src/lib.rs
rrichardson: I think you're correct.   One could call methods on the value returned from `crate::uring_builder`  without having a direct dependency on the `io_uring` crate.  I don't think that any of the io_uring Builder methods require custom types.  I'll amend this.

src/net/udp.rs
rrichardson: Agreed.

src/runtime/mod.rs
rrichardson: As far as I can tell, currently the only interaction with the Tokio Uring Driver is inside of `Runtime::block_on` and it does this via `tokio::runtime::LocalSet` which is the !Send current-thread set of tasks.

src/net/udp.rs
rrichardson: I'm of two minds about this:  

1. To an experienced rust dev that is paying attention, they'd see from the signature that the  buffer is passed by value, or their rust-analyzer would certainly let them know.  So we shouldn't bother to point it out. 

-or-

2.1.   I'm an experience Rust dev, but I rarely pay that close of attention to signatures in docs.
2.2.  This is a significant departure from pretty much every other io call in the Rust ecosystem. 
2.3.  It has architectural implications.  A person has to change the design of their program (sometimes significantly) to work with buffers that are passed by value. 

For these reasons, I'm definitely leaning towards a rule: 

"We should make any IO that passes a buffer by something other than Reference explicitly state such in the docs -  UNTIL - the new set of IO traits that Nick Cameron et al. is working on has been completed and adopted."

src/fs/directory.rs
```diff
@@ -2,6 +2,19 @@ use crate::runtime::driver::op::Op;
...
+///
+/// Returns unit on success, and ['std::io::Error'] on failure.
```
Noah-Kennedy, line 7: I don't think this is worthy of documentation; it's literally the signature.

src/runtime/mod.rs
Noah-Kennedy: This runtime is single-threaded mostly to make this as simple as possible.

Noah-Kennedy (changes requested):

src/net/udp.rs
FrankReh: Fair that you see two sides to this. But I don't want to change all the docs at this time because of the complication that you see. It would mean changes to many places. Such a change would warrant its own issue and possible PR and not be mixed in with comment changes were clearly poorer english is being replaced with better english. This repo has lots of bugs and features that need addressing, and will even go through several more big rewrites I suspect, and I don't think we should be spending so much time trying to make the comments of some functions so overly descriptive. I prefer good long comments for people who are trying to make development changes to the code base at this point.

src/runtime/mod.rs
FrankReh: Anyway, the comment on line 18 is wrong. The io_uring instances are *not* always single-threaded. How the io_uring crate is used is very dependent on the opinions of the crate wants to  use it.

src/fs/directory.rs
rrichardson: Agreed.

src/fs/directory.rs
```diff
@@ -2,6 +2,19 @@ use crate::runtime::driver::op::Op;
...
+///
+/// # Errors
+///
+/// This function will return an error in the following situations, but is not
```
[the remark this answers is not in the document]
rrichardson: Done.

src/fs/directory.rs
```diff
@@ -19,7 +32,19 @@ pub async fn create_dir<P: AsRef<Path>>(path: P) -> io::Result<()> {
     Op::make_dir(path.as_ref())?.await
 }
 
-/// Removes an empty directory.
+/// Removes a directory on the local filesystem.
+///
+/// Returns unit on success, and ['std::io::Error'] on failure.
```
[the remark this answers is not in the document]
rrichardson: Agreed.

src/fs/directory.rs
```diff
@@ -19,7 +32,19 @@ pub async fn create_dir<P: AsRef<Path>>(path: P) -> io::Result<()> {
...
+///
+/// # Errors
+///
+/// This function will return an error in the following situations, but is not
```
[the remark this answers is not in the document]
rrichardson: Done.

src/fs/file.rs
```diff
@@ -845,7 +853,13 @@ pub async fn remove_file<P: AsRef<Path>>(path: P) -> io::Result<()> {
 /// Renames a file or directory to a new name, replacing the original file if
 /// `to` already exists.
 ///
-/// This will not work if the new name is on a different mount point.
+///
```
Noah-Kennedy, line 856:
```suggestion
/// #Errors
///
```

src/lib.rs
Noah-Kennedy: I'd say that this creates the runtime but doesn't start it.

Noah-Kennedy (changes requested):

src/runtime/mod.rs
rrichardson: Ok.  I have removed that line and restructured a bit.

src/net/udp.rs
rrichardson: Ok.

src/lib.rs
rrichardson: Fixed.

rrichardson:
Hopefully that's everything?

src/net/tcp/stream.rs
```diff
@@ -70,12 +70,17 @@ impl TcpStream {
         Self { inner }
     }
 
-    /// Read some data from the stream into the buffer, returning the original buffer and
-    /// quantity of data read.
+    /// Read some data from the stream into the buffer.
+    ///
+    /// Takes ownership of the buffer for the duration of the Future.
```
FrankReh, line 75: I'm not sure about this. Again, if this method has a comment that calls out this detail about an argument, are all the other methods that do the same now outdated and incorrect in their admission? But I'll leave it to you and Noah to work out.

src/net/tcp/stream.rs
```diff
@@ -70,12 +70,17 @@ impl TcpStream {
...
+    ///
+    /// Returns the original buffer and quantity of data read.
     pub async fn read<T: BoundedBufMut>(&self, buf: T) -> crate::BufResult<usize, T> {
         self.inner.read(buf).await
     }
 
+    /// Read some data from the stream into a fixed, pre-registered buffer.
```
FrankReh, line 82: I wouldn't say `fixed` in the first line where a simple/short description is intended. Where the name `fixed` comes from is a long story with its roots in the uring kernel driver itself and I think we try not to overuse that term. Better if we just say `Read some data from the stream into a registered buffer`.

src/net/tcp/stream.rs
```diff
@@ -154,6 +162,8 @@ impl TcpStream {
         self.inner.write_all(buf).await
     }
 
+    /// Writes data into the socket from a fixed, pre-registered buffer.
```
FrankReh, line 165: Same. Just `from a registered buffer`.

src/net/tcp/stream.rs
```diff
@@ -184,16 +194,17 @@ impl TcpStream {
     ///
     /// This function will return the first error that [`write_fixed`] returns.
     ///
-    /// [`write_fixed`]: Self::write
+    /// [`write_fixed`]: Self::write_fixed
```
FrankReh, line 197: Good catch.

src/net/tcp/stream.rs
```diff
@@ -184,16 +194,17 @@ impl TcpStream {
...
         T: BoundedBuf<Buf = FixedBuf>,
     {
         self.inner.write_fixed_all(buf).await
     }
 
-    /// Write data from buffers into this socket returning how many bytes were
-    /// written.
+    /// Writes data from multiple buffers into this socket using the scatter/gather IO style.
+    ///
+    /// Takes ownership of the Vec of Buffers for the duration of the Future.
```
FrankReh, line 207: same ownership comment question.

src/net/udp.rs
```diff
@@ -220,21 +226,25 @@ impl UdpSocket {
...
-    /// Read a packet of data from the socket into the buffer, returning the original buffer and
-    /// quantity of data read.
+    /// Reads a packet of data from the socket into the buffer.
+    ///
+    /// Returns the original buffer and quantity of data read.
     pub async fn read<T: BoundedBufMut>(&self, buf: T) -> crate::BufResult<usize, T> {
         self.inner.read(buf).await
     }
 
+    /// Receives a single datagram message into a fixed, pre-registered buffer.
```
FrankReh, line 246: `into a registered buffer`.

src/net/udp.rs
```diff
@@ -253,12 +263,15 @@ impl UdpSocket {
...
-    /// Write some data to the socket from the buffer, returning the original buffer and
-    /// quantity of data written.
+    /// Writes data into the socket from the specified buffer.
+    ///
+    /// Returns the original buffer and quantity of data written.
     pub async fn write<T: BoundedBuf>(&self, buf: T) -> crate::BufResult<usize, T> {
         self.inner.write(buf).await
     }
 
+    /// Writes data into the socket from a fixed, pre-registered buffer.
```
FrankReh, line 273: ditto

src/runtime/mod.rs
```diff
@@ -13,7 +13,16 @@ thread_local! {
     pub(crate) static CONTEXT: RuntimeContext = RuntimeContext::new();
 }
 
-/// The Runtime executor
+/// The Runtime Executor
+///
+/// This is the Runtime for `tokio-uring`.
+/// It wraps the default [`Runtime`], along with the uring [`Driver`]. 
+///
+/// # Multithreading
```
FrankReh, line 21: Not sure what this header in the comment is supposed to signify. Maybe instead, a line that says this version only supports the current-thread model. But again, I don't think we need too many details comments for users. The crate isn't meant to be production ready and it isn't really intended for folks who don't know what they want from urings. Plus they need to be prepared for big changes from version to version at this point.

src/runtime/mod.rs
```diff
@@ -83,7 +94,21 @@ impl Runtime {
...
 
-    /// Runs a future to completion on the current runtime
+    /// Runs a future to completion on the tokio-uring runtime. This is the
+    /// runtime's entry point.
+    ///
+    /// This runs the given future on the current thread, blocking until it is
+    /// complete, and yielding its resolved result. Any tasks, futures, or timers
+    /// which the future spawns internally will be executed on this runtime.
+    ///
+    /// Any spawned tasks will be suspended after `block_on` returns. Calling
```
FrankReh, line 104: @Noah-Kennedy needs to judge whether this statement about `block_on` is correct and whether it is correct to imply that `block_on` could be called more than once. Maybe it can and this is absolutely correct; I'm just saying I'm not sure myself

FrankReh (changes requested):
And Noah's review is needed so maybe wait on any further changes to see what his take is. Thank you.

src/runtime/mod.rs
rrichardson: It would be good to get a comment, but I based those docs on the tokio docs for block_on 
https://docs.rs/tokio/latest/tokio/runtime/struct.Runtime.html#current-thread-scheduler

src/lib.rs
```diff
@@ -176,30 +177,30 @@ pub fn builder() -> Builder {
...
-    /// using the io_uring crate's Builder API.
+    /// Replaces the default [`io_uring::Builder`], which controls the settings for the
+    /// inner `io_uring` API.
+    ///
+    /// This interface offers many more settings and parameters,
+    /// but requires directly depending upon the `io_uring` crate.
     ///
-    /// Refer to the Builder start method for an example.
-    /// Refer to the io_uring::builder documentation for all the supported methods.
+    /// Refer to the [`io_uring::Builder`] documentation for all the supported methods.
```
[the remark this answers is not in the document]
rrichardson: Errg.. I really don't want to make an example fo rthis because I'm going to put in a PR to fix the builder stuff.

src/net/tcp/stream.rs
rrichardson: Ok.. Removed all references to functions taking ownership.

src/runtime/mod.rs
FrankReh: This comment block still confuses me. It makes sense when it is in Tokio source as they have two runtimes. We don't at this time. Anyway, the pages says it is outdated, so enough here.

src/net/tcp/stream.rs
rrichardson: Changed.

src/net/udp.rs
rrichardson: Changed.

src/net/udp.rs
rrichardson: Changed.

src/net/tcp/stream.rs
```diff
@@ -70,12 +70,15 @@ impl TcpStream {
         Self { inner }
     }
 
-    /// Read some data from the stream into the buffer, returning the original buffer and
-    /// quantity of data read.
+    /// Read some data from the stream into the buffer.
+    ///
+    /// Returns the original buffer and quantity of data read.
```
FrankReh, line 75: No need to change this now, but someday, I would want the order in the comment and the others like it to reflect the order of the values returned. I see the old comment it replaces didn't do that either.

src/lib.rs
```diff
@@ -176,30 +177,30 @@ pub fn builder() -> Builder {
...
-    /// The caller can specify even a larger cq entries count by using the uring_builder
-    /// as shown in the start example below.
-    pub fn entries(&mut self, e: u32) -> &mut Self {
-        self.entries = e;
+    /// The default value is 256.
+    /// The kernel requires the number of submission queue entries to be a power of two,
+    /// and that it be less than the number of completion queue entries.
+    /// This function will adjust the `cq_entries` value to be at least 2 times `sq_entries`
+    pub fn entries(&mut self, sqe: u32) -> &mut Self {
+        self.entries = sqe;
```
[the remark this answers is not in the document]
rrichardson: Changed.

rrichardson:
On a side note:  I am really hoping pushing a new version of the docs will fix the existing ones,  because they're broke AF

src/runtime/mod.rs
```diff
@@ -13,7 +13,16 @@ thread_local! {
...
 
-/// The Runtime executor
+/// The Runtime Executor
+///
+/// This is the Runtime for `tokio-uring`.
+/// It wraps the default [`Runtime`] using the platform-specific Driver.
+///
+/// # Multithreading
+///
+/// This executes futures and tasks within the current-thread only.
```
FrankReh, line 23: So, I don't see a need to mirror what is in the tokio doc. Why not remove the `# Multithreading` and just make sure somewhere in this file's doc is says the current state is `current-thread` tokio scheduling only. Of course if you want to get this in quickly now, you may be better off just nixing this change to the comments for now.

src/runtime/mod.rs
FrankReh: Again, if we don't hear from @Noah-Kennedy, you can just remove this change for now because while I like comments, I really care about technical comments that help us understand the tricky stuff when wanting to modify the design. The over all stuff about how to use it and how it relates to tokio I would prefer to leave to Noah over time.

src/runtime/mod.rs
rrichardson: K.

src/net/tcp/stream.rs
rrichardson: Are you referring to the contents of the `BufResult` tuple?   That'd probably be a good idea.

FrankReh (approved):
Thanks. Will wait 'till the end of the day in case @Noah-Kennedy wants to comment further, otherwise I'll merge what's there.

FrankReh:
@rrichardson Would it be easy for you to merge from mainline to pick up the change to udp.rs from earlier today and push again?

Noah-Kennedy (approved):

Noah-Kennedy:
@rrichardson I will merge this as soon as you update this branch from master and resolve conflicts.

rrichardson:
> @rrichardson I will merge this as soon as you update this branch from master and resolve conflicts.

Done.

FrankReh:
Unfortunately the latest comment addition doesn't get past a CI documentation check.

FrankReh:
Thanks!

PR #206, opened 2023-01-02
chore: fix criterion benchmarks

FrankReh:
Follow up to #151.

### DIFF
--- a/benches/criterion/no_op.rs
+++ b/benches/criterion/no_op.rs
@@ -1,14 +1,38 @@
-use criterion::{
-    criterion_group, criterion_main, BenchmarkId, Criterion, SamplingMode, Throughput,
-};
+/// Benchmark the overhead of the uring no_op operation.
+///
+/// It should be understood that a single call to even the no_op operation is relatively expensive
+/// even though the kernel is not being asked to do anything except copy the user_data from the
+/// submission queue entry to a completion queue entry.
+///
+/// The cost comes from the application creating writing to the SQE, then just before going idle,
+/// making the system call to notify the kernel that the submission queue tail has been changed by
+/// the application, then being awoken at the mio level by the kernel that something about the
+/// uring's fd has become ready, then running through the tokio_uring's completion ring handler and
+/// finding the CQE that trigger's a lookup into the slab that gets a future's awake function
+/// called, that makes a task runnable, that then allows the Tokio runtime to schedule the task
+/// that finally gets to see the no_op operation completed.
+///
+/// When run in isolation, that is to say when then concurrency parameter is 1, this round trip
+/// takes between 3.2 and 1.5 *micro* seconds on a 64 bit Linux machine - depending on the CPU
+/// speed and the cache sizes.
+///
+/// When run with enough concurrency, say 1000 tasks each doing the same thing, it can be seen the
+/// actual overhead for the tokio_uring no_op call and wait comes down to between 310 and 220ns,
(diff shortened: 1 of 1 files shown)


FrankReh:
@Noah-Kennedy This is the kind of output I'm getting from the criterion benchmark with these changes. These times jive with what I was seeing with a different form of no_op benchmark I had experimented with before: that when there is little concurrency, the time per no_op is high, say 3.2 microseconds, but when there are lots of concurrent uring accesses, the amortized time per no_op is very low, about 332 nanoseconds on my slow Linux machine.

```sh
     Running benches/criterion/no_op.rs (target/release/deps/criterion_no_op-22257b96c5b4fa14)
no_op/concurrent tasks/1
                        time:   [3.2306 µs 3.2484 µs 3.2698 µs]
no_op/concurrent tasks/2
                        time:   [1.7880 µs 1.7974 µs 1.8086 µs]
no_op/concurrent tasks/5
                        time:   [877.98 ns 882.60 ns 888.23 ns]
no_op/concurrent tasks/10
                        time:   [574.71 ns 577.88 ns 581.69 ns]
no_op/concurrent tasks/100
                        time:   [324.33 ns 326.25 ns 328.63 ns]
no_op/concurrent tasks/1000
                        time:   [309.72 ns 311.48 ns 313.65 ns]
no_op/concurrent tasks/10000
                        time:   [465.57 ns 469.27 ns 473.69 ns]
```

benches/criterion/no_op.rs
```diff
@@ -28,44 +59,52 @@ fn run_no_ops(opts: &Options, count: u64) -> Duration {
...
     let mut group = c.benchmark_group("no_op");
     let mut opts = Options::default();
-    for concurrency in [1, 32, 64, 256].iter() {
+    for concurrency in [1, 2, 5, 10, 100, 1000, 10_000].iter() {
         opts.concurrency = *concurrency;
 
         // We perform long running benchmarks: this is the best mode
         group.sampling_mode(SamplingMode::Flat);
 
-        group.throughput(Throughput::Elements(opts.iterations as u64));
```
Noah-Kennedy, line 66: Don't remove this, set it to be the concurrency value

benches/criterion/no_op.rs
```diff
@@ -28,44 +59,52 @@ fn run_no_ops(opts: &Options, count: u64) -> Duration {
...
+            // benchmark count (accounting for the number of tasks that are being spawned).
 
-                for _ in 0..opts.iterations {
-                    js.spawn_local(tokio_uring::no_op());
-                }
+            for _ in 0..opts.concurrency {
+                let opts = opts.clone();
+                js.spawn_local(async move {
+                    // Run the required number of iterations (per the concurrency amount)
+                    for _ in 0..(count / opts.concurrency) {
```
Noah-Kennedy, line 79: I'm not sure about this approach. If we just use `Throughput`, criterion will do normalization and adjustment for us.

benches/criterion/no_op.rs
```diff
@@ -28,44 +59,52 @@ fn run_no_ops(opts: &Options, count: u64) -> Duration {
...
+                        let _ = black_box(tokio_uring::no_op().await);
```
Noah-Kennedy, line 80: Good idea here

Noah-Kennedy (changes requested):
I like the refactoring and added docs here, but #205 is logically correct and this isn't. I'd recommend pulling some of the changes from that PR into here and then we can just use this PR.

PR #210, opened 2023-01-12
Update README.md to improve wording

yerke:


FrankReh:
@Noah-Kennedy I think we know by now that if the README is going to recommend a kernel, it should either say uring kernel feature development and bug fixing has gone well into the 6.1 kernel or it should say the user of this crate should check with the uring status of the kernel they are trying to target. Calling out any old kernel version explicitly sounds like the crate author is making a recommendation on what minimal kernel will work.

Noah-Kennedy:
@FrankReh I'm in complete agreement here

yerke:
Do you want me to update the wording in this PR somehow? Or should I just close it and let you handle it separately? Thanks.

FrankReh:
I suggest you leave this open and we can treat it as an open Issue. I think Noah will change the README at some point and this can then be closed. Thanks for wanting to help. At this point, it's better to leave an old comment as it should be pretty evident it is an old/obsolete recommendation.

PR #212, opened 2023-01-17
Add support for zero copy sendmsg.

MA-ETL:
This PR adds support for the IORING_OP_SENDMSG_ZC opcode added in Linux 6.1 with a new sendmsg_zc function for UDP sockets.


src/io/sendmsg_zc.rs
```diff
@@ -0,0 +1,58 @@
...
+    fd: SharedFd,
+
+    pub(crate) msghdr: libc::msghdr,
+
+    /// Hold the number of transmitted bytes
+    bytes: usize,
+}
+
+impl Op<SendMsgZc, MultiCQEFuture> {
+    pub(crate) fn sendmsg_zc(fd: &SharedFd, msghdr: &libc::msghdr) -> io::Result<Self> {
```
FrankReh, line 17: I'm not sure about parameters that don't pass ownership of the underlying buffers to the operation. Could you pass a Vec\<IoSlice>?

src/net/udp.rs
```diff
@@ -220,6 +220,11 @@ impl UdpSocket {
         self.inner.send_zc(buf).await
     }
 
+    /// Sends a message on the socket using a msghdr.
+    pub async fn sendmsg_zc(&self, msghdr: &libc::msghdr) -> (libc::msghdr, io::Result<usize>) {
```
FrankReh, line 224: I believe the convention is to pass the size result first and the buffer ownership back second.

FrankReh (changes requested):
Nice. Biggest question is why ownership of the list of buffers isn't passed in. Seems to break the convention established for all other operations.

FrankReh (changes requested):
Looking good.

Would you mind also pasting some example code showing this works into a comment in this PR so if we get around to testing on a 6.1 release, we could more easily throw a test together?

src/io/sendmsg_zc.rs
```diff
@@ -0,0 +1,107 @@
...
+    pub(crate) msghdr: libc::msghdr,
+
+    /// Hold the number of transmitted bytes
+    bytes: usize,
+}
+
+impl Op<SendMsgZc, MultiCQEFuture> {
+    pub(crate) fn sendmsg_zc(
+        fd: &SharedFd,
+        io_slices: Vec<IoSlice<'static>>,
```
ollie-etl, line 26: I'm still not sure this is right @FrankReh, @MA-ETL . Would a `Vec<T: IoBuf>` be a better choice here?

Would that not allow more flexibility in constructing a send from non-contiguous buffers? I would take ownership of the buffers, as the current ops do, and construct a `msghdr` in place, like now. The user is then freed from buffer handling, and the dreaded `static` constraint is banished

It may also open the door to compatibility with `FixedBuffers`, although I'm unsure if that implemented in io_uring yet.

FrankReh: That seems like a good point. Sorry to @MA-ETL that I didn't think of that in my first replay. Thank you @ollie-etl .

FrankReh (changes requested):
Looking really good. With no tests, I'm assuming you are seeing this work in your own tree?

src/io/sendmsg_zc.rs
```diff
@@ -0,0 +1,105 @@
...
+        })
+    }
+}
+
+impl<T, U> Completable for SendMsgZc<T, U> {
+    type Output = (io::Result<usize>, Vec<T>, Option<U>);
+
+    fn complete(self, cqe: CqeResult) -> (io::Result<usize>, Vec<T>, Option<U>) {
+        // Convert the operation result to `usize`
+        let res = cqe.result.map(|v| v as usize);
```
FrankReh, line 88: Still have the question, why self.bytes is maintained but not added to res when this is complete and successful? Sorry if I'm missing something obvious.

FrankReh (changes requested):
Nice. Probably two outstanding questions on my part. Thanks for this. I think it's time we asked @Noah-Kennedy for a review too. (I haven't run this through CI because I'm assuming it compiles and there are no new tests.)

src/io/sendmsg_zc.rs
```diff
@@ -0,0 +1,110 @@
...
+    type Output = (io::Result<usize>, Vec<T>, Option<U>);
+
+    fn complete(self, cqe: CqeResult) -> (io::Result<usize>, Vec<T>, Option<U>) {
+        // Convert the operation result to `usize`
+        let mut res = cqe.result.map(|v| v as usize);
+
+        // Add the number of bytes to res.
+        if let Ok(_res) = res {
+            res = Ok(_res + self.bytes)
+        }
```
ollie-etl, line 93:
```suggestion
        // Convert the operation result to `usize`, and add previous byte count
        let mut res = cqe.result.map(|v| self.bytes + v as usize);
```

ollie-etl:
@FrankReh Are all of your concerns addressed?

src/io/sendmsg_zc.rs
```diff
@@ -84,13 +84,8 @@ impl<T, U> Completable for SendMsgZc<T, U> {
...
     fn complete(self, cqe: CqeResult) -> (io::Result<usize>, Vec<T>, Option<U>) {
-        // Convert the operation result to `usize`
-        let mut res = cqe.result.map(|v| v as usize);
-
-        // Add the number of bytes to res.
-        if let Ok(_res) = res {
-            res = Ok(_res + self.bytes)
-        }
+        // Convert the operation result to `usize`, and add previous byte count
+        let res = cqe.result.map(|v| self.bytes + v as usize);
```
FrankReh, line 88: Yes. That looks good.

FrankReh (approved):
Thanks. Looks very good. I wish we have a test case or some sample of a test case attached to this PR so when we were ready to test with a 6.1 kernel, it would be easy for somebody new to know what to add. But it is ready to merge anyway. Just going to give @Noah-Kennedy a chance at feedback before I merge in a few hours. I realize it doesn't break or change any existing functionality anyway.

PR #213, opened 2023-01-19
Generialise FixedBufRegistry and FixedBufPool

ollie-etl:
Extends both registries to accept any `Iter` of `IoBufMut`, instead of Vec<u8>.

This is motivated by my usecase, which cannot safely provide its memory as a `Vec`. Whilst I can _unsafely_ create one,
it would be invalid to call drop on the created `Vec`

This implementation, which explicitly keeps the backing buffers present, rather then `mem::forget` and recreate means the `Drop` handling is preserved.

As a trivial example, this would now admit `Box<[u8]>` as a buffer.

I realize I could provide my own implementation of the pools, but it seem unnecessary - the generalized version serves both purposes. The overhead is an additional allocation on pool creation, although this will be optimized away by [in place specialisation](https://stdrs.dev/nightly/x86_64-unknown-linux-gnu/src/alloc/vec/in_place_collect.rs.html#153-213), for the common use case.

### DIFF
--- a/examples/tcp_listener_fixed_buffers.rs
+++ b/examples/tcp_listener_fixed_buffers.rs
@@ -4,7 +4,7 @@
 use std::{env, iter, net::SocketAddr};
 
 use tokio_uring::{
-    buf::{fixed::FixedBufRegistry, BoundedBuf},
+    buf::{fixed::FixedBufRegistry, BoundedBuf, IoBufMut},
     net::{TcpListener, TcpStream},
 }; // BoundedBuf for slice method
 
@@ -50,7 +50,11 @@ async fn accept_loop(listen_addr: SocketAddr) {
 // A loop that echoes input to output. Use one fixed buffer for receiving and sending the response
 // back. Once the connection is closed, the function returns and the fixed buffer is dropped,
 // getting the fixed buffer index returned to the available pool kept by the registry.
-async fn echo_handler(stream: TcpStream, peer: SocketAddr, registry: FixedBufRegistry) {
+async fn echo_handler<T: IoBufMut>(
+    stream: TcpStream,
+    peer: SocketAddr,
+    registry: FixedBufRegistry<T>,
+) {
     println!("peer {} connected", peer);
 
     // Get one of the two fixed buffers.
--- a/src/buf/fixed/handle.rs
+++ b/src/buf/fixed/handle.rs
@@ -85,7 +85,7 @@ unsafe impl IoBufMut for FixedBuf {
     }
 
     unsafe fn set_init(&mut self, pos: usize) {
-        if self.buf.iovec.iov_len < pos {
+        if self.buf.iovec.iov_len >= pos {
             self.buf.init_len = pos
         }
     }
--- a/src/buf/fixed/handle.rs
+++ b/src/buf/fixed/handle.rs
@@ -4,7 +4,6 @@ use crate::buf::{IoBuf, IoBufMut};
 use libc::iovec;
 use std::cell::RefCell;
 use std::fmt::{self, Debug};
-use std::mem::ManuallyDrop;
 use std::ops::{Deref, DerefMut};
 use std::rc::Rc;
 
@@ -33,8 +32,7 @@ pub(super) struct CheckedOutBuf {
 ///
 pub struct FixedBuf {
     registry: Rc<RefCell<dyn FixedBuffers>>,
-    buf: ManuallyDrop<Vec<u8>>,
-    index: u16,
+    buf: CheckedOutBuf,
 }
 
 impl Drop for FixedBuf {
@@ -44,59 +42,51 @@ impl Drop for FixedBuf {
         // maintained accordingly to the safety contracts on
         // Self::new and IoBufMut.
         unsafe {
-            registry.check_in(self.index, self.buf.len());
+            registry.check_in(self.buf.index, self.buf.init_len);
         }
     }
 }
 
 impl FixedBuf {
     // Safety: Validity constraints must apply to CheckedOutBuf members:
-    // - iovec must refer to an array allocated by Vec<u8>;
     // - the array will not be deallocated until the buffer is checked in;
     // - the data in the array must be initialized up to the number of bytes
     //   given in init_len.
     pub(super) unsafe fn new(registry: Rc<RefCell<dyn FixedBuffers>>, data: CheckedOutBuf) -> Self {
-        let CheckedOutBuf {
-            iovec,
-            init_len,
-            index,
-        } = data;
-        let buf = Vec::from_raw_parts(iovec.iov_base as _, init_len, iovec.iov_len);
         FixedBuf {
             registry,
-            buf: ManuallyDrop::new(buf),
-            index,
+            buf: data,
         }
     }
 
     /// Index of the underlying registry buffer
     pub fn buf_index(&self) -> u16 {
-        self.index
+        self.buf.index
     }
 }
 
 unsafe impl IoBuf for FixedBuf {
     fn stable_ptr(&self) -> *const u8 {
-        self.buf.as_ptr()
+        self.buf.iovec.iov_base as _
     }
 
     fn bytes_init(&self) -> usize {
-        self.buf.len()
+        self.buf.init_len
     }
 
     fn bytes_total(&self) -> usize {
-        self.buf.capacity()
+        self.buf.iovec.iov_len
     }
 }
 
 unsafe impl IoBufMut for FixedBuf {
     fn stable_mut_ptr(&mut self) -> *mut u8 {
-        self.buf.as_mut_ptr()
+        self.buf.iovec.iov_base as _
     }
 
     unsafe fn set_init(&mut self, pos: usize) {
-        if self.buf.len() < pos {
-            self.buf.set_len(pos)
+        if self.buf.iovec.iov_len < pos {
+            self.buf.init_len = pos
         }
     }
 }
@@ -105,21 +95,22 @@ impl Deref for FixedBuf {
     type Target = [u8];
 
     fn deref(&self) -> &[u8] {
-        &self.buf
+        unsafe { std::slice::from_raw_parts(self.buf.iovec.iov_base as _, self.buf.init_len) }
     }
 }
 
 impl DerefMut for FixedBuf {
     fn deref_mut(&mut self) -> &mut [u8] {
-        &mut self.buf
+        unsafe { std::slice::from_raw_parts_mut(self.buf.iovec.iov_base as _, self.buf.init_len) }
     }
 }
 
 impl Debug for FixedBuf {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        let buf: &[u8] = self;
         f.debug_struct("FixedBuf")
-            .field("buf", &*self.buf) // deref ManuallyDrop
-            .field("index", &self.index)
+            .field("buf", &buf) // as slice
+            .field("index", &self.buf.index)
             .finish_non_exhaustive()
     }
 }
--- a/src/buf/fixed/handle.rs
+++ b/src/buf/fixed/handle.rs
@@ -52,10 +52,10 @@ impl FixedBuf {
     // - the array will not be deallocated until the buffer is checked in;
     // - the data in the array must be initialized up to the number of bytes
     //   given in init_len.
-    pub(super) unsafe fn new(registry: Rc<RefCell<dyn FixedBuffers>>, data: CheckedOutBuf) -> Self {
+    pub(super) unsafe fn new(registry: Rc<RefCell<dyn FixedBuffers>>, buf: CheckedOutBuf) -> Self {
         FixedBuf {
             registry,
-            buf: data,
+            buf,
         }
     }
 
--- a/src/buf/fixed/handle.rs
+++ b/src/buf/fixed/handle.rs
@@ -82,7 +82,7 @@ unsafe impl IoBufMut for FixedBuf {
     }
 
     unsafe fn set_init(&mut self, pos: usize) {
-        if self.buf.iovec.iov_len >= pos {
+        if self.buf.init_len < pos {
             self.buf.init_len = pos
         }
     }
--- a/src/buf/fixed/handle.rs
+++ b/src/buf/fixed/handle.rs
@@ -53,10 +53,7 @@ impl FixedBuf {
     // - the data in the array must be initialized up to the number of bytes
     //   given in init_len.
     pub(super) unsafe fn new(registry: Rc<RefCell<dyn FixedBuffers>>, buf: CheckedOutBuf) -> Self {
-        FixedBuf {
-            registry,
-            buf,
-        }
+        FixedBuf { registry, buf }
     }
 
     /// Index of the underlying registry buffer
@@ -95,12 +92,14 @@ impl Deref for FixedBuf {
     type Target = [u8];
 
     fn deref(&self) -> &[u8] {
+        // Safety: The iovec points to a slice held  in self.buffers, to which no mutable reference exists.
         unsafe { std::slice::from_raw_parts(self.buf.iovec.iov_base as _, self.buf.init_len) }
     }
 }
 
 impl DerefMut for FixedBuf {
     fn deref_mut(&mut self) -> &mut [u8] {
+        // Safety: The iovec points to a slice held in self.buffers, to which no other reference exists.
         unsafe { std::slice::from_raw_parts_mut(self.buf.iovec.iov_base as _, self.buf.init_len) }
     }
 }
--- a/src/buf/fixed/pool.rs
+++ b/src/buf/fixed/pool.rs
@@ -378,14 +378,15 @@ impl<T: IoBufMut> FixedBuffers for Inner<T> {
 
 impl<T> Drop for Inner<T> {
     fn drop(&mut self) {
-        let _iovecs = unsafe {
-            Vec::from_raw_parts(self.raw_bufs.as_ptr(), self.states.len(), self.orig_cap)
-        };
         // Assert all buffers are checked in
         for state in self.states.iter() {
             if let BufState::CheckedOut = state {
                 unreachable!("all buffers must be checked in");
             }
         }
+        // Rebuild Vec<iovec>, so it's dropped
+        let _ = unsafe {
+            Vec::from_raw_parts(self.raw_bufs.as_ptr(), self.states.len(), self.orig_cap)
+        };
     }
 }
--- a/src/buf/fixed/pool.rs
+++ b/src/buf/fixed/pool.rs
@@ -266,7 +266,7 @@ enum BufState {
 impl<T: IoBufMut> Inner<T> {
     fn new(bufs: impl Iterator<Item = T>) -> Self {
         let bufs = bufs.take(cmp::min(UIO_MAXIOV as usize, u16::MAX as usize));
-        let mut buffers  = bufs.collect::<Vec<T>>();
+        let mut buffers = bufs.collect::<Vec<T>>();
         let mut iovecs = Vec::with_capacity(buffers.len());
         let mut states = Vec::with_capacity(buffers.len());
         let mut free_buf_head_by_cap = HashMap::new();
--- a/src/buf/fixed/pool.rs
+++ b/src/buf/fixed/pool.rs
@@ -265,7 +265,12 @@ enum BufState {
 
 impl<T: IoBufMut> Inner<T> {
     fn new(bufs: impl Iterator<Item = T>) -> Self {
+        // Limit the number of buffers to the maximum allowable number.
         let bufs = bufs.take(cmp::min(UIO_MAXIOV as usize, u16::MAX as usize));
+        // Collect into `buffers`, which holds the backing buffers for
+        // the lifetime of the pool. Using collect may allow
+        // the compiler to apply collect in place specialization,
+        // to avoid an allocation.
         let mut buffers = bufs.collect::<Vec<T>>();
         let mut iovecs = Vec::with_capacity(buffers.len());
         let mut states = Vec::with_capacity(buffers.len());
--- a/src/buf/fixed/pool.rs
+++ b/src/buf/fixed/pool.rs
@@ -1,6 +1,7 @@
 use super::handle::CheckedOutBuf;
 use super::{FixedBuf, FixedBuffers};
 
+use crate::buf::IoBufMut;
 use crate::runtime::CONTEXT;
 use libc::{iovec, UIO_MAXIOV};
 use std::cell::RefCell;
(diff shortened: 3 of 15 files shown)


FrankReh:
This looks good to me. The way drops are handled now, would it let me carve out the buffers from a single mmap'ed block if I wrote the iterator to unsafely step through the allocated memory? That's been my last sticking point as I think some kernel interfaces are more efficient with page aligned buffers or in some cases, even require page aligned buffers.

mzabaluev:
> The way drops are handled now, would it let me carve out the buffers from a single mmap'ed block

I think that's what the system allocator does in some cases, or maybe that's only if the allocation sizes are sufficiently big.

> some kernel interfaces are more efficient with page aligned buffers or in some cases, even require page aligned buffers.

I believe this does not apply to io-uring: I/O operations and `IORING_REGISTER_BUFFERS` do not require page alignment. The latter, in fact, locates and maps the pages that contain the buffers, precisely to make operations with fixed buffers maximally efficient.

src/buf/fixed/pool.rs
```diff
@@ -230,16 +231,17 @@ impl FixedBufPool {
...
     // same as the length of the states array.
     raw_bufs: ptr::NonNull<iovec>,
     // State information on the buffers. Indices in this array correspond to
     // the indices in the array at raw_bufs.
     states: Vec<BufState>,
-    // Original capacity of raw_bufs as a Vec.
-    orig_cap: usize,
+    // Original buffers, kept until drop
+    #[allow(dead_code)]
+    buffers: Vec<T>,
```
mzabaluev, line 244: When `T = Vec<u8>`, this adds an allocation that is completely unnecessary, since the full data of each `Vec` can be recovered from its `iovec` and `BufState`.
I'd rather not pessimize the implementations for the most common (or at least, the most convenient) parameter case.

mzabaluev (changes requested):

FrankReh:
> I believe this does not apply to io-uring: I/O operations and `IORING_REGISTER_BUFFERS` do not require page alignment

I beg to differ. (But I'm often wrong.) The tests that come with the liburing C library itself use O_DIRECT in many places, as well as mmap for that very reason.

When I have an app that will use the same buffers countless times for I/O, I will certainly want them page aligned. For small use cases, I won't even go to the trouble of registering the buffers first.

src/buf/fixed/pool.rs
ollie-etl: @mzabaluev I'm not sure about most convenient, that depends on how you allocate your memory. I've stated that for me, and probably for @FrankReh if he's using MMap's, Vec is a non starter.

However, I have realised I can address rour concern here. If I was to use `collect` rather than push, like currently, it seems collect ends up calling this specialization for vec: https://stdrs.dev/nightly/x86_64-unknown-linux-gnu/src/alloc/vec/in_place_collect.rs.html#153-213, which would avoid allocations.

ollie-etl:
> This looks good to me. The way drops are handled now, would it let me carve out the buffers from a single mmap'ed block if I wrote the iterator to unsafely step through the allocated memory? That's been my last sticking point as I think some kernel interfaces are more efficient with page aligned buffers or in some cases, even require page aligned buffers.

@FrankReh Yes: thats what I do with it.

src/buf/fixed/registry.rs
```diff
@@ -281,20 +284,12 @@ impl FixedBuffers for Inner {
     }
 }
 
-impl Drop for Inner {
+impl<T> Drop for Inner<T> {
     fn drop(&mut self) {
-        let iovecs = unsafe {
-            Vec::from_raw_parts(self.raw_bufs.as_ptr(), self.states.len(), self.orig_cap)
-        };
```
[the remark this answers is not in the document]
ollie-etl: I'll address this.

src/buf/fixed/pool.rs
ollie-etl: @mzabaluev I think the most recent commits address your concern for the common case

src/buf/fixed/registry.rs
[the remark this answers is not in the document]
ollie-etl: fixed

src/buf/fixed/pool.rs
mzabaluev: My concern is having the extra `Vec` for keeping track of the buffers, that is redundant in the special, but important, case of `Vec` buffers. How it's populated is less important.

For carving up an mmap, I would suggest a separate collection implementing `FixedBuffers` on its own. I think `IoBufMut` is an imperfect abstraction for managing segments in a larger mapped area, since you need to keep track of the overall mmap, and adding a refcount just for that looks like impacting performance without a real need for that.

With all the concern about managing mmaps explicitly, perhaps it would be a good place to explain what are the benefits of this when applied to fixed buffers, compared to just getting buffers from the default allocator? Otherwise, this looks like a solution without a clearly stated problem.

ollie-etl: Ok: I specifically work pages allocated from a device driver. The _only_ way I have to access them is through a `mmap`.

ollie-etl: But i think you missed my point above. The extra Vec is not allocated in the special but important case of passing in a vec. It is stored, sure, but i think a point and length is not a significant overhead. In that case _no allocation for the backing memory of that Vec is performed_ the compiler should just take ownership of the impl Iterator passed in, though specialisation

FrankReh (changes requested):
I like it.

This and the related PR about supporting a `next` Future are making this feel like a very tight aspect of the crate.

@mzabaluev Would you be willing to give this a thumbs up?

src/buf/fixed/registry.rs
```diff
@@ -202,23 +206,23 @@ enum BufState {
...
-impl Inner {
-    fn new(bufs: impl Iterator<Item = Vec<u8>>) -> Self {
+impl<T: IoBufMut> Inner<T> {
+    fn new(bufs: impl Iterator<Item = T>) -> Self {
         let bufs = bufs.take(cmp::min(UIO_MAXIOV as usize, u16::MAX as usize));
-        let (size_hint, _) = bufs.size_hint();
-        let mut iovecs = Vec::with_capacity(size_hint);
-        let mut states = Vec::with_capacity(size_hint);
-        for mut buf in bufs {
+        let mut buffers = bufs.collect::<Vec<T>>();
```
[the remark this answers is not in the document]
ollie-etl: @FrankReh You're not alone in getting long in the tooth,, and losing the ability to compute 2 things at once. I think it makes you code better! I'll add the comment

src/buf/fixed/pool.rs
```diff
@@ -369,6 +373,9 @@ impl<T: IoBufMut> FixedBuffers for Inner<T> {
 
 impl<T> Drop for Inner<T> {
     fn drop(&mut self) {
+        let _iovecs = unsafe {
```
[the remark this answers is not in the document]
ollie-etl: done

src/buf/fixed/registry.rs
```diff
@@ -284,6 +288,9 @@ impl<T: IoBufMut> FixedBuffers for Inner<T> {
 
 impl<T> Drop for Inner<T> {
     fn drop(&mut self) {
+        let _iovecs = unsafe {
```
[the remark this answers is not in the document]
ollie-etl: done

ollie-etl:
@FrankReh are all of your concerns addressed?

ollie-etl:
@mzabaluev, is there anything you would like addressing?

ollie-etl:
@FrankReh  I wanted to reply to this.

> This looks good to me. The way drops are handled now, would it let me carve out the buffers from a single mmap'ed block if I wrote the iterator to unsafely step through the allocated memory? That's been my last sticking point as I think some kernel interfaces are more efficient with page aligned buffers or in some cases, even require page aligned buffers.

You can do this in a unsafe way if that is your use case. Your iterator will produce into something like `Segment`, and @mzabaluev is correct that normally you'd pay the (small) Rc penalty. They'd be something like

```rust
struct Segment {
   backing_buffer: Rc<MmapRaw>,
   ptr: *mut u8
   init: usize,
   len: usize
}
```

However, because you know they that if the pool is being dropped, none are in use, and that all remain alive under the pool is dropped, you can replace backing_buffer with `Option<MmapRaw`, with all but a single Segment being `None`, removing the Rc overhead. Yuk, I know.

FrankReh:
> You can do this in a unsafe way if that is your use case. Your iterator will produce into something like `Segment`, and @mzabaluev is correct that normally you'd pay the (small) Rc penalty.

Thanks. I don't mind tapping into unsafe. I had hoped the iterator could be implemented programmatically, but really just to keep from polluting the heap with unnecessary stuff - not to safe space or startup/tear-down time really. For me, setting up a pool would be once at app, or thread start up, and then the buffers get reused for hours or days before a reboot anyway.

Also for pools that I want to use, I don't see that tracking init per buf is important. If the compiler lets me, I would avoid tracking init and simply zero all the memory ahead of time.

FrankReh (approved):
Looks good to me. I'll plan on merging tomorrow to give others time to still comment. Thanks!

src/buf/fixed/pool.rs
mzabaluev: Sure, I didn't mean the buffers themselves are duplicated. But the bookkeeping in the `buffers` array is redundant in case of `Vec<u8>`, and it's still not correct: in `check_in_internal`, the buffer's initialized data length must be updated, but the current code does not do it. This "does not matter" with `Vec<u8>` buffers, but with generic `IoBufMut` I think the implementation is obliged to update it.

ollie-etl: I'd certianly overlooked the vec internall to FixedBuf: I think that does need changing in this PR.

I don't follow with the dual accounting though. The only location where I think the Pool or registry might need to call set_init on the IoBufMut is on the Drop of the pool or registry. My reasoning being that the buffer is wholly owned by the registry or pool whilst they are in existence, and it knows the capacity and init_len setting. If you view the Ownership -> Drop of the pool as a single operation, then its only obliged to update it before it drops, which might return the buffer elsewhere, if its a Rc backed buffer. Whilst it in the pool, nothing else has visibility of those values.

mzabaluev:
I still think this is a wrong place to extend, because it's an imperfect fit for either of the discussed use cases:

1. It introduces redundant bookkeeping in the `Vec<u8>` case.
2. It forces "segments out of a large mmap" case into extra refcounting in the buffer handles.

I like generics, but not when they de-optimize their most used instantiations. And we already have an extension point in the `FixedBuffers` vtable, which has a runtime cost. So we might as well amortize the flexibility into that.

> Also for pools that I want to use, I don't see that tracking init per buf is important. If the compiler lets me, I would avoid tracking init and simply zero all the memory ahead of time.

While this is a special case where it may be not important (I think zero-initializing is better than admitting UB into your program, though), the proposed change introduces a generic implementation that should work correctly for all buffer types.
Hence why I'm suggesting to address specialized cases at the `FixedBuffers` level which the current design is already allows (pending making the trait public).

ollie-etl:
> 1. It introduces redundant bookkeeping in the Vec<u8> case.

I really don't follow this logic, sorry. I see no extra overhead

mzabaluev:
> When I have an app that will use the same buffers countless times for I/O, I will certainly want them page aligned.

I believe you get fairly aligned buffers with something like `iter::repeat_with(|| Vec::with_capacity(16_384)).take(N)`.
Also, I've learned that any speculative talks about performance are empty until demonstrated with an actual benchmark.

mzabaluev:
> > 1. It introduces redundant bookkeeping in the Vec case.
> 
> I really don't follow this logic, sorry. I see no extra overhead.

A member of the `buffers` array has exactly the same information as the corresponding `iovec` and the `init_len` member of the `BufState::Free` struct. Furthermore, it's currently allowed to go out of sync with the `init_len` value that is updated on buffer check-ins, which technically results in UB when the buffers' `Vec<u8>` instances are dropped.

src/buf/fixed/pool.rs
ollie-etl: @mzabaluev I believe I have addressed this. set_init is called before ownership of the buffers passes out of the poll or registry

ollie-etl:
> A member of the buffers array has exactly the same information as the corresponding iovec and the init_len member of the BufState::Free struct.  

Yes. But:
1. its no more memory allocated, the buffers are present but "forgotten" in the current implementation.
2. No extra account is done on those buffers (until registry drop)
 
> Furthermore, it's currently allowed to go out of sync with the init_len value that is updated on buffer check-ins, which technically results in UB when the buffers' Vec<u8> instances are dropped is probably OK for this specific case, but not generically.

Its not UB. However, i did update the drop implementation to update the buffers on drop, should they be some Rc backed implementation which takes them elsewhere, and you want to know how many bytes currently rest in them. I'd be very upset if the compiler didn't strip that completely in the case you do just drop, so don't believe this is an overhead.

ollie-etl:
Hmm, that last change does appear to have broken test - will check

src/buf/fixed/handle.rs
```diff
@@ -44,59 +42,51 @@ impl Drop for FixedBuf {
...
 unsafe impl IoBufMut for FixedBuf {
     fn stable_mut_ptr(&mut self) -> *mut u8 {
-        self.buf.as_mut_ptr()
+        self.buf.iovec.iov_base as _
     }
 
     unsafe fn set_init(&mut self, pos: usize) {
-        if self.buf.len() < pos {
-            self.buf.set_len(pos)
+        if self.buf.iovec.iov_len >= pos {
```
ollie-etl, line 88: I found this interesting @mzabaluev . Is the original in error? You can't ever decrease the number of initialized bytes?
Changed to what I believe to be the intention, which is you cannot have more initialized bytes than the capacity.

ollie-etl:
@FrankReh You'll want to review the changes here also. I don't think they change the Pr much

src/buf/fixed/handle.rs
```diff
@@ -33,8 +32,7 @@ pub(super) struct CheckedOutBuf {
 ///
 pub struct FixedBuf {
     registry: Rc<RefCell<dyn FixedBuffers>>,
-    buf: ManuallyDrop<Vec<u8>>,
-    index: u16,
+    buf: CheckedOutBuf,
```
ollie-etl, line 35: There didn't seem any point in duplicating the fields

FrankReh:
I have to wait until much later today to look. Family dentistry emergency.

I like the discussion, when it's not personal. It seems the code is getting better as the two main protagonists here are talking.

I'll remind everyone the whole idea behind supporting uring is to get at performance gains between the userland and the kernel/hardware and we generally do it in good faith, without benchmarks to prove that we are on the right track. And my opinion is we try to present the most generic approach we can that fits into the boundaries of the tokio ecosystem. If things work out of the box, that makes for a nice user experience. 

(I hope my opinions aren't used against me.)

Thanks. (And a reminder for everyone to take care of their teeth while they can.)

src/buf/fixed/handle.rs
mzabaluev: This is OK with the proposed generalization: there is no longer necessarily a `Vec` managing the buffer's memory, so it's better to just store the raw fields. The manually dropped `Vec` was used for convenient access to the slice.

src/buf/fixed/handle.rs
mzabaluev: No, `init_len` is for tracking the length of the initialized bytes, and once they are known to be initialized, there is no reason to declare them uninitialized. The original code is correct (for `Vec`).

ollie-etl: Whilst I guess thats true, it goes against what I would expect

```rust
file1.read_at(buf,0);
file2.read_at(buf,0);
```
I guess i'd expect `bytes_init` to represent the number of bytes from file 2 at this point.

src/buf/fixed/handle.rs
ollie-etl: That was all i meant - after the changes in FixedBuf, you'd want all the fields from checkedbuf, so reused them.

ollie-etl:
> I like the discussion, when it's not personal. It seems the code is getting better as the two main protagonists here are talking.

Indeed. Whilst its true i can find getting things review in this particular repository frustrating, I'll have to admit the review always resulted in better code. As for contributers, I've nothing but thanks for people who give theirs (or their companies time) for improving OS code. I'll note that a lot of time seems to occur here out of hours.

examples/tcp_listener_fixed_buffers.rs
```diff
@@ -50,7 +50,11 @@ async fn accept_loop(listen_addr: SocketAddr) {
 // A loop that echoes input to output. Use one fixed buffer for receiving and sending the response
 // back. Once the connection is closed, the function returns and the fixed buffer is dropped,
 // getting the fixed buffer index returned to the available pool kept by the registry.
-async fn echo_handler(stream: TcpStream, peer: SocketAddr, registry: FixedBufRegistry) {
+async fn echo_handler<T: IoBufMut>(
```
FrankReh, line 53: Just a side note. I added this example long before there was a pool layer built on top of fixed buffers and well before I understood how integration tests could be added to crate. It's interesting, for this PR, how just the signature of the handler function had to be modified. I guess that is a pretty standard occurrence in Rust creates.

src/buf/fixed/registry.rs
```diff
@@ -281,21 +290,23 @@ impl FixedBuffers for Inner {
...
-            match self.states[i] {
-                BufState::Free { init_len } => {
-                    let ptr = iovec.iov_base as *mut u8;
-                    let cap = iovec.iov_len;
-                    let v = unsafe { Vec::from_raw_parts(ptr, init_len, cap) };
-                    mem::drop(v);
+        for (i, state) in self.states.iter().enumerate() {
+            match state {
+                BufState::Free { init_len, .. } => {
+                    // Update buffer initalisation.
```
FrankReh, line 298: typo

src/buf/fixed/registry.rs
```diff
@@ -281,21 +290,23 @@ impl FixedBuffers for Inner {
...
+                    // The buffer is about to dropped, but this may release it
```
FrankReh, line 299: .. about to *be* dropped, ..

src/buf/fixed/pool.rs
FrankReh: Just a comment. It's interesting this `buffers` field is used purely to be able to call set_init for each usable IoBufMut in the Inner drop.

FrankReh (changes requested):
It's looking good. A lot of brain cycles have been put into this little corner of the crate.

src/buf/fixed/handle.rs
```diff
@@ -44,59 +42,51 @@ impl Drop for FixedBuf {
...
 
 unsafe impl IoBuf for FixedBuf {
     fn stable_ptr(&self) -> *const u8 {
-        self.buf.as_ptr()
+        self.buf.iovec.iov_base as _
     }
 
     fn bytes_init(&self) -> usize {
-        self.buf.len()
+        self.buf.init_len
```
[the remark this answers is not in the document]
oliverbunting: I, rightly or wrongly, tend to use bytes-int to give me the number of valid bytes. I.e, after a read operation.

src/buf/fixed/handle.rs
```diff
@@ -105,21 +95,22 @@ impl Deref for FixedBuf {
     type Target = [u8];
 
     fn deref(&self) -> &[u8] {
-        &self.buf
+        unsafe { std::slice::from_raw_parts(self.buf.iovec.iov_base as _, self.buf.init_len) }
```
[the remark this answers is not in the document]
oliverbunting: Sure

src/buf/fixed/handle.rs
```diff
@@ -44,59 +42,48 @@ impl Drop for FixedBuf {
...
     fn stable_mut_ptr(&mut self) -> *mut u8 {
-        self.buf.as_mut_ptr()
+        self.buf.iovec.iov_base as _
     }
 
     unsafe fn set_init(&mut self, pos: usize) {
-        if self.buf.len() < pos {
-            self.buf.set_len(pos)
+        if self.buf.iovec.iov_len >= pos {
+            self.buf.init_len = pos
```
FrankReh, line 86: I'm still confused by this, meaning it doesn't look right to me. Am I the one looking at this wrong? The original code this PR is replacing was growing the value of the Vec<u8> length by comparing the new pos to the vec length (the vec's length was being used to track the amount initialized). This appears to be using the io_vec's length as the comparison which isn't really the question, the io_vec's length serves as the buffer's capacity I believe. Shouldn't it read
```rust
if self.buf.init_len < pos {
    self.buf.init_len = pos
}
// or
self.buf.init_len = cmp::max(self.buf.init_len, pos);
```

src/buf/fixed/handle.rs
```diff
@@ -105,21 +92,24 @@ impl Deref for FixedBuf {
     type Target = [u8];
 
     fn deref(&self) -> &[u8] {
-        &self.buf
+        // Safety: The iovec points to a slice held  in self.buffers, to which no mutable reference exists.
```
FrankReh, line 95: Thanks for the safety comments. I don't think I would have come up with these reasons as I'm not as focused on the self.buffers holding the ownership but I can't say the comments are incorrect.

FrankReh (changes requested):
One unresolved comment left I think.

src/buf/fixed/handle.rs
ollie-etl: The check I've put in is simply a bounds check, i.e, that its valid to set the bytes_init to such the given value. The previous version had no such bounds check. Te previous one _is_ consisnte with the other instances of IoBuf in the crate, of which there are 2:

https://github.com/tokio-rs/tokio-uring/blob/d2f1494d1ab92bd7355a2af6069dd5e4ad5f3415/src/buf/io_buf_mut.rs#L37-L47

https://github.com/tokio-rs/tokio-uring/blob/d2f1494d1ab92bd7355a2af6069dd5e4ad5f3415/src/buf/io_buf_mut.rs#L50-L60

I'm going to argue again that not allowing a buffer to mark fewer bytes to be marked init after a read is just weird.

1. Firstly, we're dealing in `u8`'s. I'm not convinced that there is such as thing as an uninitialized u8. I'll await someone to point out some deeply unintuitive optimization llvm does here...

2. What if I want to use [std::ptr::copy](https://doc.rust-lang.org/std/ptr/fn.copy.html) between buffers? I can't mark the target buffer as less init, although it is possible for bytes from the src to be "un-init"

3. Its really unintuitive that after some read operation `bytes_init` doesn't reflect the value actually read. Certainly, looking at any of the read-like ops, i.e
 
https://github.com/tokio-rs/tokio-uring/blob/d2f1494d1ab92bd7355a2af6069dd5e4ad5f3415/src/io/read.rs#L55-L60

you'd be forgiven for assuming the buffer init value was the number of bytes actually read. 

4. Its not repeatable. If i read hello world from a file into 2 different buffers which may have been previously used, I've no idea what the bytes_init value would be, and they certainly have no guarantees of being equal.


If I'm told I'm just wrong, I'll change this, but I still hold its not a sane API.

ollie-etl: Actually, I'm irritated enough by this to issue a PR. It can have a discussion in its own right

ollie-etl: Will change here, discuss there

PR #214, opened 2023-01-19
Add support for `fallocate` and `statx`

hellertime:
This PR adds support for both the `IORING_OP_FALLOCATE` and the `IORING_OP_STATX` opcodes, and extends `File` to expose `fallocate()` and `statx()` methods. 

A basic test which uses both `fallocate` and `statx` is included as well.

### DIFF
--- a/src/io/statx.rs
+++ b/src/io/statx.rs
@@ -21,7 +21,7 @@ impl Op<Statx> {
             x.handle().expect("not in a runtime context").submit_op(
                 Statx {
                     fd: fd.clone(),
-                    statx: unsafe { Box::new(std::mem::zeroed()) },
+                    statx: Box::new(unsafe { std::mem::zeroed() }),
                 },
                 |statx| {
                     opcode::Statx::new(
--- a/src/io/statx.rs
+++ b/src/io/statx.rs
@@ -1,0 +1,48 @@
+use std::{ffi::CStr, io};
+
+use io_uring::{opcode, types};
+
+use crate::runtime::{
+    driver::op::{Completable, CqeResult, Op},
+    CONTEXT,
+};
+
+use super::SharedFd;
+
+pub(crate) struct Statx {
+    fd: SharedFd,
+    statx: Box<libc::statx>,
+}
+
+impl Op<Statx> {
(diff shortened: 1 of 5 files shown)


src/io/statx.rs
```diff
@@ -0,0 +1,48 @@
...
+}
+
+impl Op<Statx> {
+    pub(crate) fn statx(fd: &SharedFd) -> io::Result<Op<Statx>> {
+        CONTEXT.with(|x| {
+            let empty_path = CStr::from_bytes_with_nul(b"\0").unwrap();
+            x.handle().expect("not in a runtime context").submit_op(
+                Statx {
+                    fd: fd.clone(),
+                    statx: unsafe { Box::new(std::mem::zeroed()) },
```
FrankReh, line 24: Confused by this unsafe and the Box. Why is it needed? Should it come with a safety contract comment or am I being dense today? I'll admit to not having looked at the opcode::State::new implementation in the other crate yet.

FrankReh: Okay, I'm guessing there needs to be some memory that is ready for the kernel to write into, and Box is our preferred way, allowing the future to be moved around?

FrankReh (changes requested):
Looks reasonable to me. @Noah-Kennedy ?

src/io/statx.rs
FrankReh: Okay. Sorry but would you mind swapping the unsafe and the Box, so it has the same structure as the places we start with a boxed msghdr? Rest looks good.

FrankReh (changes requested):

src/io/statx.rs
hellertime: Sure. That's no problem to change.

FrankReh:
Also, @hellertime, have you seen this work in your setup?

FrankReh (approved):
Nice comments and thanks for the unit test too.

hellertime:
I've run the tests cases on my hardware (AMD EPYCs running Ubuntu 22.04) , if thats what you mean. I've not yet integrated this into the code I was porting though.

PR #215, opened 2023-01-23
API: Fix unintuitive / broken `set_init()` behaviour

ollie-etl:
The original discussion came up here https://github.com/tokio-rs/tokio-uring/pull/213#discussion_r1084501623 I'll rephase my arguments here:

Having `bytes_init()` only ever go up is broken. Consider the following example for copying data from a socket into a file

``` rust
let mut buf = Vec::with_capacity(2048);
loop {
    b = socket.recv(buf).await.1;
    let _ = target.write_at(b,0).await;
}
```

This example is broken,  because the `set_init` function is a one way ratchet, which can never be unset. It is called in `recv`, and all other read-like apis which fill buffers, to set the number of bytes read. However, set_init, in the current implmentation, ignores this value if its less than the previous number of initialised bytes. Unfortunately, large parts of write-like functions in the current API (and me) assume that `bytes_init` is equivalent to _valid bytes of interest_. As it stands `IoBuf` has no concept of length, only of bytes which have historically been valid (historical max len).

It is never possible to call a write-like function with a reused buffer. The IoBuf does not carry information on how many bytes to write.

This PR gets rid of the lifetime initialization counter, and replaces it with a the meaning of _currently initialized bytes_. I.e, after a read, the number of initialized bytes is the number read.

### DIFF
--- a/src/buf/fixed/handle.rs
+++ b/src/buf/fixed/handle.rs
@@ -95,9 +95,7 @@ unsafe impl IoBufMut for FixedBuf {
     }
 
     unsafe fn set_init(&mut self, pos: usize) {
-        if self.buf.len() < pos {
-            self.buf.set_len(pos)
-        }
+        self.buf.set_init(pos)
     }
 }
 
--- a/src/buf/io_buf_mut.rs
+++ b/src/buf/io_buf_mut.rs
@@ -23,14 +23,12 @@ pub unsafe trait IoBufMut: IoBuf {
 
     /// Updates the number of initialized bytes.
     ///
-    /// If the specified `pos` is greater than the value returned by
-    /// [`IoBuf::bytes_init`], it becomes the new water mark as returned by
-    /// `IoBuf::bytes_init`.
-    ///
     /// # Safety
     ///
     /// The caller must ensure that all bytes starting at `stable_mut_ptr()` up
     /// to `pos` are initialized and owned by the buffer.
+    ///
+    /// The number of bytes must be less than or equal to bytes_total()
     unsafe fn set_init(&mut self, pos: usize);
 }
 
@@ -40,9 +38,7 @@ unsafe impl IoBufMut for Vec<u8> {
     }
 
     unsafe fn set_init(&mut self, init_len: usize) {
-        if self.len() < init_len {
-            self.set_len(init_len);
-        }
+        self.set_len(init_len);
     }
 }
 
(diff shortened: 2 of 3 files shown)


ollie-etl:
@FrankReh @carllerche @Noah-Kennedy @mzabaluev  . Many opinions wanted on this

mzabaluev:
You can get the number of bytes that are actually read from the return value of the `recv` operation, and slice the data of the buffer with it. See the examples for how the APIs should be used.
You should not ignore the number of bytes returned by `write_at`, either, because it can be less than the number of bytes in the input buffer/slice.
I agree that this API is not intuitive, but this is what we have to account for two separate concerns:

* Tracking the length of initialized bytes in the buffer. There is no reason to ever mark bytes as uninitialized after they've been initialized, and the only purpose of this water mark is to prevent safe access into uninitialized data.
* Tracking the length of the data filled by the last read operation.

See [`ReadBuf`](https://docs.rs/tokio/1.24.2/tokio/io/struct.ReadBuf.html) in tokio as an API that keeps track of both.

mzabaluev:
> When a buffer is going to be used many times, it has to be better for throughput if it were to be zeroed once and then reused with no init watermark using data space and code space.

I think it would make sense for fixed buffers at least. However, this is not what the `IoBuf/IoBufMut` abstraction provides in general. And there may be applications that do not reuse buffers, but allocate them from the heap once per each I/O operation. I feel that this case should be fully supported with no degradation in performance (assuming the allocator is not a bottleneck there).

mzabaluev:
Also yes, there should be an issue to track this as an architectural change.

ollie-etl:
> 
> I think it would make sense for fixed buffers at least. However, this is not what the IoBuf/IoBufMut abstraction provides in general. And there may be applications that do not reuse buffers, but allocate them from the heap once per each I/O operation. I feel that this case should be fully supported with no degradation in performance (assuming the allocator is not a bottleneck there).

The change in API proposed would make no difference the single use case

FrankReh:
> I feel that this case should be fully supported with no degradation in performance (assuming the allocator is not a bottleneck there).

Fair enough. I know it's just my opinion. (I use that term a lot for this repo's issues.)

FrankReh:
> I cannot think of a single implementation where unsafety would be caused by reverting the watermark to the current read level.

Not brought up before but another use for reading into buffers is reading into buffer slices. Like reading into a Vec slice doesn't change the Vec's length, our reads don't need to change a property of the buffer itself. There is a return argument for the length. True other io calls can just return the length and don't have to return the buffer.

PR #219, opened 2023-02-01
[buf/fixed] FixedBufAllocator implementation

hellertime:
This PR introduces the concept of a `FixedBufAllocator`.

I wanted to provide a third option for obtaining `FixedBuf` from the `tokio-uring` runtime. While both the `FixedBufRegistry` and the `FixedBufPool` are excellent choices when it is known precisely what sized buffers you will need upfront. In times when that isn't well known, falling back on dynamic memory management is a good option.

This PR makes some modifications to the internals of the `FixedBuf` so that the drop behavior makes sense wether we need to "check-in" the buffer or "free" it.

There is also a test case included to exercise the allocator.

One limitation of the allocator approach at the moment is we cannot grow a heap beyond a single registered `iovec`. There may be a way around this limitation, if we can ensure allocations never cross certain byte addresses (if thats possible we can just map multiple `iovec` to contiguous ranges on the heap), but for now heaps are limited to ~1GiB (the allocator itself consumes a bit of the heap for alignment and bookkeeping).

### DIFF
--- a/src/buf/fixed/allocator.rs
+++ b/src/buf/fixed/allocator.rs
@@ -1,0 +1,218 @@
+use std::{alloc::Layout, cell::RefCell, io, ptr::NonNull, rc::Rc};
+
+use buddy_alloc::{buddy_alloc::BuddyAlloc, BuddyAllocParam};
+use libc::iovec;
+
+use crate::runtime::CONTEXT;
+
+use super::{buffers::AllocatableBuffers, handle::AllocatedBuf, FixedBuf, FixedBuffers};
+
+/// An allocateable I/O buffer heap pre-registered with the kernel.
+///
+/// `FixedBufAllocator` allows the application to manage a buffer heap
+/// allocated in memory, that can be registered in the current `tokio-uring`
+/// context using the [`register`] method. Unlike [`FixedBufRegistry`],
+/// individual buffers are not retrieved by index; instead, an available
+/// buffer matching a specified capacity can be retrieved with the [`allocate_fixed`]
+/// method. This allows some flexibility in managing buffers without pre-selecting
+/// the desired capacities required, at the expense of some bookeeping overhead.
+///
+/// A `FixedBufAllocator` value is a lightweight handle for a buffer heap.
+/// Cloning of a `FixedBufAllocator` creates a new reference to
+/// the same buffer heap.
+///
+/// The buffers of the collection are not deallocated until:
+/// - all `FixedBufAllocator` references to the heap have been dropped;
+/// - all [`FixedBuf`] handles to individual buffers in the heap have
+///   been dropped, including the buffer handles owned by any I/O operations
+///   in flight;
+/// - The `tokio-uring` [`Runtime`] the buffers are registered with
+///   has been dropped.
+///
+/// [`register`]: Self::register
+/// [`allocate_fixed`]: Self::allocate_fixed
+/// [`Runtime`]: crate::Runtime
+#[derive(Clone)]
+pub struct FixedBufAllocator {
+    inner: Rc<RefCell<Inner>>,
+}
+
+impl FixedBufAllocator {
+    /// Creates a new buffer heap of the requested size.
+    ///
+    /// Currently this is limited to the size of a single iovec that can
+    /// be passed to `io_uring` which limits us to 1GiB of heap per allocator.
+    ///
+    /// There is a hard limit to the heap of [`RLIMIT_MEMLOCK`], of course
+    /// this is not a per-runtime setting, so the practical limit will possibly be
+    /// lower.
+    ///
+    /// [`RLIMIT_MEMLOCK`]: nix::sys::resource::Resource::RLIMIT_MEMLOCK
+    pub fn new(size: usize) -> io::Result<Self> {
+        // the buddy allocator will use some of the space in the buffer for metadata
+        // and so the layout.size() will be greater than allocator.available_bytes()
+        let layout = Layout::from_size_align(size, 4096)
+            .map_err(|err| io::Error::new(io::ErrorKind::InvalidInput, err))?;
+        let heap = unsafe {
+            let data = std::alloc::alloc(layout) as *mut u8;
+            NonNull::new(data).unwrap()
+        };
+        Ok(FixedBufAllocator {
+            inner: Rc::new(RefCell::new(Inner::with_heap(heap, layout))),
+        })
+    }
+
+    /// Available bytes in the allocator
+    pub fn available_bytes(&self) -> usize {
+        self.inner.borrow().allocator.available_bytes()
+    }
+
+    /// Registers the buffer heap with the kernel.
+    ///
+    /// This method must be called in the context of a `tokio-uring` runtime.
+    /// The registration persists for the lifetime of the runtime, unless
+    /// revoked by the [`unregister`] method. Dropping the
+    /// `FixedBufAllocator` instance this method has been called on does not revoke
+    /// the registration or deallocate the buffers.
+    ///
+    /// [`unregister`]: Self::unregister
+    ///
+    /// This call can be blocked in the kernel to complete any operations
+    /// in-flight on the same `io-uring` instance. The application is
+    /// recommended to register buffers before starting any I/O operations.
+    ///
+    /// # Errors
+    ///
+    /// If collection of buffers is currently registered in the context
+    /// of the `tokio-uring` runtime this call is made in, the function returns
+    /// an error.
+    pub fn register(&self) -> io::Result<()> {
+        CONTEXT.with(|x| {
+            x.handle()
+                .as_ref()
+                .expect("Not in a runtime context")
+                .register_buffers(Rc::clone(&self.inner) as _)
+        })
+    }
+
+    /// Unregisters this buffer heap.
+    ///
+    /// This method must be called in the context of a `tokio-uring` runtime,
+    /// where the buffers should have been previously registered.
+    ///
+    /// This operation invalidates any `FixedBuf` handles checked out from
+    /// this registry instance. Continued use of such handles in I/O
+    /// operations may result in an error.
+    ///
+    /// # Errors
+    ///
+    /// If another collection of buffers is currently registered in the context
+    /// of the `tokio-uring` runtime this call is made in, the function returns
+    /// an error. Calling `unregister` when no `FixedBufPool` is currently
+    /// registered on this runtime also returns an error.
+    pub fn unregister(&self) -> io::Result<()> {
+        CONTEXT.with(|x| {
+            x.handle()
+                .as_ref()
+                .expect("Not in a runtime context")
+                .unregister_buffers(Rc::clone(&self.inner) as _)
+        })
+    }
+
+    /// Returns a buffer of requested capacity from this allocator
+    /// that is not currently owned by any other [`FixedBuf`] handle.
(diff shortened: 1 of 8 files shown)


hellertime:
It appears the CI environment has a more restrictive `RLIMIT_MEMLOCK` than my development host. Is it possible to see what the current limit is/raise that limit somehow?

src/buf/fixed/allocator.rs
```diff
@@ -0,0 +1,214 @@
...
+
+impl FixedBufAllocator {
+    /// Creates a new buffer heap of the requested size.
+    pub fn new(size: usize) -> io::Result<Self> {
+        // the buddy allocator will use some of the space in the buffer for metadata
+        // and so the layout.size() will be greater than allocator.available_bytes()
+        let layout = Layout::from_size_align(size, 4096)
+            .map_err(|err| io::Error::new(io::ErrorKind::InvalidInput, err))?;
+        let heap = unsafe {
+            let data = std::alloc::alloc(layout) as *mut u8;
```
oliverbunting, line 45: Did you consider anonymous mmap here instead.

hellertime: I'm not sure I know what the advantages/disadvantages are here. What would using the anonymous mmap gain us?

FrankReh:
Thank you for providing a series of commits to break up the functionality changes into logic units. Should make it easier to understand but have to admit, I don't have the powers of concentration at the moment to understand the implications of the first commit. I understand it makes the API more general without adding anything yet - which is great.

So will it make it general in the way that @mzabaluev and @ollie-etl and I have always hoped? So other registries or pools can be dropped in? Would it allow the second commit to done in an outside crate?

And is this entire PR a core piece that should come with fixed buffers or is this now the third incarnation that could or should have been put into an `extras` crate. I felt like we should have an extras crate even for the first fixed buffer registry, at least that's what I recall, and with every review and prolonged discussion, I wish more and more that's what we had done.

It's  possible the first commit is extending the API in a core way that is useful and the commit that wants to pull in a new allocation crate is a prime example of an opinion that many might share, but that wouldn't make it into a `tokio` sponsored repo.
As I've said at other times, a new crate being introduced to `tokio-uring` probably won't fly if it isn't already used by `tokio`. No one more central to `tokio` corrected that line of reasoning (maybe didn't see it).

And while I want to get PRs committed as soon as possible when they don't have conflicts, we already have two PRs centered on fixed buffers that need to be resolved first. Doesn't mean we shouldn't review and comment now if possible.

Sorry if this came across as too blunt. It wasn't my intent but I know the text-only communication we have with this medium doesn't show my true interest and I didn't want to wait a whole day before responding. Contributions are valuable and the more people who get good at understanding things like the fixed buffer registry and pool there are, the better. So already it's impressive.

hellertime:
> Thank you for providing a series of commits to break up the functionality changes into logic units. Should make it easier to understand but have to admit, I don't have the powers of concentration at the moment to understand the implications of the first commit. I understand it makes the API more general without adding anything yet - which is great.
> 
> So will it make it general in the way that @mzabaluev and @ollie-etl and I have always hoped? So other registries or pools can be dropped in? Would it allow the second commit to done in an outside crate?

This is a great question! The commit does make it easier to develop other registries within the `tokio-uring` crate, it does not expose enough power to build a new registry on top of the `tokio-uring` crate. The issue comes down to the implementation of `FixedBuf`, its not possible for a user of `tokio-uring` to construct a new `FixedBuf` directly (as we must here in the allocator changes in the subsequent commit: https://github.com/tokio-rs/tokio-uring/pull/219/commits/2834ed59a9f50c96ba47bb0be1fccc8a4c5f4101#diff-e8d535e190483efc25b4b518284f1e99e0e20add5c0191f2ebb462638fffc9dcR139), as `new()` is not public, any registry needs to return `FixedBuf` to meet the call signature of any of the "fixed" calls.

I do actually have another implementation on a branch that creates an entirely new pair of `BoundedFixedBut/Mut` types and associated `IoBufFixed/Mut` traits which expose the `buf_index()` needed for "fixed" calls.

However I prefer this implementation as it feels unclean to let users create types which expose a "buf_index" arbirarily.

It also is a much more intrusive patch-set, as every call that touches a fixed call is currently bound to `BoundedBufMut<BufMut=FixedBuf>`.

BUT! I would love to have a cleaner way to implement this outside of `tokio-uring`. I have no idea what that API shape would be just yet!

> 
> And is this entire PR a core piece that should come with fixed buffers or is this now the third incarnation that could or should have been put into an `extras` crate. I felt like we should have an extras crate even for the first fixed buffer registry, at least that's what I recall, and with every review and prolonged discussion, I wish more and more that's what we had done.

An extras crate for the registries makes sense only if there is at least one way to obtain `FixedBuf` in the default crate. 

> 
> It's possible the first commit is extending the API in a core way that is useful and the commit that wants to pull in a new allocation crate is a prime example of an opinion that many might share, but that wouldn't make it into a `tokio` sponsored repo. As I've said at other times, a new crate being introduced to `tokio-uring` probably won't fly if it isn't already used by `tokio`. No one more central to `tokio` corrected that line of reasoning (maybe didn't see it).

I don't have a good grounding here, so can't add much. I do understand the desire for minimal dependencies. If we could find an approach to expose access to registered buffers in a way that is safe(ish) and clean, it would be nice to rip out all the registry/allocator logic.

> 
> And while I want to get PRs committed as soon as possible when they don't have conflicts, we already have two PRs centered on fixed buffers that need to be resolved first. Doesn't mean we shouldn't review and comment now if possible.

Mind linking these into this PR, it feels reasonable to consider these blocking changes, or at least question if they should be in this PR.

> 
> Sorry if this came across as too blunt. It wasn't my intent but I know the text-only communication we have with this medium doesn't show my true interest and I didn't want to wait a whole day before responding. Contributions are valuable and the more people who get good at understanding things like the fixed buffer registry and pool there are, the better. So already it's impressive.

No worries, again the initial question is a great one. I am eager to make use of this code, so I would be happy to see the PR land. But I can continue to leverage the `FixedBufPool` for now while I wait.

FrankReh:
@hellertime @ollie-etl @mzabaluev Nibbling a little at a time ... are we in agreement that to allow external creation of fixed buffers, we have to define a trait that gives us the fixed buf index when we need it?

Somehow I thought the question could boil down to a general allocator and iterator for setup and drop for deallocation and the indexing would be the purview of the tokio_uring crate. But maybe allocation can't always be known up front?

There is the `provided buffers` feature of urings also that we should take into account. Does continuing to extend abilities with `fixed buffers` come at an unnecessary cost when _simply_ starting to support `provided buffers` would be solving more with less? I seem to recall the `provided buffers` come with a group_id and there can be many groups of provided buffers, each with their own group_id.

I'm much more familiar with the `buf_ring` feature at the moment but they are best suited for the multi read  operations which we will eventually get to I think. The earlier uring buffer options are a little complicated without going over the uring man pages again because those same man pages never try to lay out the pros and cons of each scheme. They just seem to fold new man pages along with new features as the uring device evolves.

I know `fixed buffers` was the first to be supported so would cover more kernels, but if we keep trying to support the earliest kernels, we will continue to make very slow progress.

hellertime:
apologies. I unintentionally closed this PR.

rrichardson:
> There is the provided buffers feature of urings also that we should take into account. Does continuing to extend abilities with fixed buffers come at an unnecessary cost when simply starting to support provided buffers would be solving more with less?  

It seems that going with provided buffers would allow the use of any old Buffer wrapper.  It wouldn't know or care about an index.  Using the IORING_REGISTER_PBUF_RING approach would,  I think, not even require a buffer to know its buffer group ID.  It'd just need to know how to recycle itself back into the allocator/pool whence it came, to be handed back to the ring. 

That said,  PBUF_RING makes this allocator scheme even more valuable,  since it would allow us to add new allocation groups at runtime.

hellertime:
> @hellertime @ollie-etl @mzabaluev Nibbling a little at a time ... are we in agreement that to allow external creation of fixed buffers, we have to define a trait that gives us the fixed buf index when we need it?
> 
> Somehow I thought the question could boil down to a general allocator and iterator for setup and drop for deallocation and the indexing would be the purview of the tokio_uring crate. But maybe allocation can't always be known up front?
> 
> There is the `provided buffers` feature of urings also that we should take into account. Does continuing to extend abilities with `fixed buffers` come at an unnecessary cost when _simply_ starting to support `provided buffers` would be solving more with less? I seem to recall the `provided buffers` come with a group_id and there can be many groups of provided buffers, each with their own group_id.
> 
> I'm much more familiar with the `buf_ring` feature at the moment but they are best suited for the multi read operations which we will eventually get to I think. The earlier uring buffer options are a little complicated without going over the uring man pages again because those same man pages never try to lay out the pros and cons of each scheme. They just seem to fold new man pages along with new features as the uring device evolves.
> 
> I know `fixed buffers` was the first to be supported so would cover more kernels, but if we keep trying to support the earliest kernels, we will continue to make very slow progress.

If I understand the use case of both the Provided Buffers and PBUF_RING these are really useful for reads, but don't support writes. So until such support is added (if ever) I think the fixed buffers have a place in the tokio-uring ecosystem.

As for focusing on FixedBuf over the other implementations, I agree it shouldn't take up all the bandwidth. 

As for the FixedBuf implementation itself, I'd rather not develop a trait that exposes the `buf_index()` as that feels like it opens up an avenue for abuse.

It *would* be interesting if we could split off a new `FixedBuf` from an existing `FixedBuf`.

For instance, if we could do the following:

```rust
use tokio_uring::buf::fixed::FixedBufRegistry
use buddy_alloc::buddy_alloc::BuddyAlloc;
use buddy_alloc::BuddyAllocParam;

let registry = FixedBufRegistry::new(vec![0u8;1<<20]);
tokio_uring::start(async{
    registry.register()?;
    let fixed_buf = registry.check_out(0).unwrap();
    let mut alloc = unsafe { BuddyAlloc::new(BuddyAllocParam::new(fixed_buf.stable_ptr(), fixed_buf.bytes_total(), 4096)) };
    let ptr = alloc.malloc(32);
    let fixed_buf2 = fixed_buf.split_fixed_buf().addr(ptr).len(32).dtor(|ptr| alloc.free(ptr)).build().unwrap();
    ...
    Ok(())
}
```

The key point here is the builder would only succeed if `addr` and `len` were contained within `stable_ptr()` and `bytes_total()`. The new `FixedBuf` would use the `dtor` on drop, and so it wouldn't "check in" to the registry that created it. It would be something akin to a `Slice` but not a different type. It would inherit the same index as the `FixedBuf` it was split from. And it would allow creating this allocator instance externally to the `tokio-uring` crate.

I'm not even sure this requires the Registry to exist, but for now thats how we obtain the initial `FixedBuf` containing the entire `iovec` that was registered.

oliverbunting:
I don't have the bandwidth to comment currently, sorry

FrankReh:
Thanks Chris. I appreciate the help in trying to keep the submissions, their order, and ourselves sane. Feel free to weigh in on other discussion and you can always open this again or even just ask to look more closely at that first commit again. I was going to do that later today in order to better understand the kind of generality you were looking for with it.

I don't expect it to happen, but I would pleasantly surprised if we can get all the uring enabled buffer types into a scheme that everyone is happy with that fits into the `Completable` async model people have talked about for a few years now. This crate suffers by being at the cross-roads in a few ways, one of them being can it ever be slipped into higher level functionality like `hyper` and also support all the special use cases for the `io_uring` device that people who want to work with `tokio` could dream up.

PR #220, opened 2023-02-01
add writev_all to io modules

FrankReh:
And add File::writev_at_all as the first user facing example.

The writev_all function will call the io_uring writev operation as many times as necessary to get all the bytes written or to hit an error, which ever comes first.

### DIFF
--- a/src/fs/file.rs
+++ b/src/fs/file.rs
@@ -292,14 +292,15 @@ impl File {
         op.await
     }
 
-    /// Like `writev_at` but will call the `io_uring` `writev` operation multiple times if necessary.
+    /// Like `writev_at` but will call the `io_uring` `writev` operation multiple times if
+    /// necessary.
     ///
-    /// Parameter `pos` is an Option<u64> to allow this function to be used for both files that are
-    /// seekable and those that are not. The caller is responsible for knowing this.
+    /// Parameter `pos` is an `Option<u64>` to allow this function to be used for both files that
+    /// are seekable and those that are not. The caller is responsible for knowing this.
     ///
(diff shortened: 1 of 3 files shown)


src/fs/file.rs
```diff
@@ -292,6 +292,59 @@ impl File {
...
+    /// TODO define writev_all functions for net/unix/stream, net/tcp/stream, io/socket.
+    /// TODO remove usize from result, to be consistent with other write_all_vectored functions.
+    /// TODO find a way to test this with some stress to the file so the writev calls don't all
+    /// succeed on their first try.
+    /// TODO consider replacing the current `write_all` and `write_all_at` functions with a similar
+    /// mechanism so all the write-all logic is in one place, in the io/write_all.rs file.
+    pub async fn writev_at_all<T: BoundedBuf>(
+        &self,
+        buf: Vec<T>,
+        pos: Option<u64>, // Use None for files that can't seek
```
mzabaluev, line 342: This looks like an API wart. I think most callers would prefer two different methods, the other being `writev_all` without an offset just like in other objects.

mzabaluev (commented):
I don't understand the need to introduce a non-elementary kind of `Op`, but I'm not up to date on this code. Wasn't it so, originally, that `Op` represents elementary SQEs that can be submitted to the kernel, and the more complex logic is realized in async fns? In this case the individual op would be `writev` and the async method would drive the procedure to completion, submitting individual operations in a loop.

FrankReh:
@mzabaluev The same function will be called from TcpStream and UnixStream so a common place like the io directory seemed reasonable to me.

Noah-Kennedy (approved):

PR #221, opened 2023-02-03
add second statx command that accepts flags and mask

FrankReh:


FrankReh:
Think I'm going to replace this with a StatxBuilder.

PR #222, opened 2023-02-03
fixed descriptor support

FrankReh:
Support uring fixed descriptors, often also called direct descriptors by the io_uring man pages.

### DIFF
--- a/src/io/shared_fd.rs
+++ b/src/io/shared_fd.rs
@@ -81,7 +81,8 @@ impl SharedFd {
         match self.inner.fd {
             CommonFd::Raw(raw) => raw,
             CommonFd::Fixed(_fixed) => {
-                unreachable!("fixed files aren't actually created yet");
+                // TODO Introduce the fixed option for file read and write first.
+                unreachable!("fixed file support not yet added for this call stack");
             }
         }
     }
--- a/src/io/shared_fd.rs
+++ b/src/io/shared_fd.rs
@@ -10,6 +10,7 @@ use std::{
 
 use crate::io::shared_fd::sealed::CommonFd;
 use crate::runtime::driver::op::Op;
+use crate::runtime::CONTEXT;
 
 // Tracks in-flight operations on a file descriptor. Ensures all in-flight
 // operations complete before submitting the close.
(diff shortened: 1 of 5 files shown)


src/io/shared_fd.rs
```diff
@@ -156,3 +198,58 @@ impl Drop for Inner {
...
+
+/// A file descriptor that has not been registered with io_uring.
+#[derive(Debug, Clone, Copy)]
+pub struct Raw(pub RawFd); // Note: io-uring names this Fd
+
+/// A file descriptor that has been registered with io_uring using
+/// [`Submitter::register_files`](crate::Submitter::register_files) or [`Submitter::register_files_sparse`](crate::Submitter::register_files_sparse).
+/// This can reduce overhead compared to using [`Fd`] in some cases.
+#[derive(Debug, Clone, Copy)]
+pub struct Fixed(pub u32); // TODO consider renaming to Direct (but uring docs use both Fixed descriptor and Direct descriptor)
```
Noah-Kennedy, line 212: Do we want to allow users to create their own fixed fds like this?

FrankReh: That's a good question I don't know the answer to. On one level, it's just a different kind of number, or different type of number. Like a user can create their own file descriptor values if they happen to want to.

On the other hand, if the user's don't need it, not providing certainly reduces their confusion when confronted so much new in an API.

There could be some APIs where the user decides how to partition the fixed table space - we can't be the arbiters for all uses. So I guess that's why I'm inclined to expose the type, for now. I still prefer thinking of this crate as a means to marry tokio with urings, and let the user decide what they will do with it.

I also think its okay to expose things on a trial bases reserving the right to change things in the next breaking version. Otherwise we might not have enough to reach a critical mass.

But if you think we can build what we need without this being public and we want to ask for someone to say they need to do their work, we can revisit the public nature of that type.

thomasbarrett:
Awesome! This is another feature I am looking for 😄. This also has interesting implications on the Send semantics of File which should probably be discussed. With normal fds, it is possible to make the File type Send. With fixed fds, I don't believe that the File type can be Send. Would it make sense to make two different File types (e.g. File and FixedFile), one of which uses normal Fds and is Send and one which uses Fixed Fds and is !Send?

FrankReh:
I'm not inclined to want two forms of File, FileFd and FileFixed, unless we find more reasons for letting the compiler differentiate one from the other. Maybe a different method set? Maybe a different lifetime or close semantics?

FrankReh:
Will have to resolve conflicts tomorrow.

FrankReh:
Trying to convert this to draft while it relies on my branch for the io-uring dependency.

src/io/shared_fd.rs
FrankReh: This is gone.

FrankReh:
Replaced by another PR, #256, which has the same code. The diffs against the mainline should be easier to follow.

PR #224, opened 2023-02-04
create_dir_all and DirBuilder

FrankReh:
uring, therefore async, versions of create_dir_all and DirBuilder.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -23,6 +23,7 @@ libc = "0.2.80"
 io-uring = { version = "0.5.12", features = ["unstable"] }
 socket2 = { version = "0.4.4", features = ["all"] }
 bytes = { version = "1.0", optional = true }
+futures = "0.3.26"
 
 [dev-dependencies]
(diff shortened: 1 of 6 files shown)


FrankReh:
For now, this comes with an example binary that was useful in development and testing of the feature. It doesn't make a create example for normal uring use though so I'm happy to consider it temporary and we can remove it as part of this PR or any time later.

Also there is an async version of `is_dir` that is used solely by this feature which may be useful in its own right one day or at least parts of it as it has to look at libc bits of the statx structure and asking whether a path or a File is a directory is fairly common so there are normally helper functions or methods provided for that. For now our `statx` call returns a libc::statx structure so the caller is own their own to track down the details of how to do it.

Noah-Kennedy (approved):

FrankReh:
FYI. I'm working on a builder version of statx and that will allow getting the statx structure without first having opened the file, so the `is_dir` function in this `create_dir_all` can be greatly simplified after that.

PR #232, opened 2023-02-08
fix: rework SharedFd

FrankReh:
This allows a 'file.close.wait' to return a true close return value. And it fixes the problem of indefinitely hanging when the SharedFd was still held by an in-flight operation.

It simplifies the low level close logic, using the async uring close operation when the caller is using file.close.wait but using the synchronous close mechanism provided by the std library when the SharedFd is allowed to go out of scope without having been closed first.

Manual closing of the file should be encouraged because it allows the user to see the return code.

In order to let the file owner call close and not hang when an in-flight was still in progress, this reworks the future to return when it sees the strong reference count drop to 1.

Fixes 
- #122.

### DIFF
--- a/src/fs/file.rs
+++ b/src/fs/file.rs
@@ -824,14 +824,17 @@ impl File {
         Op::statx(&self.fd)?.await
     }
 
-    /// Closes the file.
+    /// Closes the file using the uring asynchronous close operation and returns the possible error
+    /// as described in the close(2) man page.
     ///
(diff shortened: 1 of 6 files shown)


FrankReh:
I can add a test case that highlights the problem. I realized it doesn't require a large file being created. Even a small buffer is enough to trigger the Future being polled when that particular select! mechanism is used.

Noah-Kennedy (approved):

PR #233, opened 2023-02-08
Make msghdr.msg_name optional

redbaron:
sendmsg can be used on connected sockets without msg_name.

### DIFF
--- a/src/net/udp.rs
+++ b/src/net/udp.rs
@@ -261,6 +261,9 @@ impl UdpSocket {
     /// > it replaces per byte copy cost with page accounting and completion
     /// > notification overhead. As a result, zero copy is generally only effective
     /// > at writes over around 10 KB.
+    ///
+    /// Can be used with socket_addr: None on connected sockets, which can have performance
+    /// benefits if multiple datagrams are sent to the same destination address.
     pub async fn sendmsg_zc<T: BoundedBuf, U: BoundedBuf>(
         &self,
         io_slices: Vec<T>,
--- a/src/net/udp.rs
+++ b/src/net/udp.rs
@@ -52,7 +52,7 @@ use std::{
 ///         assert_eq!(b"hello world", &buf[..n_bytes]);
 ///
 ///         // write data using send on connected socket
-///         let (result, _) = socket.send_to(b"hello world via send".as_slice(), None).await;
+///         let (result, _) = socket.send(b"hello world via send".as_slice()).await;
 ///         result.unwrap();
 ///
 ///         // read data
@@ -82,7 +82,7 @@ use std::{
 ///         let buf = vec![0; 32];
 ///
 ///         // write data
-///         let (result, _) = socket.send_to(b"hello world".as_slice(), Some(second_addr)).await;
+///         let (result, _) = socket.send_to(b"hello world".as_slice(), second_addr).await;
 ///         result.unwrap();
 ///
 ///         // read data
@@ -172,7 +172,7 @@ impl UdpSocket {
     ///
     ///         // write data
     ///         let (result, _) = std_socket
-    ///             .send_to(b"hello world".as_slice(), Some(second_addr))
+    ///             .send_to(b"hello world".as_slice(), second_addr)
     ///             .await;
     ///         result.unwrap();
     ///
@@ -208,15 +208,25 @@ impl UdpSocket {
         self.inner.connect(SockAddr::from(socket_addr)).await
     }
 
+    /// Sends data on the connected socket
+    ///
+    /// On success, returns the number of bytes written.
+    pub async fn send<T: BoundedBuf>(
+        &self,
+        buf: T,
+    ) -> crate::BufResult<usize, T> {
+        self.inner.send_to(buf, None).await
+    }
+
     /// Sends data on the socket to the given address or to previously connected remote peer.
     ///
     /// On success, returns the number of bytes written.
     pub async fn send_to<T: BoundedBuf>(
         &self,
         buf: T,
-        socket_addr: Option<SocketAddr>,
+        socket_addr: SocketAddr,
     ) -> crate::BufResult<usize, T> {
-        self.inner.send_to(buf, socket_addr).await
+        self.inner.send_to(buf, Some(socket_addr)).await
     }
 
     /// Sends data on the socket. Will attempt to do so without intermediate copies.
@@ -251,8 +261,6 @@ impl UdpSocket {
     /// > it replaces per byte copy cost with page accounting and completion
     /// > notification overhead. As a result, zero copy is generally only effective
     /// > at writes over around 10 KB.
-    ///
-    /// Note: Using fixed buffers [#54](https://github.com/tokio-rs/tokio-uring/pull/54), avoids the page-pinning overhead
     pub async fn sendmsg_zc<T: BoundedBuf, U: BoundedBuf>(
         &self,
         io_slices: Vec<T>,
--- a/src/net/udp.rs
+++ b/src/net/udp.rs
@@ -72,7 +72,7 @@ use std::{
 ///         let buf = vec![0; 32];
 ///
 ///         // write data
-///         let (result, _) = socket.send_to(b"hello world".as_slice(), second_addr).await;
+///         let (result, _) = socket.send_to(b"hello world".as_slice(), Some(second_addr)).await;
 ///         result.unwrap();
 ///
 ///         // read data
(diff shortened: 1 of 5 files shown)


FrankReh:
These changes may be a step in the right direction but they are not complete. The branch doesn't compile.

Please run `cargo build` and `cargo test` on your branch before starting a PR. Also `cargo fmt` and `cargo clippy`. All with the most recent compiler. In this case, you can make additional commits to your branch. See all those commands pass and then push up to GitHub again.

Here, even the file src/io/sendmsg_zc.rs which you changed won't compile, let along the entire library not compiling.

And while your comment about sendmsg is correct, please include a link to an online manpage from some place like arch linux so anyone reading through this can be starting with the same understanding.

Thank you.

redbaron:
I am really sorry for incomplete PR, IDK how it happened. Pushed fixed version.

src/net/udp.rs
```diff
@@ -246,7 +246,7 @@ impl UdpSocket {
     pub async fn sendmsg_zc<T: BoundedBuf, U: BoundedBuf>(
```
redbaron, line 246: Regarding comment above:

> /// Note: Using fixed buffers [#54](https://github.com/tokio-rs/tokio-uring/pull/54), avoids the page-pinning overhead

I don't think it is true. If I read Linux kernel code correctly, fixed buffers can be used with send_zc only:
https://elixir.bootlin.com/linux/v6.1.10/source/io_uring/net.c#L983

it makes sense, SQE has only one buf_index, but msghdr can provide multiple buffers in msg_iov

FrankReh: You are right. Good catch. Please remove that comment line. I'll thoroughly review tomorrow morning.

By the way, is there an actual unit test that had used an address that wasn't necessary that you could copy and show that the None option works as intended?

redbaron: > By the way, is there an actual unit test that had used an address that wasn't necessary that you could copy and show that the None option works as intended? 

I updated one of the doctests, let me know if you would like to move it elsewhere

redbaron: > In fact, I don't see this crate supporting fixed buffers for send_zc

Yeah, there should be `send_fixed_zc` with  `BoundedBuf<Buf = FixedBuf>` and corresponding op, which prepares  IORUNG_SEND_ZC SQE with necessary flags set.

FrankReh: Would you mind opening two issues for the missing send_zc_fixed in io_uring and tokio_uring. I think we should follow the naming convention used in liburing when possible. In this case, they had chosen to add _fixed after the send_zc.

Doesn't mean you are on the hook for submitting PRs for them.

src/net/udp.rs
```diff
@@ -198,13 +208,13 @@ impl UdpSocket {
         self.inner.connect(SockAddr::from(socket_addr)).await
     }
 
-    /// Sends data on the socket to the given address.
+    /// Sends data on the socket to the given address or to previously connected remote peer.
```
FrankReh, line 211: Rather than modify this function to take an Optional address, please revert the changes to this comment and the argument and create a new function right above it, `send`. This will more closely mirror the Linux man pages for send and sendto.

The Optional address for the sendmsg_zc is good the way you've changed it in this PR.

This PR also highlights that we don't have a regular `sendmsg` for UDP. You don't have to do anything about that but you or I should open an Issue about that.

src/net/udp.rs
```diff
@@ -246,7 +256,7 @@ impl UdpSocket {
     pub async fn sendmsg_zc<T: BoundedBuf, U: BoundedBuf>(
         &self,
         io_slices: Vec<T>,
-        socket_addr: SocketAddr,
+        socket_addr: Option<SocketAddr>,
```
FrankReh, line 259: Please update the sendmsg_zc comment to say that None can be used when the socket has been connected and that it may be more efficient when sending multiple datagrams to the same address. I don't think you have to mention why it may be more efficient. I figure it's enough to trace through the code and see that a boxed value can be skipped for the uring operation and I haven't looked, but the kernel may also have already done its address parsing when the socket is connected so it could help on the kernel side too.

FrankReh (changes requested):
Looking good. Just asking for some small changes, all in net/udp.rs I think.

src/net/udp.rs
redbaron: You want `send` to not accept address and `send_to` to always require address, right?

FrankReh: Yes. That would mirror what I see from the Linux man page for send and sendto. Does that sound reasonable to you? I haven't missed either but I haven't had to send flags along with any send yet so I'm just looking for what makes the most sense at least in some vein. With all these ways of sending data, there's bound to be opinions.

FrankReh: Only make these changes to net/udp.rs. Both the send and send_to in that file can then use the common io::send_to as you've changed it to optionally take an address.

src/net/udp.rs
redbaron: Maybe create new struct UdpSocketConnected to be returned from UdpSocket::connect? It will have mostly same methods, but `sendmsg_zc` and `send_to` will be not taking `socket_addr` at all?

FrankReh: Please add that as a developer's comment so the idea isn't lost but I would rather not try to enforce this with a compiler. The io_uring device should return an error in those cases.

src/net/udp.rs
redbaron: It wont be different from `UdpSocket::write`, do we need it at all then? Maybe leave non ZC methods as they are?

src/net/udp.rs
redbaron: Compile time error is better than runtime. Isn't it better not to provide incorrect API at all, rather than documenting quirks and asking users to be careful?

src/net/udp.rs
FrankReh: I think we should support send_to and send and write for udp. Send and write use different uring opcodes and over time, we will have a builder for some or all of these operations to let the user get at more of the optional fields defined by the io-uring's opcodes and this is a good time to introduce send since it takes no address while send_to does. Even before we expose the options through a builder, the io_uring device offers different opcodes and we can let the user of this library decide which they prefer.

src/net/udp.rs
redbaron: https://github.com/tokio-rs/io-uring/issues/185
https://github.com/tokio-rs/tokio-uring/issues/235

src/net/udp.rs
redbaron: added `UdpSocket::send<T: BoundedBuf>(&self, buf: T) -> crate::BufResult<usize, T>`

src/net/udp.rs
redbaron: Added comment as  requested

src/net/udp.rs
```diff
@@ -198,15 +208,25 @@ impl UdpSocket {
...
+    ///
+    /// On success, returns the number of bytes written.
+    pub async fn send<T: BoundedBuf>(
+        &self,
+        buf: T,
+    ) -> crate::BufResult<usize, T> {
+        self.inner.send_to(buf, None).await
+    }
+
+    /// Sends data on the socket to the given address or to previously connected remote peer.
```
FrankReh, line 221: Since your comment here says `or to a previously connected remote peer`, want to add that `send` might be preferred in the case of a previously connected remote peer as it saves a box?

FrankReh (changes requested):

src/net/udp.rs
redbaron: I removed comment about connected sockets, because `UdpSocket::send_to`can't be used without SocketAdddr now

ollie-etl (approved):

FrankReh (approved):

FrankReh:
Thank you!

redbaron:
> Thank you!

I didn't realise you'd merge it straight away, my intention was to get approval then squash all changes. sorry for many useless commits

FrankReh:
We squash the changes for every PR in this repo. At least we mean to. There may have been some that slipped through. All the commits weren't a problem. The GitHub merge will do the squash for us if we like. I did end up pulling this branch locally to be able to look through things when there were problems. The final merge from master I didn't see needing to wait for you as it might have taken another day for the turn around. And there are a number of outstanding PRs that can get merged when another maintainer has a look so I wanted as much committed and remerged to the other PRs as quickly as possible.

PR #248, opened 2023-02-22
Implement IoBuf for Rc<T> where T: IoBuf

redbaron:
Using reference counted IoBuf allows to have multiple Slices for the same backing buffer, which is useful for situations where same buffer has to be written to multiple sockets


ollie-etl:
@redbaron Do you still have an interest in this merging? If so, could you update to pass `fmt`

PR #250, opened 2023-02-26
op: add recv_multi

FrankReh:
The first operation that supports streaming CQE results.

  Adds a Streamable trait, along the lines of the Completable trait.
    Comes with stream_next and stream_complete methods.

  Adds MultiCQEStream struct, along the lines of the MultiCQEFuture
  struct.

  Adds a submit_op_stream along the lines of submit_op.

  Adds poll_next_op along the lines of poll_op.

  The Lifecycle gets two additional methods: take_index, and data.

    take_index: returns the Lifecycle index and replaces it with
    usize::MAX to show the Lifecycle is no longer represented in the
    slab. This feature only used by the new poll_next_op.

    data: returns a reference to the Lifecycle's data, to be able to use
    it for the Streamable's stream_next calls which only require a
    reference. Ownership is still transferred in stream_complete.

Adds the io/recv_multi operation.

The tcp stream recv_multi is a function that
bridges the private types with a public function.

There is no cancel yet for the multishot command but the code can be
written to break out of the loop. Also, when the connection is closed,
the command should fail. It's not tested, but unregistering the buf_ring
might cancel the command too - but maybe not.

Unit tests:

  net unit tests and helper functions return io::Return

  add recv_multi cases

  add BufRingProps to net unit tests - to quiet empty buffer warnings
  which are intentional in some tests

### DIFF
--- a/src/io/recv_multi.rs
+++ b/src/io/recv_multi.rs
@@ -1,0 +1,87 @@
+use crate::buf::bufgroup::BufX;
+use crate::io::SharedFd;
+
+use crate::buf::bufring::BufRing;
+use crate::runtime::driver::op::{CqeResult, MultiCQEStream, Op, Streamable};
+use crate::runtime::CONTEXT;
+use std::io;
+
+pub(crate) type RecvMultiStream = Op<RecvMulti, MultiCQEStream>;
+
+pub(crate) struct RecvMulti {
+    /// Holds a strong ref to the FD, preventing the file from being closed
+    /// while the operation is in-flight.
+    _fd: SharedFd,
+
+    /// The bufgroup that supplies the bgid and the get_buf function.
+    group: BufRing,
+}
+
+impl RecvMultiStream {
+    pub(crate) fn recv_multi(
+        fd: &SharedFd,
+        group: BufRing,
+        flags: Option<i32>,
+    ) -> io::Result<Self> {
+        use io_uring::{opcode, types};
+
+        CONTEXT.with(|x| {
+            x.handle()
+                .expect("Not in a runtime context")
+                .submit_op_stream(
+                    RecvMulti {
+                        _fd: fd.clone(),
+                        group,
+                    },
+                    |s| {
+                        opcode::RecvMulti::new(types::Fd(fd.raw_fd()), s.group.bgid())
+                            .flags(flags.unwrap_or(0))
+                            .build()
+                    },
+                )
+        })
+    }
+}
+
+impl Streamable for RecvMulti {
+    type Item = io::Result<BufX>;
+
+    fn stream_next(&self, cqe: CqeResult) -> Self::Item {
+        // The kernel is not expected to provide an error when the `more` bit was set,
+        // but it's just as easy to handle the case here.
+        let res = cqe.result?;
+        let flags = cqe.flags;
+
+        // Safety: getting a buffer from the group requires the res and flags values accurately
+        // identify a buffer and the length which was written to by the kernel. The res and flags
+        // passed here are those provided by the kernel.
+        let get_buf_res = unsafe { self.group.get_buf(res, flags) };
+
+        match get_buf_res {
+            Ok(Some(bufx)) => Ok(bufx),
+            Ok(None) => {
+                unreachable!("unexpected combo of 'more', no error, and no buffer");
+            }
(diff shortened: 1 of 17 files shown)


FrankReh:
This starts as one commit over
- #245

So 245 should be reviewed first.

src/io/recv_multi.rs
```diff
@@ -0,0 +1,87 @@
...
+use crate::io::SharedFd;
+
+use crate::buf::bufring::BufRing;
+use crate::runtime::driver::op::{CqeResult, MultiCQEStream, Op, Streamable};
+use crate::runtime::CONTEXT;
+use std::io;
+
+pub(crate) type RecvMultiStream = Op<RecvMulti, MultiCQEStream>;
+
+pub(crate) struct RecvMulti {
```
ollie-etl, line 11: Have you thought about the generalized verions `recvmsg_multi` too? I'm curious how the API works when receiving a single packet into multiple buffers

FrankReh: I don't follow the question. You've seen Luca's excellent contribution to io-uring? 
- https://github.com/tokio-rs/io-uring/pull/192

Each recvmsg result is going into a single buffer, pulled from the buf_ring pool by the kernel. What am I missing?

ollie-etl: Does it unconditional _have_ too? [recvmsg](https://linux.die.net/man/2/recvmsg) takes an array of iovec. See PR #251 for the non multi-shot version, and also the existing sendmsg implementation. I read the docs around `io_uring_recvmsg_out` as an extra on top of the provided iovec specification

FrankReh: I really don't know where you are coming from with this. Can you refer to the liburing documentation to point out what you think I'm missing? The 251 PR doesn't seem relevant to how kernel io_uring implements multishot recvmsg at all?

ollie-etl: > I really don't know where you are coming from with this. Can you refer to the liburing documentation to point out what you think I'm missing? The 251 PR doesn't seem relevant to how kernel io_uring implements multishot recvmsg at all?

My apologies: #252 is relevant. #251 is completely irrelevant

FrankReh: If you linked to something by mistake, can you edit your comments to remove the link, at least the pound sign?

ollie-etl: > I updated the original link to point to the correct PR

FrankReh: And my point is the multi recvmsg doesn't do multiple buffers - it  couldn't. Luca made that clear with his investigation and his comments.

FrankReh: > I updated the original link to point to the correct PR

251 is still preceded by a hash in one of your comments. Now I'm totally wondering why all these PRs have to be linked with each other.

ollie-etl: > And my point is the multi recvmsg doesn't do multiple buffers - it couldn't. Luca made that clear with his investigation and his comments.

That may well be true, I'm really not an expert here, but I'm not able to see that spelt out definitely in that PR or the docs.

I have tried tracing the kernel source, but I'm none the wiser either way on where that limitation comes into force.

FrankReh: > The question was, have you considered how to cope with streaming operations where multiple buffers are provided in the original msghdr: that implies a collections of buffers

I have considered that collections of buffers are not part of the io_uring multishot receiving APIs so we don't need a streaming API for such an animal.

ollie-etl: > I have considered that collections of buffers are not part of the io_uring multishot receiving APIs so we don't need a streaming API for such an animal.

Thats a shame: that's a difference in API vs the non multishot API and the linux syscall, and precludes some of the memory locality use cases i spelt out in #252

oliverbunting: I don't wish to detract from this pr: I like this a great deal, especially the provided ring buffers required to make it work.

I'm looking at this API, as with most, with my use case in mind, which is UDP packet processing on a woefully underpowered aarch64 platform. Outperforming recv_mmsg is what brought me to this crate in the first place. 

Personally I'm not invested in tokio api compliance at all - I simply don't care if we completely diverge. The API I always compare to are the base syscalls

PR #252, opened 2023-02-27
Add recvmsg implmentation

ollie-etl:
Adds the missing `recvmsg` function to the udp socket.

the level of duplication required vs `recv_from` is uncomfortable, but required due to the lack of specialization.

### DIFF
--- a/src/io/mod.rs
+++ b/src/io/mod.rs
@@ -23,6 +23,8 @@ mod readv;
 
 mod recv_from;
 
+mod recvmsg;
+
 mod rename_at;
 
 mod send_to;
--- a/src/io/recvmsg.rs
+++ b/src/io/recvmsg.rs
@@ -1,0 +1,97 @@
+use crate::runtime::driver::op::{Completable, CqeResult, Op};
+use crate::runtime::CONTEXT;
+use crate::{buf::BoundedBufMut, io::SharedFd, BufResult};
+use socket2::SockAddr;
+use std::{
+    io::IoSliceMut,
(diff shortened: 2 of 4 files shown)


ollie-etl:
@FrankReh as you've just been looking at multi shot recv, this is probably quite current for you.

FrankReh:
I'm going to look at new  PRs once Noah has gotten to my queue of PRs and he may or may not say the same being in the middle of an API rework himself.

ollie-etl:
This is such a trivial PR: does it really have to be gated behind unbounded rework?

FrankReh:
> This is such a trivial PR

Even trivial PRs to review mean a cascade of main being remerged into other PRs. Tell me this PR allows you to do work that couldn't be done otherwise.

ollie-etl:
> Even trivial PRs to review mean a cascade of main being remerged into other PRs. Tell me this PR allows you to do work that couldn't be done otherwise.

Well, yes. It allows the user to receive a packet into separate buffers. If the user cares about the memory layout post receive, this is important. Say you want to  de-interleave internal packet header / payload, leaving multiple headers and multiple payloads in contiguous memory, then this is no alternative: there is no other API for splitting the memory location within a single recv call. This may not be a problem for TCP, which is pulling from a receive window buffer, but for UDP, recieve consumes the packet, I can't call it off on demand.

Noah-Kennedy:
I left the old APIs in so that people could still work during the refactor. The internal changes I am doing do not block this.

Noah-Kennedy (approved):

PR #255, opened 2023-02-28
op: poll and poll_multi

FrankReh:


FrankReh:
We still have to think about what kind of file to pass to the poll and poll_multi. Anything with a SharedFd will do. Once fixed file descriptors are supported, those can be passed also.

Suggestions are welcome.

ileixe:
Hi @FrankReh. We're looking for poll op and found your PR. I think poll() part at least is clear enough to introduce though, any plan do you have to proceed? (Review seems to be main blocker btw :innocent: )

FrankReh:
> any plan do you have to proceed?
No plan on my part with this. I've had to go my own way with this fork as a basis. If I see adoption and then find my own bugs, I would contribute them back. But for now, I think folks who are interested can try out the forked branches and then decide if they want to provide new PRs for this repo. Or they could take ownership of this PR or suggest it be closed for having gotten too old. The technology is very good (io-uring with Rust) - but it has never been clear how Tokio itself would allow this to be folded in, in a timely fashion, so I think forking Tokio is also called for by the most ardent.

One big thing that's changed since I tried to help here is that Hyper has released a 1.0 version that purports to be agnostic to  readiness based or completion based IO; so now there is a very valuable layer that could use an io-uring inspired Tokio. I haven't tried it myself yet but many crates have already moved to the hyper 1.0 api. I had created a Futures bridge, also suggested by withoutboats years ago I found, for putting io-uring into AsyncRead and AsyncWrite wrappers.

ileixe:
> One big thing that's changed since I tried to help here is that Hyper has released a 1.0 version that purports to be agnostic to readiness based or completion based IO; so now there is a very valuable layer that could use an io-uring inspired Tokio. I haven't tried it myself yet but many crates have already moved to the hyper 1.0 api. I had created a Futures bridge, also suggested by withoutboats years ago I found, for putting io-uring into AsyncRead and AsyncWrite wrappers.

That's interesting. So are you working with wrapper to implement AsyncRead and AsyncWrite without any runtime contexts? How then you integrated with uring submission and completion? We are looking for similar approach (runtime agnostic uring API) and appreciated if you can share idea
.

FrankReh:
@ileixe No, not without runtime contexts, with the tokio-uring and tokio contexts. So my take was dissimilar. I've gone the route of leveraging tokio and eventually hyper.

PR #260, opened 2023-03-20
bug: msghdr lifetime in SendMesgZc

ollie-etl:
Box msghdr in `SendMsgZc` struct to ensure it is stable through the lifetime of the op.


ollie-etl:
@Noah-Kennedy this should be uncontentious

Icelk:
@ollie-etl Have you managed to pass a `msg_control` larger than 15 bytes?

ollie-etl:
@Icelk I haven't tried. My interest in these is the vectored send / recv they provide.

PR #263, opened 2023-03-28
Implement opcode sendmsg (non-zc) for UdpStream.

### LINKED ISSUE #261: Support _msg syscalls
One can't add headers (shortened to hdr) when sending or receiving packets. This prevents QUIC implementations from using this library, as those require header control.

Would it be possible to give the user functions which take headers (e.g. [this opcode](https://docs.rs/io-uring/0.5.13/io_uring/opcode/struct.RecvMsg.html))?

quinn-rs/quinn#1452 seems to be blocked on this.

Icelk:
Fixes #261 

Should I maybe try to reduce the code duplication in `src/io/sendmsg_zc.rs` and `src/io/sendmsg.rs`?

### DIFF
--- a/src/io/sendmsg.rs
+++ b/src/io/sendmsg.rs
@@ -1,0 +1,103 @@
+use crate::buf::BoundedBuf;
+use crate::io::SharedFd;
+use crate::runtime::driver::op::{Completable, CqeResult, Op};
+use crate::runtime::CONTEXT;
+use socket2::SockAddr;
+use std::io;
+use std::io::IoSlice;
+use std::net::SocketAddr;
+
+pub(crate) struct SendMsg<T, U> {
+    _fd: SharedFd,
+    _io_bufs: Vec<T>,
+    _io_slices: Vec<IoSlice<'static>>,
+    _socket_addr: Option<Box<SockAddr>>,
+    msg_control: Option<U>,
+    msghdr: libc::msghdr,
+}
+
+impl<T: BoundedBuf, U: BoundedBuf> Op<SendMsg<T, U>> {
+    pub(crate) fn sendmsg(
+        fd: &SharedFd,
+        io_bufs: Vec<T>,
+        socket_addr: Option<SocketAddr>,
+        msg_control: Option<U>,
+    ) -> io::Result<Self> {
+        use io_uring::{opcode, types};
+
+        let mut msghdr: libc::msghdr = unsafe { std::mem::zeroed() };
+
+        let mut io_slices: Vec<IoSlice<'static>> = Vec::with_capacity(io_bufs.len());
+
+        for io_buf in &io_bufs {
+            io_slices.push(IoSlice::new(unsafe {
+                std::slice::from_raw_parts(io_buf.stable_ptr(), io_buf.bytes_init())
+            }))
+        }
+
+        msghdr.msg_iov = io_slices.as_ptr() as *mut _;
+        msghdr.msg_iovlen = io_slices.len() as _;
+
+        let socket_addr = match socket_addr {
+            Some(_socket_addr) => {
+                let socket_addr = Box::new(SockAddr::from(_socket_addr));
+                msghdr.msg_name = socket_addr.as_ptr() as *mut libc::c_void;
+                msghdr.msg_namelen = socket_addr.len();
+                Some(socket_addr)
+            }
+            None => {
+                msghdr.msg_name = std::ptr::null_mut();
+                msghdr.msg_namelen = 0;
+                None
+            }
+        };
+
+        match msg_control {
+            Some(ref _msg_control) => {
+                msghdr.msg_control = _msg_control.stable_ptr() as *mut _;
+                msghdr.msg_controllen = _msg_control.bytes_init();
+            }
+            None => {
+                msghdr.msg_control = std::ptr::null_mut();
+                msghdr.msg_controllen = 0_usize;
+            }
+        }
+
+        CONTEXT.with(|x| {
+            x.handle().expect("Not in a runtime context").submit_op(
+                SendMsg {
+                    _fd: fd.clone(),
+                    _io_bufs: io_bufs,
+                    _socket_addr: socket_addr,
+                    _io_slices: io_slices,
+                    msg_control,
+                    msghdr,
+                },
+                |sendmsg| {
+                    opcode::SendMsg::new(
+                        types::Fd(sendmsg._fd.raw_fd()),
+                        &sendmsg.msghdr as *const _,
+                    )
+                    .build()
+                },
+            )
+        })
+    }
+}
+
+impl<T, U> Completable for SendMsg<T, U> {
+    type Output = (io::Result<usize>, Vec<T>, Option<U>);
+
+    fn complete(self, cqe: CqeResult) -> (io::Result<usize>, Vec<T>, Option<U>) {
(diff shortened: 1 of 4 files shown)


src/io/sendmsg.rs
```diff
@@ -0,0 +1,118 @@
...
+        })
+    }
+}
+
+impl<T, U> Completable for SendMsg<T, U> {
+    type Output = (io::Result<usize>, Vec<T>, Option<U>);
+
+    fn complete(self, cqe: CqeResult) -> (io::Result<usize>, Vec<T>, Option<U>) {
+        // Convert the operation result to `usize`, and add previous byte count
+        let res = cqe.result.map(|v| self.bytes + v as usize);
```
FrankReh, line 101: The cqe.result will be the number of bytes written, so no need to use map and expect to add anything. But you wouldn't have figured that out from my other comments. I just wanted to help avoid any possible confusion.

FrankReh (changes requested):
Mostly looks good. There were no unit tests for the zc version so none expected for this one either.

src/io/sendmsg.rs
Icelk: Thanks!

src/io/sendmsg.rs
```diff
@@ -0,0 +1,118 @@
...
+use crate::io::SharedFd;
+use crate::runtime::driver::op::{Completable, CqeResult, MultiCQEFuture, Op, Updateable};
+use crate::runtime::CONTEXT;
+use socket2::SockAddr;
+use std::io;
+use std::io::IoSlice;
+use std::net::SocketAddr;
+
+pub(crate) struct SendMsg<T, U> {
+    #[allow(dead_code)]
```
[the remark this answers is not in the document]
Icelk: I changed those to use underscores.

src/io/sendmsg.rs
Icelk: I'm just curious: why does CqeResult give an u32? Does that mean io-uring doesn't handle slices longer than 4GB?

src/io/sendmsg.rs
```diff
@@ -0,0 +1,103 @@
...
+            )
+        })
+    }
+}
+
+impl<T, U> Completable for SendMsg<T, U> {
+    type Output = (io::Result<usize>, Vec<T>, Option<U>);
+
+    fn complete(self, cqe: CqeResult) -> (io::Result<usize>, Vec<T>, Option<U>) {
+        // Convert the operation result to `usize`, and add previous byte count
```
FrankReh, line 92: a left over commit.

FrankReh (changes requested):

src/io/sendmsg.rs
FrankReh: Yes, almost. It actually comes from a signed 32 bit field, so 2GB will be the largest value that can be returned. I don't expect you can get that changed but if you want to ask for background, the team at liburing seems to be open to answering questions about io_uring. Here's is the relevant field in their structure.

https://github.com/axboe/liburing/blob/9134244a4e16e68cb514ea5192fcd7e9beca96f5/src/include/liburing/io_uring.h#L358

src/io/sendmsg.rs
```diff
@@ -0,0 +1,103 @@
...
+pub(crate) struct SendMsg<T, U> {
+    _fd: SharedFd,
+    _io_bufs: Vec<T>,
+    _io_slices: Vec<IoSlice<'static>>,
+    _socket_addr: Option<Box<SockAddr>>,
+    msg_control: Option<U>,
+    msghdr: libc::msghdr,
+}
+
+impl<T: BoundedBuf, U: BoundedBuf> Op<SendMsg<T, U>, SingleCQE> {
```
FrankReh, line 19: Super small nit. Does it still compile if you leave out the SingleCQE here? I think that would make it consistent with other files in this src.

Just to keep complexity down, we often want to just follow other examples when there is more than one right way to do things.

FrankReh (changes requested):
If you have time. Just one more nit. Sorry I didn't catch it a few minutes ago.

src/io/sendmsg.rs
Icelk: Thanks!

FrankReh:
@Icelk You've been very responsive to my comments today. Once the existing tests pass, and I merge, will you be able to confirm this does what you want? As we don't have a unit test for any of the UDP sendmsg* methods, and I don't use them myself, it would help to know it works at least for one use case, before you and I loose the context of this PR.

FrankReh (approved):

Icelk:
> @Icelk You've been very responsive to my comments today.

You too, thanks!

> Once the existing tests pass, and I merge, will you be able to confirm this does what you want?

I will certainly test if they work, but I won't be using the headers in production for a while.

Icelk:
I did some testing.

I got `sendmsg_zc` and `sendmsg` to work, after modifying the code: **they're broken right now**. The `msghdr` needs to be boxed for both operations. Should I send a PR?

I also had some trouble setting `msg_control`'s len to `>= 16`. I got:

<details>
<pre><code>thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: Os { code: 22, kind: InvalidInput, message: "Invalid argument" }', examples/udp_socket.rs:26:75
</code></pre></details>

FrankReh:
Okay. Glad you tried it out. Just submit a fix for sendmsg. #260 exists for SendMsgZc.

Icelk:
#264

FrankReh:
> I also had some trouble setting `msg_control`'s len to `>= 16`

I don't have a good idea yet. Are you comfortable that the buffer you passed as the optional msg_control has a safe `stable_ptr` definition and that the bytes_init() value of the buffer reflects the length correctly?

FrankReh:
@Icelk Have you had a chance to see if using `as_ref` fixes the Invalid Input error when length is greater than 15? I know the two don't sound related but sometimes it is hard to predict how a change in a test affects the stack usage.

FrankReh:
> `quinn` seems to pass a 88-byte buffer to the syscall

Does `quinn` also send an 88-byte zero filled buffer or does it fill in the cmsghdr and does it ensure the buffer is aligned properly? I just read alignment can be required and there are cmsg_len, cmsg_level and cmsg_type fields to fill in. I'm more confused why it works for shorter lengths at the moment.

Icelk:
I'm so sorry. You are correct. `quinn` encodes the `msg_control` data, but after the pointer was passed (which is probably why I got confused).

In [this code](https://docs.rs/quinn-udp/latest/src/quinn_udp/unix.rs.html#389-440), they set the control data. [This files](https://docs.rs/quinn-udp/latest/src/quinn_udp/cmsg.rs.html) handles the encoding.

It seems like it would be beneficial for us to expose the entire `hdr`, instead of just the control. But I believe we don't do that to disallow invalid `hdr`s, since we set the IP there. What do you think we could design such an API?

Icelk:
Sorry for the very late response. Lately, I've implemented HTTP/3 and therefore use this code. I've opened another issue, #274. I now realize only the `msg_control` is important, so allowing the user control over `iovec` or the `msg_name` is redundant & unsafe, as you said. The current solution is very nice, indeed.

I've managed to make `tokio-uring` work with `quinn`, with some fiddling. I'll just link the following thread for posterity regarding quinn & tokio-uring compat: https://github.com/quinn-rs/quinn/issues/915

Issue #178, opened 2022-11-19
Redesign `BufResult` to be `std::result::Result`


mzabaluev:
`BufResult` being aliased to a tuple makes it less ergonomic than a standard `Result` could be.
In future, a `std::ops::Try` implementation could help, but it needs a local type to be implemented on.

I propose to make use of the standard infrastructure provided by `Result` and provide the buffer with a new error type:

```rust
use std::error::Error;
use std::io;

pub type BufResult<T, B> = Result<(T, B), BufError<B>>;

#[derive(Debug)]
pub struct BufError<B>(pub io::Error, pub B);

impl<B: Debug> Error for BufError<B> {
    // ...
}
```

mzabaluev:
The currently promoted pattern of error handling could be enabled by an adapter trait:

```rust
pub trait LiftBuf {
    type Output;
    type Buf;
    fn lift_buf(self) -> (io::Result<Self::Output>, Self::Buf);
}

impl<T, B> LiftBuf for BufResult<T, B> {
    type Output = T;
    type Buf = B;

    fn lift_buf(self) -> (io::Result<Self::Output>, Self::Buf) {
        match self {
            Ok((out, buf)) => (Ok(out), buf),
            Err(BufError(e, buf)) => (Err(e), buf),
        }
    }
}
```

FrankReh:
I like it. It places the buffer as the second of an Ok tuple or an Err tuple. Returns would be
```rust
Ok(val, buf)
// or
Err(err, buf)
```

Noah-Kennedy:
I completely agree with this proposed change.

mzabaluev:
Responding to points raised in a [zulip thread](https://rust-lang.zulipchat.com/#narrow/stream/187312-wg-async/topic/IO.20traits/near/311409554):

> I think the issue that it doesn't work like a `Result` is kind of necessary and using `Result<(T, buf), (E, buf)>` is worse than `(Result<T, E>, buf)`. This is because in the error case the caller is going to want to deal with the buffer before propagating the error. If you just propagate the error then I think the buffer cannot be handled properly.

I find that the pattern of error handling depends a lot on the code around the call site. If the calling function does not have the buffer in its return signature, it can't be propagated anyway and you need to drop it. But I believe more often than not the buffer will need to be propagated, so a `BufError` will need to be destructured to its constituent parts or even returned straight in the `Err`.

[...]

PR #267, opened 2023-04-14
Update `BufResult<T, B>` to be `Result<(T,B), BufError<B>>`

Maaarcocr:
# Why? 

Solves https://github.com/tokio-rs/tokio-uring/issues/178

# How? 

I've changed the type to be what was proposed in the issue and then followed the compiler warnings to make rustc happy. 

I've also added a free function `map_buf` which allows to map a function over the buffer on `BufResult`. Let me know if that's okay, but it made writing some code easier.

### DIFF
--- a/examples/cat.rs
+++ b/examples/cat.rs
@@ -29,8 +29,7 @@ fn main() {
 
         loop {
             // Read a chunk
-            let (res, b) = file.read_at(buf, pos).await;
-            let n = res.unwrap();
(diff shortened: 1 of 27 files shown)


ollie-etl (approved):

ileixe (approved):
We've been used this change for a while and its rationale really works well for our side.

Highly recommend to merge this change.

ollie-etl:
Updated in #295

PR #275, opened 2023-07-20
Add `msg_control` access for `UdpSocket::recvmsg`.

### LINKED ISSUE #274: Return msghdr from `recvmsg`
Only the [source and read bytes](https://github.com/tokio-rs/tokio-uring/blob/master/src/io/recvmsg.rs#L64) are returned from `recvmsg`. Can we also return the `msg_control` (consistent with other parts of the API) or the entire `msghdr`?

Icelk:
Closes #274.
The parameter is optional, so this is an opt-in change. However, it changes the public facing API.
I can add another public facing function with this functionality while keeping the old API, if you'd like :)

### DIFF
--- a/src/io/recvmsg.rs
+++ b/src/io/recvmsg.rs
@@ -7,18 +7,23 @@ use std::{
     {boxed::Box, io, net::SocketAddr},
 };
 
-pub(crate) struct RecvMsg<T> {
+pub(crate) struct RecvMsg<T, U = Vec<u8>> {
     #[allow(dead_code)]
     fd: SharedFd,
     pub(crate) buf: Vec<T>,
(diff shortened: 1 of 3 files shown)


Icelk:
My implementation of HTTP/3 (with interop between `tokio-uring` and `quinn`) currently blocks on this. Would it be possible to get this merged?

When is the next version/patch expected to be released? As I mentioned in my initial comment, adding another function would keep the API from breaking.

Icelk:
BTW, the CI isn't broken because of me 🫣

Icelk:
Hi!
Is there any chance this can be merged?

Thanks,
Icelk

PR #295, opened 2024-02-13
Refresh: Update `Result<T, B>` to be `Result<(T,B), Error<B>`

ollie-etl:
This PR is a update of #267, which seems to have gone stale, due to inactivity on in this repo.

Changes the return type to be a Result of tuples, instead of a tuple containing a result and a buffer. This allows this crate to work in a more ergenomic fashion with `?` operator

In addition to updating the original PR, the following changes are made in this one, not in the original

- `BufError` renamed `Error`. This follows the more usual convention of using crate prefixes to qualify error types, rather than the type name itself.
- `BufResult` renamed `Result`. This is the same rational
- `map_buf` helper function implrmented with a sealed trait, to give method calling convention, rather than qualified calls

### DIFF
--- a/src/io/read.rs
+++ b/src/io/read.rs
@@ -1,6 +1,6 @@
 use crate::buf::BoundedBufMut;
 use crate::io::SharedFd;
-use crate::BufResult;
+use crate::Result;
 
 use crate::runtime::driver::op::{Completable, CqeResult, Op};
 use crate::runtime::CONTEXT;
@@ -43,7 +43,7 @@ impl<T> Completable for Read<T>
 where
     T: BoundedBufMut,
 {
-    type Output = BufResult<usize, T>;
+    type Output = Result<usize, T>;
 
     fn complete(self, cqe: CqeResult) -> Self::Output {
         // Convert the operation result to `usize`
@@ -59,6 +59,9 @@ where
             }
         }
 
-        (res, buf)
+        match res {
+            Ok(n) => Ok((n, buf)),
+            Err(e) => Err(crate::Error(e, buf)),
+        }
     }
 }
--- a/src/io/read_fixed.rs
+++ b/src/io/read_fixed.rs
@@ -2,7 +2,7 @@ use crate::buf::fixed::FixedBuf;
 use crate::buf::BoundedBufMut;
 use crate::io::SharedFd;
 use crate::runtime::driver::op::{self, Completable, Op};
-use crate::BufResult;
+use crate::Result;
 
 use crate::runtime::CONTEXT;
 use std::io;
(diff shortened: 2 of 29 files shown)


ollie-etl:
@ileixe would you care to review?
@Noah-Kennedy  / @carllerche / other maintainer? Any opinions?

src/lib.rs
```diff
@@ -264,7 +261,55 @@ impl Builder {
...
+}
+
+impl<T> std::error::Error for Error<T> {
+    fn source(&self) -> Option<&(dyn std::error::Error + 'static)> {
+        Some(&self.0)
+    }
+}
+
+impl<B> Error<B> {
+    /// Applies a function to the contained buffer, returning a new `BufError`.
```
ollie-etl, line 287:
```suggestion
    /// Applies a function to the contained buffer, returning a new `Error`.
```

src/io/read.rs
```diff
@@ -59,6 +59,9 @@ where
             }
         }
 
-        (res, buf)
+        match res {
```
ileixe, line 62: Actually, we've maintained an adaptor trait for this 

```rust
/// Adapter trait to convert Result<T, E> to BufResult<T, B> where E can be
/// converted to std::io::Error.
pub trait WithBuf<T, B>: Sized {
    fn with_buf(self, buf: B) -> T;
}

/// Adaptor implementation for Result<T, E> to BufResult<T, B>.
impl<T, B, E> WithBuf<BufResult<T, B>, B> for Result<T, E>
where
    E: Into<std::io::Error>,
{
    fn with_buf(self, buf: B) -> BufResult<T, B> {
        match self {
            Ok(res) => Ok((res, buf)),
            Err(err) => Err(BufError(err.into(), buf)),
        }
    }
}
```

then, use it here

```rust
res.with_buf(buf)
```

I think it's very common to convert `std::io::Result` to `tokio_uring::Result`, so glad to see the implementation in tokio-uring itself.

ileixe (approved):
Thanks for pushing this forward. We're waiting for a while this :) 

I usually prefer qualifying `Error` and `Result`, but on the other hand I thought Buf prefix worth its own as it's a bit different from plain Result. (We're using BufResult and Result in the same place). Let me lave it to someone else.

src/io/read.rs
ollie-etl: That would certainly remove a heap of boilerplate. I'll add that as a sealed trait and tidy those instances up.

ollie-etl:
@ileixe Does this address your comment?

ileixe:
@ollie-etl Exactly, thanks!

ollie-etl:
@ileixe That makes sense - done

src/io/read_fixed.rs
```diff
@@ -68,6 +68,9 @@ where
             }
         }
 
-        (res, buf)
+        match res {
+            Ok(n) => Ok((n, buf)),
+            Err(e) => Err(crate::Error(e, buf)),
+        }
```
ollie-etl, line 74:
```suggestion
        res.with_buffer(buf)
```

src/io/read_fixed.rs
```diff
@@ -2,7 +2,7 @@ use crate::buf::fixed::FixedBuf;
 use crate::buf::BoundedBufMut;
 use crate::io::SharedFd;
 use crate::runtime::driver::op::{self, Completable, Op};
-use crate::BufResult;
+use crate::Result;
```
ollie-etl, line 5:
```suggestion
use crate::Result;
use crate::WithBuffer;
```

PR #305, opened 2024-06-14
feature: add support for flags in function write_at

kz2d:
Hi. I need support for flags in write operation, and implemented it in this PR.

PR implement struct for flags and don't implement all flags.  
And then pass this flags on write.

Currently I haven't added tests, so if code is alright I would like too add tests.

### DIFF
--- a/src/fs/file.rs
+++ b/src/fs/file.rs
@@ -10,6 +10,47 @@ use std::io;
 use std::os::unix::io::{AsRawFd, FromRawFd, IntoRawFd, RawFd};
 use std::path::Path;
 
+/// A set of flags to pass to `io_uring` along with the operation
+///
+/// For each flag, there are `with_FLAG`, `no_FLAG` and `is_FLAG` methods
+/// for setting, clearing and checking the flag respectively.
+///
+/// The flags are:
+///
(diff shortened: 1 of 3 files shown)


ileixe:
You don't have to add new API for setting flag as it can be added in Unsubmit struct. 

But if it's added in there, I found exposing flag to user may lead to unsafe behavior (e.g. linked op should be submitted at once). So I think here, we at least make the set_flag API unsafe.

kz2d:
Hi I have added options to set flags on `UnsubmittedOneshot` like you have said and made it unsafe.

ileixe (commented):
Also, I think you don't have to redefune flags as io-uring crates already defined it. 

We've used custom set_flag() in our project so I think the change would be fine, but not sure how others think of it.

kz2d:
Hi, can you look at PR, please.

kz2d:
Hi. Any help here?

ileixe:
Hm.. unfortunately, I think nobody currently maintains this repo for now. @ollie-etl is the only one that I know responding. Any advice ollie?

ollie-etl:
I've the same frustration. The repo is dead, as far as I can tell

PR #308, opened 2024-07-17
Add `fs::symlink` function

daniel-levin:
Hello there! I am a happy user of tokio and I am quite excited about what io_uring can bring to the space.

This change adds the `symlink` function to tokio-uring, with the identical signature and documentation as in tokio itself. I added one unit test. I'm merrily doing blocking I/O inside the async block in the unit test, which I _think_ should be fine because it's just a test.

### DIFF
--- a/src/fs/mod.rs
+++ b/src/fs/mod.rs
@@ -20,3 +20,6 @@ mod statx;
 pub use statx::is_dir_regfile;
 pub use statx::statx;
 pub use statx::StatxBuilder;
+
+mod symlink;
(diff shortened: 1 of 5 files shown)


evanrittenhouse:
This LGTM (cc @Noah-Kennedy)

Noah-Kennedy (approved):

Noah-Kennedy:
@daniel-levin I've merged your change regardless of CI, as those failures are unrelated to this change and need to be fixed independently of this.

PR #319, opened 2024-12-14
Switch from io_uring to rustix_uring for broad compatibility

Gelbpunkt:
rustix wraps syscalls without the libc crate, which does not expose a bunch of io-uring related APIs such as statx on musl targets.

With these changes, tokio-uring compiles fine when built for musl targets.

TODO: test and split the minvers fixes + maybe split the change for the listen syscall into their own commits

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -17,10 +17,11 @@ keywords = ["async", "fs", "io-uring"]
 # See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
 
 [dependencies]
-tokio = { version = "1.2", features = ["net", "rt", "sync"] }
-slab = "0.4.2"
+tokio = { version = "1.12", features = ["net", "rt", "sync"] }
+slab = "0.4.4"
 libc = "0.2.80"
-io-uring = "0.6.0"
+rustix = { version = "0.38.42", features = ["net", "io_uring"] }
+rustix-uring = "0.2.0"
 socket2 = { version = "0.4.4", features = ["all"] }
 bytes = { version = "1.0", optional = true }
 futures-util = { version = "0.3.26", default-features = false, features = ["std"] }
--- a/examples/test_create_dir_all.rs
+++ b/examples/test_create_dir_all.rs
@@ -1,3 +1,4 @@
+use rustix::fs::StatxFlags;
 use std::io;
 use std::path::Path;
 use tokio_uring::fs;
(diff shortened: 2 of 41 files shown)


ileixe:
Just out of curiosity, do you know what's the plan for io-uring crate (Do rustix replace it)? Looks like maintainer is same.

Gelbpunkt:
> Just out of curiosity, do you know what's the plan for io-uring crate (Do rustix replace it)? Looks like maintainer is same.

The maintainers aren't the same from what I can see, rustix-uring is a fork of the io-uring crate. I'm not involved with the project so I don't know more than that.

ileixe:
Oh I did not aware it's fork of io-uring (so I saw same maintainer in github contributor view)

Anyway, here's the reason from io-uring side:
 https://github.com/tokio-rs/io-uring/pull/207 

I think it makes sense.

quininer:
If you want musl compatibility, there is no need to switch to `rustix_uring`, just use `linux-raw-sys` (or any other linux headers library) with `io_uring`.

like
```rust
use io_uring::opcode;
use linux_raw_sys::general::statx;

let statx = Box::new(statx {
    // ...
});
let op = opcode::Statx::new(fd, path, Box::into_raw(statx).cast());
```

manio:
Much thanks for this... I was trying to compile on riscv64 with musl and encountered statx missing in libc as well.
First I was doing this kind of workarounds:
```diff
diff --git a/src/fs/statx.rs b/src/fs/statx.rs
index 5a3ccb1..7cbb13e 100644
--- a/src/fs/statx.rs
+++ b/src/fs/statx.rs
@@ -2,6 +2,9 @@ use super::File;
 use crate::io::{cstr, SharedFd};
 use crate::runtime::driver::op::Op;
 use std::{ffi::CString, io, path::Path};
+use io_uring::types::statx;
+use crate::fs::STATX_ALL;
+use crate::fs::STATX_TYPE;
 
 impl File {
     /// Returns statx(2) metadata for an open file via a uring call.
@@ -28,9 +31,9 @@ impl File {
     ///     f.close().await.unwrap();
     /// })
     /// ```
-    pub async fn statx(&self) -> io::Result<libc::statx> {
+    pub async fn statx(&self) -> io::Result<statx> {
         let flags = libc::AT_EMPTY_PATH;
-        let mask = libc::STATX_ALL;
+        let mask = STATX_ALL;
         Op::statx(Some(self.fd.clone()), None, flags, mask)?.await
     }
 
@@ -73,7 +76,7 @@ impl File {
             file: Some(self.fd.clone()),
             path: None,
             flags: libc::AT_EMPTY_PATH,
-            mask: libc::STATX_ALL,
+            mask: STATX_ALL,
         }
     }
 }
 ```
and then I encounter this PR so hopefully this is a proper fix
btw: These changes was also necessary: https://github.com/tokio-rs/tokio-uring/pull/334

PR #326, opened 2025-03-09
Allow specifying R/W flags for reads and writes on ```tokio_uring::fs::File```

ChrysoliteAzalea:
Hello everyone! I propose adding methods to ```tokio_uring::fs::File``` that would allow passing ```RWF_``` flags to read and write operations.

## Explanation

Along with ordinary I/O system calls (```read()```/```write()```), vectored I/O (```readv()```/```writev()```), positional I/O (```pread()```/```pwrite()```) and vectored positional I/O (```preadv()```/```pwritev()```), Linux also has file I/O system calls ```preadv2()``` and ```pwritev2()``` that, in addition to ```preadv()```/```pwritev()```, allow passing read/write flags, that affect individual I/O operations. Examples of such flags are ```RWF_NOWAIT``` (for reading, which means if it has to perform some possibly-blocking operations, it instead should fail with ```EAGAIN```) and ```RWF_APPEND``` (perform an atomic append even if the file is opened without ```O_APPEND``` flag). io_uring also allows passing these flags in submissions.

## Possible benefits

This would be useful for implementing writing to stdout and stderr using io_uring. Stdout and/or stderr may point to regular files, but don't have the ```O_APPEND``` flag set, which means, if several processes have their stdout or stderr redirected to such file, they would have to specify ```RWF_APPEND``` when writing to such file to avoid overwriting each other's data.


ChrysoliteAzalea:
Example of writing to stdout with io_uring:

```
async fn write_to_stdout(data: &str) -> std::io::Result<()>
{
   let b_fd = std::io::stdout().as_fd().try_clone_to_owned()?;
   let file_object = std::fs::File::from(b_fd);
   let iouring_object = tokio_uring::fs::File::from_std(file_object);
   let heapcopy = data.to_owned().into_bytes();
   Ok(iouring_object.write_all_at_with_flags(heapcopy, 0, libc::RWF_APPEND).await.0?)
}
```

ileixe:
It looks too destructive to change all the public API and I believe tokio-uring has similar mental model not to expose I/O flag but provides higher abstraction like Stream.

